PR #1520, opened 2020-05-24
New charge point level refactoring

LucasBrazi06:


jerome-benoit:
The sub-modules build is failing because the commit id referenced does not exist on the sub-module repo.

LucasBrazi06:
> The sub-modules build is failing because the commit id referenced does not exist on the sub-module repo.

I know, check my email, I cannot perform npm install anymore on ev-server.

PR #1523, opened 2020-05-25
Smart charging instant amps in consumption values

OliveGerste:
It needs to be checked, if the migration task runs successfully. (It worked on my site)
If the migration task fails due too many updated entries in the collection at once, it needs to be adjusted again.


OliveGerste:
The conversion of the instant Watts to Amps needs to be handled by @LucasBrazi06, because of the refactoring, which is currently being done.

src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/OCPISessionsService.ts
```diff
@@ -257,6 +257,7 @@ export default class OCPISessionsService {
         endedAt: new Date(session.last_updated),
         consumption: transaction.currentConsumptionWh,
         instantPower: Math.round(transaction.currentConsumption),
+        instantAmps: transaction.currentConsumption / 230, // Utils convert method was updated. Needs to be handled by Serge.
```
jerome-benoit, line 260: It's not that hard to just merge with the master-qa branch and follow the API change of the method: it's just one argument added to the conversion function.

src/server/ocpp/services/OCPPService.ts
```diff
@@ -1170,6 +1170,7 @@ export default class OCPPService {
       } else {
         consumption.consumption = transaction.currentConsumptionWh;
         consumption.instantPower = Math.round(transaction.currentConsumption);
+        consumption.instantAmps = transaction.currentConsumption / 230; // Utils convert method was updated. Needs to be handled by Serge.
```
jerome-benoit, line 1173: It's not that hard to just merge with the master-qa branch and follow the API change of the method: it's just one argument added to the conversion function.

src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/OCPISessionsService.ts
LucasBrazi06: This is consumption from OCPI, we have no voltage available here on charging station or site area.
@ellavas Why in consumption there is neither siteID nor siteAreaID props ?

PR #1538, opened 2020-05-28
Socket.io: do not fill the notifications array buffer when socket.io

jerome-benoit:
server is disabled.

The buffer was never emptied and it probably has cause huge memory
comsuption triggering crash in the past.


jerome-benoit:
> I see no changes related to this issue, am I missed something?

Indirectly related, it's because of the global code review of all socket.io usage code paths required to solve it that I found that bug.

PR #1550, opened 2020-06-02
Add helper methods to get local HTTP server port.

jerome-benoit:


src/server/ocpp/json/JsonCentralSystemServer.ts
```diff
@@ -48,8 +49,8 @@ export default class JsonCentralSystemServer extends CentralSystemServer {
     return null;
   }
 
-  public get address(): string|WebSocket.AddressInfo {
-    return this.wsServer.address();
+  public get port(): number {
```
LucasBrazi06, line 52: Use classic getter/setter

PR #1615, opened 2020-06-24
socket.io: fill the notification buffer if there's a client on a tenant to notify

jerome-benoit:


LucasBrazi06:
Have you tested that the rooms is empty when you have a disconnection with the client?

jerome-benoit:
> Have you tested that the rooms is empty when you have a disconnection with the client?

Yes, the attribute monitored on the socket.io server object that list the number of socket.io clients connected in a room work as expected in case of room client connection, deconnetion.

LucasBrazi06:
> > Have you tested that the rooms is empty when you have a disconnection with the client?
> 
> Yes, the attribute monitored on the socket.io server object that list the number of socket.io clients connected in a room work as expected in case of room client connection, deconnetion.

Perfect!
Thanks!

PR #1657, opened 2020-07-03
New Unit Tests cases (car managment)

mahdibenromdhane:


src/integration/pricing/export/convergent-charging/index.ts
```diff
@@ -1 +1 @@
-export { default } from '../../convergent-charging';
+export { default } from '../../dummy/DummyPricingIntegration';
```
LucasBrazi06, line 1: This file should not be pushed.

src/integration/refund/export/concur/index.ts
```diff
@@ -1 +1 @@
-export { default } from '../../concur';
+export { default } from '../../dummy/DummyRefundIntegration';
```
LucasBrazi06, line 1: This file should not be pushed.

src/integration/pricing/export/convergent-charging/index.ts
mahdibenromdhane: Done

src/integration/refund/export/concur/index.ts
mahdibenromdhane: Done

mahdibenromdhane:
> Unit tests are failing due maybe to my last changes and I wanted to check the logs but no where in the test execution I can find the tenant:
> 
> image
> 
> Like OCPP:
> image
> 
> Same remark as before, check how it works/has been done in similar areas and reproduce it.

Done :)

LucasBrazi06:
> Done :)

Thanks a lot!

PR #1661, opened 2020-07-03
Change Car Converter

mahdibenromdhane:


src/types/Car.ts
```diff
@@ -137,9 +137,23 @@ export interface Car extends CreatedUpdatedProps {
   userIDs?: string;
   carUsers?: UserCar[];
   type?: CarType;
-  converterType?: string;
+  converter?: Converter;
 }
 
+export interface Converter {
+  powerWatts: number;
+  powerAmps: number;
```
jerome-benoit, line 145: You should use here ampPerPhase or ampPhase to follow your way of naming. There is no such a thing like Amperage Power in electricity.

src/types/Car.ts
```diff
@@ -137,9 +137,23 @@ export interface Car extends CreatedUpdatedProps {
   userIDs?: string;
   carUsers?: UserCar[];
   type?: CarType;
-  converterType?: string;
+  converter?: Converter;
 }
 
+export interface Converter {
+  powerWatts: number;
```
jerome-benoit, line 144: power should be enough. 
If you want to handle the unit, you should then add a powerUnit field.

PR #1672, opened 2020-07-09
Site Area - Updated check of phases with chargepoint

OliveGerste:


src/server/rest/service/SiteAreaService.ts
```diff
@@ -154,8 +154,8 @@ export default class SiteAreaService {
           value: chargingStationID
         });
       }
-      for (const connector of chargingStation.connectors) {
-        if (connector.numberOfConnectedPhase !== 1 && siteArea.numberOfPhases === 1 && action === ServerAction.ADD_CHARGING_STATIONS_TO_SITE_AREA) {
+      for (const chargePoints of chargingStation.chargePoints) {
```
LucasBrazi06: Not all the charging stations have charge points (only the ones in the templates).
We have Utils methods to retrieve such informations like:
```
        const chargePoint = Utils.getChargePointFromID(chargingStation, connector.chargePointID);
        const numberOfConnectedPhase = Utils.getNumberOfConnectedPhases(chargingStation, chargePoint, connector.connectorId);
```
I'll make the change, keep that in mind for the next time ;-).

PR #1702, opened 2020-07-27
Implemented handling of an Error report of a user

OliveGerste:


src/notification/remote-push-notification/RemotePushNotificationTask.ts
```diff
@@ -173,6 +173,11 @@ export default class RemotePushNotificationTask implements NotificationTask {
     return Promise.resolve();
   }
 
+  public async sendEndUserErrorNotification(data: EndUserErrorNotification, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void> {
```
LucasBrazi06, line 176: I would expect that we also send a remote push notif

OliveGerste: Tested in debugger.  
I got the app but the mobileToken for my user will be updated in next db backup for local env. I guess.

LucasBrazi06: > Tested in debugger.
> I got the app but the mobileToken for my user will be updated in next db backup for local env. I guess.

Right!
I'll send it to you by email ;-)

OliveGerste: Sounds good :)

PR #1703, opened 2020-07-28
Check if charging station is offline before sending notification

OliveGerste:
I hope that this is the right approach.  Has to be tested on site.

### DIFF
--- a/src/scheduler/tasks/CheckOfflineChargingStationsTask.ts
+++ b/src/scheduler/tasks/CheckOfflineChargingStationsTask.ts
@@ -1,16 +1,17 @@
-import ChargingStationClientFactory from '../../client/ocpp/ChargingStationClientFactory';
-import ChargingStationStorage from '../../storage/mongodb/ChargingStationStorage';
-import { CheckOfflineChargingStationsTaskConfig } from '../../types/TaskConfig';
-import Constants from '../../utils/Constants';
-import { LockEntity } from '../../types/Locking';
-import LockingManager from '../../locking/LockingManager';
-import Logging from '../../utils/Logging';
-import NotificationHandler from '../../notification/NotificationHandler';
-import SchedulerTask from '../SchedulerTask';
-import { ServerAction } from '../../types/Server';
-import Tenant from '../../types/Tenant';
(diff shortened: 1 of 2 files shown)


src/scheduler/tasks/CheckOfflineChargingStationsTask.ts
```diff
@@ -26,14 +27,27 @@ export default class CheckOfflineChargingStationsTask extends SchedulerTask {
...
-          // Send notification
-          await NotificationHandler.sendOfflineChargingStations(
-            tenant.id, {
+          for (let i = 0; i < chargingStations.result.length; i++) {
+            const chargingStationClient = await ChargingStationClientFactory.getChargingStationClient(tenant.id, chargingStations.result[i]);
+            if ((await chargingStationClient.getConfiguration({})).configurationKey.length > 0) {
+              // Update Heartbeat
+              chargingStations.result[i].lastHeartBeat = new Date();
+              // Save
+              await ChargingStationStorage.saveChargingStation(tenant.id, chargingStations.result[i]);
```
LucasBrazi06, line 36: Use ChargingStationStorage.saveChargingStationHeartBeat() instead

src/scheduler/tasks/CheckOfflineChargingStationsTask.ts
```diff
@@ -26,14 +27,27 @@ export default class CheckOfflineChargingStationsTask extends SchedulerTask {
...
+              // Remove from notification
+              chargingStations.result.splice(i);
```
LucasBrazi06, line 38: Did you test that code ?
Because splice will remove the record from the list and the next iteration will miss the next charging station!
I would expect you start from the end:
```
      for (let i = chargingStations.result.length - 1; i >= 0; i--) {
        // Remove
        chargingStations.result.splice(i, 1);
      }

```
And that you test each line of code in the debugger.
Thanks.

OliveGerste: You are right.
I tested it in the debugger for the slf and cah tenant and after some changes it does what it is supposed to do now.

PR #1712, opened 2020-08-04
Charging profiles issues

OliveGerste:
@LucasBrazi06 
Hi Serge, I fixed the cp authorization issues by adding the _siteId_ as argument to the authorization method.
The cp loading issue was fixed by adding the _siteId_ to the db lookup like it is done e.g. in charging stations.
I tried to test everything with basic and admin user and did not find any error.
Could you double check?

### DIFF
--- a/src/server/rest/service/security/ChargingStationSecurity.ts
+++ b/src/server/rest/service/security/ChargingStationSecurity.ts
@@ -199,8 +199,8 @@ export default class ChargingStationSecurity {
       if (filteredChargingProfile && Object.entries(filteredChargingProfile).length !== 0) {
         filteredChargingProfiles.push(filteredChargingProfile);
       }
-      chargingProfiles.count = chargingProfiles.result.length;
     }
+    chargingProfiles.count = filteredChargingProfiles.length;
     chargingProfiles.result = filteredChargingProfiles;
   }
 
--- a/src/server/rest/service/security/ChargingStationSecurity.ts
+++ b/src/server/rest/service/security/ChargingStationSecurity.ts
@@ -196,10 +196,11 @@ export default class ChargingStationSecurity {
     }
     for (const chargingProfile of chargingProfiles.result) {
       const filteredChargingProfile = this.filterChargingProfileResponse(chargingProfile, loggedUser);
-      if (filteredChargingProfile) {
+      if (filteredChargingProfile && Object.entries(filteredChargingProfile).length !== 0) {
         filteredChargingProfiles.push(filteredChargingProfile);
       }
     }
+    chargingProfiles.count = filteredChargingProfiles.length;
     chargingProfiles.result = filteredChargingProfiles;
   }
 
--- a/src/server/rest/service/security/ChargingStationSecurity.ts
+++ b/src/server/rest/service/security/ChargingStationSecurity.ts
@@ -209,7 +209,7 @@ export default class ChargingStationSecurity {
       return null;
     }
     // Check auth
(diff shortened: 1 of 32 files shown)


LucasBrazi06:
Have you taken into account the fact that the organization may not be activated?

src/storage/mongodb/ChargingStationStorage.ts
```diff
@@ -762,6 +762,17 @@ export default class ChargingStationStorage {
       // Convert
       DatabaseUtils.pushConvertObjectIDToString(aggregation, 'chargingStation.siteAreaID');
       DatabaseUtils.pushConvertObjectIDToString(aggregation, 'chargingStation.siteArea.siteID');
+      // Check Site ID
+      if (!Utils.isEmptyArray(params.siteIDs)) {
```
LucasBrazi06, line 766: Filter has always to be added at the beginning

src/server/rest/service/security/ChargingStationSecurity.ts
```diff
@@ -209,7 +209,7 @@ export default class ChargingStationSecurity {
       return null;
     }
     // Check auth
-    if (Authorizations.canReadChargingProfile(loggedUser)) {
+    if (Authorizations.canReadChargingProfile(loggedUser, chargingProfile.chargingStation.siteArea.siteID)) {
```
LucasBrazi06, line 212: Handle the Site Area = null

OliveGerste:
> Have you taken into account the fact that the organization may not be activated?

Indeed there was a problem with that. (empty objects in the response)
I implemented a check in the cp filter method which removes empty objects in the response.

OliveGerste:
> Check my comments.
> You also wrote that you tested with Basic and Admin, did you mean Site Admin?

I tested admin role and 
basic role combined with site admin.

src/storage/mongodb/ChargingStationStorage.ts
```diff
@@ -707,6 +707,33 @@ export default class ChargingStationStorage {
...
+      // Convert
+      DatabaseUtils.pushConvertObjectIDToString(aggregation, 'chargingStation.siteAreaID');
+      DatabaseUtils.pushConvertObjectIDToString(aggregation, 'chargingStation.siteArea.siteID');
+      // Check Site ID
+      if (!Utils.isEmptyArray(params.siteIDs)) {
+        // Build filter
+        aggregation.push({
+          $match: {
+            'chargingStation.siteArea.siteID': {
+              $in: params.siteIDs.map((id) => id)
```
LucasBrazi06, line 730: How did you manage to test when the Side ID is an ObjectID and the conversion here is missing?
I did the change but you have to test it again:
image
Let me know

src/storage/mongodb/ChargingStationStorage.ts
```diff
@@ -707,6 +707,33 @@ export default class ChargingStationStorage {
...
+        tenantID, aggregation, localField: 'chargingStationID', foreignField: '_id',
+        asField: 'chargingStation', oneToOneCardinality: true, oneToOneCardinalityNotNull: false
+      });
+      // Site Areas
+      DatabaseUtils.pushSiteAreaLookupInAggregation({
+        tenantID, aggregation, localField: 'chargingStation.siteAreaID', foreignField: '_id',
+        asField: 'chargingStation.siteArea', oneToOneCardinality: true, oneToOneCardinalityNotNull: false
+      });
+      // Convert
+      DatabaseUtils.pushConvertObjectIDToString(aggregation, 'chargingStation.siteAreaID');
```
LucasBrazi06, line 722: This you can let it at the end.
If you don't get the meaning of these lines your won't know where to put them.
I moved them.

PR #1719, opened 2020-08-09
Robustify the log message anonymisation: 

jerome-benoit:
+ Fix the recursion ending conditions and call;
+ Log missed message anonymisation; 
+ Anonymize also message string (probably not needed but in case of mistake, it's better do not disclose sensitive information).


src/migration/tasks/AddNotificationsFlagsToUsersTask.ts
```diff
@@ -49,24 +49,24 @@ export default class AddNotificationsFlagsToUsersTask extends MigrationTask {
...
-        sendChargingStationRegistered: user.notifications.sendChargingStationRegistered ? user.notifications.sendChargingStationRegistered : false,
-        sendOcpiPatchStatusError: user.notifications.sendOcpiPatchStatusError ? user.notifications.sendOcpiPatchStatusError : false,
-        sendSmtpAuthError: user.notifications.sendSmtpAuthError ? user.notifications.sendSmtpAuthError : false,
-        sendOfflineChargingStations: user.notifications.sendOfflineChargingStations ? user.notifications.sendOfflineChargingStations : false,
-        sendBillingSynchronizationFailed: user.notifications.sendBillingSynchronizationFailed ? user.notifications.sendBillingSynchronizationFailed : false,
-        sendCarCatalogSynchronizationFailed: user.notifications.sendCarCatalogSynchronizationFailed ? user.notifications.sendCarCatalogSynchronizationFailed : false,
-        sendEndUserErrorNotification: user.notifications.sendEndUserErrorNotification ? user.notifications.sendEndUserErrorNotification : false,
+        sendSessionStarted: user.notifications?.sendSessionStarted ? user.notifications.sendSessionStarted : false,
```
LucasBrazi06, line 52: ? works only for optional parameter and Angular template

src/utils/Logging.ts
```diff
@@ -658,36 +658,42 @@ export default class Logging {
...
-                key === 'captcha') {
-              // Anonymize
-              message[key] = Constants.ANONYMIZED_VALUE;
-            }
-          }
-        }
-      }
+    } else if (typeof message === 'string') {
+      // Anonymize
+      message.replace(/((repeat|)[pP]assword|captcha)(\s)(=|:)(\s)(.*)/g, '$1$3$4$5' + Constants.ANONYMIZED_VALUE);
```
LucasBrazi06, line 666: Not understandable by a human being.

src/migration/tasks/AddNotificationsFlagsToUsersTask.ts
jerome-benoit: It's a typescript syntax extension enabled by default.

> PR again melted with import sorting -> difficult to read, I want only few files to review in this PR

Disable the buggy code extensions that do imports sorting like I've already told you and that will never happen again. Every developer must disable it.

src/utils/Logging.ts
jerome-benoit: The regex is simple, more simple than the ones you can read in Constants.ts.
The replace https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/String/replace documentation is clear enough to explain what is done.

src/migration/tasks/AddNotificationsFlagsToUsersTask.ts
LucasBrazi06: > It's a typescript syntax extension enabled by default.
Weird, it didn't work at the time I tested this in ts files but only in template!

LucasBrazi06: > Disable the buggy code extensions that do imports sorting like I've already told you and that will never happen again. Every developer must disable it.

I disabled sort-imports already, is it the right one?

I use this one:
image

Which is related to the Auto Import here:
image

LucasBrazi06:
Ureadable PR

src/migration/tasks/AddNotificationsFlagsToUsersTask.ts
jerome-benoit: No, it's a code extension named sort-imports, not a code settings related to automatic importing.

jerome-benoit:
> Unreadable PR

I'm not going to spend more time on fixing the logic, the level of security and the performance of the recursion used to do log message anonymisation during my vacations.

A correct, secure and optimised implementation is available here in case you're interested: https://github.com/jerome-benoit/ev-server/blob/anonymisation-fixes/src/utils/Logging.ts

PR #1724, opened 2020-08-19
Charging Station: Add a flag named 'inactive'

Namachi06:


LucasBrazi06:
> > > How that flag is going to be mapped with OCPI statuses ?
> > > For example, if a charging station is public and made inactive, the OCPI code should also change the status pushed: let it public but with inoperative status?
> > 
> > 
> > This is the second step we have to discuss, should an inactive charger should refuse the authorization, block the notifications when offline, not being displayed...
> 
> Ok.
> 
> I think the same goes for other extra statuses flags added to charging station data structure like deleted, etc.: the OCPI code does not seem to be handling them at all.



> I think the same goes for other extra statuses flags added to charging station data structure like deleted, etc.: the OCPI code does not seem to be handling them at all.

Yes, we have to fill the gaps of unfinished job ;-)

PR #1730, opened 2020-08-26
Charging plan  no inactivity when tx profile charging plan is ongoing

OliveGerste:


src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -613,6 +614,10 @@ export default class OCPPUtils {
         userID: transaction.userID,
         endedAt: Utils.convertToDate(meterValue.timestamp),
       } as Consumption;
+      // Handle current Connector limitation
+      await OCPPUtils.addConnectorLimitationToConsumption(tenantID, chargingStation, transaction.connectorId, consumption);
+      // Handle current Site Area limitation
+      await Utils.addSiteLimitationToConsumption(tenantID, chargingStation.siteArea, consumption);
```
LucasBrazi06, line 620: It time consuming to have these method here, move it just after the 'if' below if you agree:
```
      // Handle Consumption (Wh/kWh)
      } else if (OCPPUtils.isEnergyActiveImportMeterValue(meterValue)) {
```

LucasBrazi06 (changes requested):
Check my comments

src/server/ocpp/utils/OCPPUtils.ts
OliveGerste: You are right. I moved it.
The issue will be fully solved, when the currentConnectorLimit Method is working correctly.

PR #1754, opened 2020-09-09
Charging station edit define the order of the phase per connector

OliveGerste:
Let me know if any of these points should be changed:

1. Migration Task:
          - CS are updated for three phased site areas to default phase assignment (Phase1 to L1 and so on)
          - CS on single phased site area get the property with a null value 

2. In the updateParameter method of CS, there is no check if the structure of the phaseAssignment is correct (like in the other connector properties)

3. If a siteArea is changed to single phased or to three phased the phaseAssignment of the stations not touched.

4. When a charging station is created, the phase assignment needs to be set. (I guess this has to be done in the submodule charging-station-templates right?)


LucasBrazi06 (changes requested):
- [ ] Handle new charger in BootNotification and if assigned to a Site Area, provide a default assignment: goal is to have no manual change needed in the UI
- [ ] When Charging Station is assigned to a Site Area for the first time, set also the default value.
- [ ] Provide Unit Test

OliveGerste:
> * [ ]  Handle new charger in BootNotification and if assigned to a Site Area, provide a default assignment: goal is to have no manual change needed in the UI
> * [ ]  When Charging Station is assigned to a Site Area for the first time, set also the default value.
> * [ ]  Provide Unit Test

@LucasBrazi06  For the unit test: I would test if the phaseAssignment-property is created at connectorLevel when a boot notification is executed. 
Is there something else what should be tested and where?

LucasBrazi06:
> For the unit test: I would test if the phaseAssignment-property is created at connectorLevel when a boot notification is executed.
> Is there something else what should be tested and where?

Normally the properties at connector level are resolved at Status Notification level that occurs after the Boot Notification.
In SiteAreaService and in ChargingStationService when you assign a Charger to a Site Area or vice versa.

PR #1762, opened 2020-09-17
Transaction   store the phases used in transaction

OliveGerste:
The changes are really prone to errors. I needed a lot of time to eliminate all bugs, which occured.

Following test cases were handled in debugger:
Single phased station (single phased car + three phased car) (TransactionStart, StatusChange, first meterValue with phases)
Three phased station (single phased car + three phased car) (TransactionStart, StatusChange, first meterValue with phases)
Dc Charger (TransactionStart, StatusChange, first meterValue)

Let me know if I missed something.


src/server/ocpp/services/OCPPService.ts
```diff
@@ -356,6 +356,10 @@ export default class OCPPService {
             // Save all
             await ConsumptionStorage.saveConsumption(headers.tenantID, consumption);
           }
+          if (!transaction.phasesUsed && transaction.currentInstantAmps > 0
+          && transaction.numberOfMeterValues >= 1) {
```
LucasBrazi06, line 360: Do we have to check also Amps on L1, L2 and L3 because we may not have them.

src/storage/mongodb/TransactionStorage.ts
```diff
@@ -93,6 +93,7 @@ export default class TransactionStorage {
       currentInstantAmpsL2: Utils.convertToInt(transactionToSave.currentInstantAmpsL2),
       currentInstantAmpsL3: Utils.convertToInt(transactionToSave.currentInstantAmpsL3),
       currentInstantAmpsDC: Utils.convertToInt(transactionToSave.currentInstantAmpsDC),
+      phasesUsed: transactionToSave.phasesUsed,
```
LucasBrazi06, line 96: Do not assign a whole json document but rather each property with the right usage of Utils methods to convert to the right type.

src/utils/Utils.ts
```diff
@@ -89,17 +89,43 @@ export default class Utils {
     return threePhases;
   }
 
+  public static getUsedPhasesInTransactionInProgress(chargingStation: ChargingStation, transaction: Transaction): CSPhasesUsed {
+    const currentType = Utils.getChargingStationCurrentType(chargingStation, null, transaction.connectorId);
+    if (currentType === CurrentType.AC && transaction.currentInstantAmps > 0) {
```
LucasBrazi06, line 94: Double check L1/2/3 also?
Helper method in Utils to check the 4 of them?
Consumption in W.h is always provided and if non null and you don't have Amps info, default on 3 phases.

LucasBrazi06 (changes requested):
Check my comments

src/utils/Utils.ts
OliveGerste: I implemented a method, but left out the check on the W.h, because I found metervalues in the logs, where instant amps were given, but W.h were 0 anyway.

PR #1774, opened 2020-09-25
Tags management: ensure all inputed tag are upper case.

jerome-benoit:
+ Avoid tag id duplication in DB;
+ Fix authentication issue because of the duplication.

That should fix all the tag authentication bugs encountered during latest new
chargers integration (Asana A17).

The OCPI authentication code path have not been checked.


LucasBrazi06 (changes requested):
Tag can be lower case!

jerome-benoit:
> Tag can be lower case!

As discussed, the solution is to have a case insensitive tag management: 

+ Store tag always upper case: create, update;  
+ Case insensitive tag querying: read. 

I can remove the tag case checks in that PR and just keep the comments refinement and small cleanups.

LucasBrazi06:
As discussed, just opened an issue too solve this use case

PR #1813, opened 2020-10-07
OCPP Params: Store a new Custom param in the configuration collection in DB

WillyLellouch:
Custom Param in DB are now tagged with the boolean property 'custom' which identify them as such.
To test : 
- with a live Charging Station, enter a known custom parameter that would be accepted by the station. 
- verify in db that the property is set to true

This is the back-end component


LucasBrazi06:
Feedback sent by email!!!!

PR #1818, opened 2020-10-08
Add default property to tag

mahdibenromdhane:


src/storage/mongodb/UserStorage.ts
```diff
@@ -775,6 +790,9 @@ export default class UserStorage {
     }
     if (!Utils.isEmptyArray(params.userIDs)) {
       filters.userID = { $in: params.userIDs.map((userID) => Utils.convertToObjectID(userID)) };
+      if(params.defaultTag){
```
LucasBrazi06: Spaces in IF statement!

src/migration/tasks/AddDefaultPropertyToTagsTask.ts
```diff
@@ -0,0 +1,50 @@
...
+  async migrate(): Promise<void> {
+    const tenants = await TenantStorage.getTenants({}, Constants.DB_PARAMS_MAX_LIMIT);
+    for (const tenant of tenants.result) {
+      await this.migrateTenant(tenant);
+    }
+  }
+
+  async migrateTenant(tenant: Tenant): Promise<void> {
+    // Add the default property to tags
+    const result = await global.database.getCollection<any>(tenant.id, 'tags').updateMany(
```
LucasBrazi06, line 21: I replaced your multiple updates with only one MongoDB query.
Just keep it in mind

src/migration/tasks/AddDefaultPropertyToTagsTask.ts
```diff
@@ -0,0 +1,50 @@
+import Constants from '../../utils/Constants';
```
LucasBrazi06, line 1: This file was wrongly indented (4 instead of 2).
Maybe due to your new laptop not yet fully set up ;-)

LucasBrazi06 (approved):
Approved but check my comments to avoid same mistake in the future.

PR #1820, opened 2020-10-10
Configuration key: handle also unknown configuration key. 

jerome-benoit:
Asana task A28

And see how DELTA will handle it.


LucasBrazi06:
To be discussed.

jerome-benoit:
> To be discussed.

The goal here is to see if vendors are using it and how. Once we have a clue on it for the current vendors in production, then we can discuss how to properly handle them. We can even change the key name to identify them easily (+ 'Unknown key'). It's harmless and easily reversible.

jerome-benoit:
That's not the correct way to handle the unknownKey array in the response JSON payload. 
In fact, we should dig that array only when custom parameters are set to make explicit error message: the key is rejected because unknown != the key is rejected because the value set is not correct. 
Probably not difficult to add and useful for operators.

LucasBrazi06:
> That's not the correct way to handle the unknownKey array in the response JSON payload.
> In fact, we should dig that array only when custom parameters are set to make explicit error message: the key is rejected because unknown != the key is rejected because the value set is not correct.
> Probably not difficult to add and useful for operators.

In that case, this call will always have an unknown param and we may not need to log an issue in the end:
image

jerome-benoit:
> That's not the correct way to handle the unknownKey array in the response JSON payload.
> In fact, we should dig that array only when custom parameters are set to make explicit error message: the key is rejected because unknown != the key is rejected because the value set is not correct.
> Probably not difficult to add and useful for operators.
> 
> In that case, this call will always have an unknown param and we may not need to log an issue in the end:
>  <https://user-images.githubusercontent.com/25266945/95904080-4c85ec00-0d97-11eb-826d-23b9c0c03366.png>
> 

No, because charging stations shall respond to both keys (with the same response) even if they announce only one. Both keys are part of the standard OCPP parameters and only one vendor fail to respect the standard here.

From the dashboard, we only have one way to set OCPP parameters keys that might be unknown: it’s the custom input field and giving a more informative error message in case of failure is really easy to add.

--
Jérôme Benoit aka fraggle
OpenPGP Key ID : 27B535D3

LucasBrazi06:
> No, because charging stations shall respond to both keys (with the same response) even if they announce only one. Both keys are part of the standard OCPP parameters and only one vendor fail to respect the standard here.

Ok, so we do not do anything when we have unknown key here as you proposed.

> From the dashboard, we only have one way to set OCPP parameters keys that might be unknown: it’s the custom input field and giving a more informative error message in case of failure is really easy to add.

If you try to change an unknown key, it is already rejected by the charging station, it has nothing to do with getConf.

So we can close this issue I think.

jerome-benoit:
> From the dashboard, we only have one way to set OCPP parameters keys that might be unknown: it’s the custom input field and giving a more informative error message in case of failure is really easy to add.
> 
> If you try to change an unknown key, it is already rejected by the charging station, it has nothing to do with getConf.
> 
> So we can close this issue I think.
> 

Giving more informative error message for that use case is of course not a priority but that can be kept as a low priority task for beginner or new comer dev that want to contribute a small feature while learning a piece of the code.

It’s a common way to allow devs to start with something easy but still useful, like it or similar.

++

--
Jérôme Benoit aka fraggle
OpenPGP Key ID : 27B535D3

PR #1824, opened 2020-10-13
Send invoice with own mail framework

Ninjeneer:
image


LucasBrazi06 (changes requested):
Provide template for ES and DE.

src/integration/billing/BillingIntegration.ts
```diff
@@ -364,6 +366,28 @@ export default abstract class BillingIntegration<T extends BillingSetting> {
         detailedMessages: { error: error.message, stack: error.stack }
       });
     }
+    const tenant = await TenantStorage.getTenant(this.tenantID);
```
LucasBrazi06, line 369: I would expect the code above (line 351 -> 368 to not be in that method but in the one which create the invoice.
Here I want only to see the code to send the invoice that's it.

src/integration/billing/BillingIntegration.ts
```diff
@@ -364,6 +366,28 @@ export default abstract class BillingIntegration<T extends BillingSetting> {
...
+      Utils.generateGUID(),
+      invoice.user,
+      {
+        user: invoice.user,
+        evseDashboardInvoiceURL: await Utils.buildEvseBillingInvoicesURL(this.tenantID),
+        evseDashboardURL: Utils.buildEvseURL(tenant.subdomain),
+        invoiceDownloadUrl: await Utils.buildEvseBillingDownloadInvoicesURL(this.tenantID, invoice.id),
+        invoice: invoice
+      }
+    ).catch((error) => {
```
LucasBrazi06, line 382: Don't use this form of catch here, this is an async call and we don't want to handle errors here.

src/integration/billing/BillingIntegration.ts
```diff
@@ -364,6 +366,28 @@ export default abstract class BillingIntegration<T extends BillingSetting> {
         detailedMessages: { error: error.message, stack: error.stack }
       });
     }
+    const tenant = await TenantStorage.getTenant(this.tenantID);
+    // Send async notification
+    NotificationHandler.sendBillingNewInvoiceNotification(
+      this.tenantID,
+      Utils.generateGUID(),
```
LucasBrazi06, line 373: You can pass the invoice ID here

src/utils/Utils.ts
```diff
@@ -1055,28 +1055,33 @@ export default class Utils {
...
+  }
+
+  public static async buildEvseBillingInvoicesURL(tenantID: string): Promise<string> {
+    const tenant = await TenantStorage.getTenant(tenantID);
+    return `${Utils.buildEvseURL(tenant.subdomain)}/invoices`;
+  }
+
+  public static async buildEvseBillingDownloadInvoicesURL(tenantID: string, invoiceID: string): Promise<string> {
+    const tenant = await TenantStorage.getTenant(tenantID);
+    return `${Utils.buildEvseURL(tenant.subdomain)}/invoices?InvoiceID=${invoiceID}#all`;
```
LucasBrazi06, line 1084: I removed all the useless intermediate constants in all buildxxxURL() methods.
I changed the invoiceID -> InvoiceID (first letter in uppercase when using Query params)
Update the code in the front-end to handle this change

LucasBrazi06 (changes requested):
- Send translation files to Tobias for Portuguese ongoing translation and integrate it back in this PR.
- Change the link to a button:
image

src/integration/billing/BillingIntegration.ts
Ninjeneer: Moved it into the concrete billing implementation, as the invoice finalization depends on how the billing provider works.
In this case with Stripe, not at the invoice creation but at the state switch DRAFT -> OPEN

src/integration/billing/BillingIntegration.ts
Ninjeneer: Removed

src/integration/billing/BillingIntegration.ts
Ninjeneer: Done

Ninjeneer:
I had to change the body template structure, to handle multiple buttons in mail notifications, which explains the hundreds of edited files. Nothing relevant, just a massive addition of brackets.

PR #1831, opened 2020-10-15
RESTful API: define specific action for the API.

jerome-benoit:
+ Lower case only,
+ More standard auth enpoints name.

That will lead to duplication during the transition.

### DIFF
--- a/src/server/rest/router/GlobalRouter.ts
+++ b/src/server/rest/router/GlobalRouter.ts
@@ -7,17 +7,16 @@ import { utilRouter } from './util/UtilRouter';
 
 export const globalRouter = express.Router();
 
-// API version 1 routes
-const currentVersion = 1;
-const currentVersionBaseURI = `/v${currentVersion}`;
+// API version 1 top level routes
+
 // Protected API
-globalRouter.use(currentVersionBaseURI + '/api', AuthService.authenticate(), tenantRouter);
+globalRouter.use('/v1/api', AuthService.authenticate(), tenantRouter);
 
 // Auth
-globalRouter.use(currentVersionBaseURI + '/auth', authRouter);
+globalRouter.use('/v1/auth', authRouter);
 
 // Util
-globalRouter.use(currentVersionBaseURI + '/util', utilRouter);
+globalRouter.use('/v1/util', utilRouter);
 
 // Docs
-globalRouter.use(currentVersionBaseURI + '/docs', swaggerRouter);
+globalRouter.use('/v1/docs', swaggerRouter);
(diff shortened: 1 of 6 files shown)


src/types/Server.ts
```diff
@@ -368,4 +368,12 @@ export enum ServerAction {
   HTTP_REQUEST = 'HttpRequest',
   HTTP_RESPONSE = 'HttpResponse',
   HTTP_ERROR = 'HttpError',
+
+  // RESTful API
+  SIGNIN = 'signin',
+  SIGNON = 'signon',
+  SIGNOUT = 'signout',
+  RESTful_PASSWORD_RESET = 'passwordreset',
```
LucasBrazi06, line 376: RESTful_PASSWORD_RESET => REST_PASSWORD_RESET or RESTFUL_PASSWORD_RESET
All in uppercase.

src/server/rest/router/GlobalRouter.ts
```diff
@@ -7,14 +7,17 @@ import { utilRouter } from './util/UtilRouter';
 
 export const globalRouter = express.Router();
 
-// API
-globalRouter.use('/v1/api', AuthService.authenticate(), tenantRouter);
+// API version 1 routes
+const currentVersion = 1;
+const currentVersionBaseURI = `/v${currentVersion}`;
```
LucasBrazi06, line 12: If we add a directory v1, no need to add a variable here I think

src/server/rest/router/util/UtilRouter.ts
```diff
@@ -1,10 +1,11 @@
 import express, { NextFunction, Request, Response } from 'express';
 
+import { ServerAction } from '../../../../types/Server';
 import { StatusCodes } from 'http-status-codes';
 
 export const utilRouter = express.Router();
 
-utilRouter.get('/ping', async (req: Request, res: Response, next: NextFunction) => {
+utilRouter.get('/' + ServerAction.RESTful_PING, async (req: Request, res: Response, next: NextFunction) => {
```
LucasBrazi06, line 8: '/' + ServerAction.RESTful_PING => `/${ServerAction.RESTFUL_PING}` ? :-)

src/types/Server.ts
jerome-benoit: Done.

src/server/rest/router/GlobalRouter.ts
jerome-benoit: You will not have the directory name mapping into the router out of box, unless you do some directory listing and parsing. It's better to just use a constant.

src/server/rest/router/util/UtilRouter.ts
jerome-benoit: Done.

PR #1845, opened 2020-10-20
return users with car list

mahdibenromdhane:


src/server/rest/v1/service/security/CarSecurity.ts
```diff
@@ -264,6 +264,7 @@ export default class CarSecurity {
         type: car.type,
         converter: car.converter,
         carCatalog: car.carCatalog,
+        carUsers: car.carUsers
```
LucasBrazi06, line 267: Use the User filter method instead.

LucasBrazi06 (changes requested):

src/server/rest/v1/service/security/CarSecurity.ts
mahdibenromdhane: we already use the filterMinimalUserResponse in the line after this (same as carCatalog)

LucasBrazi06: Okay :)

PR #1852, opened 2020-10-21
Session Hash not working with several instances of REST

LucasBrazi06:


jerome-benoit (approved):
The cluster module can be used for testing that the application can scale horizontally for stateless protocols.

LucasBrazi06:
> The cluster module can be used for testing that the application can scale horizontally for stateless protocols.

Good, thanks!
Can we still use it in production?

jerome-benoit:
> > The cluster module can be used for testing that the application can scale horizontally for stateless protocols.
> 
> Good, thanks!
> Can we still use it in production?

I do not think so. 
Only to replicate locally and logically the AWS horizontal scaling architecture and ensure there's no issue like that one for stateless services and protocols. It's more like a test bed for devs that want to validate that the code they've done still works properly with HA and horizontal scaling hosting architecture.

PR #1856, opened 2020-10-22
Tags with only anonymized session can be deleted/reassigned

mahdibenromdhane:


src/storage/mongodb/UserStorage.ts
```diff
@@ -845,7 +847,9 @@ export default class UserStorage {
         tenantID, aggregation: aggregation, localField: '_id', foreignField: 'tagID',
         count: true, asField: 'transactionsCount', oneToOneCardinality: false,
         objectIDFields: ['createdBy', 'lastChangedBy']
-      });
+      }, [{
+        '$match': { 'userID': { $exists: true, $ne: null } }
+      }]);
```
LucasBrazi06, line 852: Nbr of transaction is related to tags then this filter is not right.
Use the param withUser flag to set this one.

LucasBrazi06 (changes requested):

LucasBrazi06:
@mahdibenromdhane Could you provide a description for the QA to test (including the auto assignment of the sessions to the new assigned user).
Thanks.

mahdibenromdhane:
> @mahdibenromdhane Could you provide a description for the QA to test (including the auto assignment of the sessions to the new assigned user).
> Thanks.

Done :)

PR #1865, opened 2020-10-27
Added Auth REST Documentation

Ninjeneer:


src/assets/server/rest/v1/docs/open-api-standard.json
```diff
@@ -518,4 +705,4 @@
       }
     }
   }
-}
\ No newline at end of file
+}
```
jerome-benoit, line 708: The whole old auth endpoints have not been ported to the new API. 
First list all the endpoints in the old API and translate them to the new API scheme.

jerome-benoit (changes requested):

src/types/Server.ts
```diff
@@ -374,7 +374,11 @@ export enum ServerAction {
   REST_SIGNIN = 'signin',
   REST_SIGNON = 'signon',
   REST_SIGNOUT = 'signout',
-  REST_PASSWORD_RESET = 'passwordreset',
+  REST_PASSWORD_RESET = 'passwordReset',
```
jerome-benoit, line 377: The URI naming in standard REST API is always lower case. 

Please read https://restfulapi.net/resource-naming/ and the issue on the project where a lengthy discussion happen on REST API standard design. 

After the reading, you will for example deduce that all mail related operations should be under .../mail: 

.../mail/verify
.../mail/verify/resend

And EULA related: 

.../eula
.../eula/verify 

So before you start coding, I think you have to make a proposal on the naming scheme of the REST API endpoints you want to implement to avoid having to change code in tree places ;) 

For example here, verify keyword in URI is debatable, I would really prefer to replace it by check or control. 

So from now on, the sub path in the endpoints URI does not match totally the ServerAction, so you probably have to think about data structure that permits to express: that endpoints will permit to do control on mail or control on eula.

jerome-benoit (changes requested):

src/types/Server.ts
Ninjeneer: Ok, thanks for the informations and advices !

So the Auth API should look like this ?
/auth/signin
/auth/signon
/auth/signout
/auth/password/reset
/auth/mail/check
/auth/mail/resend
/auth/eula
/auth/eula/check

And what about ServerActions ? 
Do I only create an entity constant (e.g REST_MAIL = 'mail') and then call manually REST_MAIL + '/check' or do I create a constant for each endpoint (e.g REST_CHECK_MAIL = 'mail/check') ?

jerome-benoit: I'm fine with your proposal for authentication endpoints that follows that kind of scheme: 

/feature/data/action

For the path, just put the action in ServerAction to begin with, not the full URI subpath with slash (never put slash in constants definition, that can lead to various parsing side effects and portability issues). And hardcode the data part.  
After a bit more work on the new API, we will see what other part of the URI can be 'constanstified'. That kind of cleanups are better targeted after coding, before, we are not sure that we not have overlooked or missed something.

src/types/Server.ts
```diff
@@ -374,7 +374,9 @@ export enum ServerAction {
   REST_SIGNIN = 'signin',
   REST_SIGNON = 'signon',
   REST_SIGNOUT = 'signout',
-  REST_PASSWORD_RESET = 'passwordreset',
+  REST_PASSWORD = 'password',
+  REST_MAIL = 'mail',
+  REST_EULA = 'eula',
```
jerome-benoit, line 379: Well, a mail, password or eula is more close to a data structure than an action :)

An action is more like send (a mail); check (a mail, a password); accept (an eula); reset (a password) ... something you do on a data structure. That's what I mean by putting the action in the URI in ServerAction and let alone the data structure name alone for now ;)

src/server/rest/v1/router/auth/AuthRouter.ts
```diff
@@ -3,6 +3,7 @@ import express, { NextFunction, Request, Response } from 'express';
 import AuthService from '../../service/AuthService';
 import RouterUtils from '../RouterUtils';
 import { ServerAction } from '../../../../../types/Server';
+import sanitize from 'mongo-sanitize';
```
jerome-benoit, line 6: Unused import.

jerome-benoit (changes requested):

src/server/rest/v1/router/auth/AuthRouter.ts
```diff
@@ -38,8 +42,32 @@ export default class AuthRouter {
   }
 
   protected buildRoutePasswordReset(): void {
-    this.router.post(`/${ServerAction.REST_PASSWORD_RESET}`, async (req: Request, res: Response, next: NextFunction) => {
-      await RouterUtils.handleServerAction(AuthService.handleUserPasswordReset.bind(this), ServerAction.REST_PASSWORD_RESET, req, res, next);
+    this.router.post(`/password/${ServerAction.REST_RESET}`, async (req: Request, res: Response, next: NextFunction) => {
```
LucasBrazi06, line 45: Use full path: REST_PASSWORD_RESET and not REST_RESET
The code is not readable and the logs' actions won't be either.
Everything after 'v1/auth' should be a REST constant.

src/server/rest/v1/router/auth/AuthRouter.ts
```diff
@@ -38,8 +42,32 @@ export default class AuthRouter {
...
+      await RouterUtils.handleServerAction(AuthService.handleUserPasswordReset.bind(this), ServerAction.REST_RESET, req, res, next);
+    });
+  }
+
+  protected buildRouteResendVerificationMail(): void {
+    this.router.post(`/mail/${ServerAction.REST_RESEND}`, async (req: Request, res: Response, next: NextFunction) => {
```
LucasBrazi06, line 51: Use full path

src/server/rest/v1/router/auth/AuthRouter.ts
```diff
@@ -38,8 +42,32 @@ export default class AuthRouter {
...
+      await RouterUtils.handleServerAction(AuthService.handleResendVerificationEmail.bind(this), ServerAction.REST_RESEND, req, res, next);
+    });
+  }
+
+  protected buildRouteVerifyMail(): void {
+    this.router.get(`/mail/${ServerAction.REST_CHECK}`, async (req: Request, res: Response, next: NextFunction) => {
```
LucasBrazi06, line 57: Use full path

src/server/rest/v1/router/auth/AuthRouter.ts
```diff
@@ -38,8 +42,32 @@ export default class AuthRouter {
...
+      await RouterUtils.handleServerAction(AuthService.handleVerifyEmail.bind(this), ServerAction.REST_CHECK, req, res, next);
+    });
+  }
+
+  protected buildRouteEndUserLicenseAgreement(): void {
+    this.router.get('/eula', async (req: Request, res: Response, next: NextFunction) => {
```
LucasBrazi06, line 63: Use full path

src/server/rest/v1/router/auth/AuthRouter.ts
```diff
@@ -38,8 +42,32 @@ export default class AuthRouter {
...
+      await RouterUtils.handleServerAction(AuthService.handleGetEndUserLicenseAgreement.bind(this), ServerAction.END_USER_LICENSE_AGREEMENT, req, res, next);
+    });
+  }
+
+  protected buildRouteCheckEndUserLicenseAgreement(): void {
+    this.router.get(`/eula/${ServerAction.REST_CHECK}`, async (req: Request, res: Response, next: NextFunction) => {
```
LucasBrazi06, line 69: Use full path

test/api/client/AuthenticationApi.ts
```diff
@@ -23,7 +24,7 @@ export default class AuthenticationApi {
     // Send
     const response = await this._baseApi.send({
       method: 'POST',
-      url: '/client/auth/Login',
+      url: '/v1/auth/' + ServerAction.REST_SIGNIN,
```
LucasBrazi06, line 27: Use full path

test/api/client/AuthenticationApi.ts
```diff
@@ -42,7 +43,7 @@ export default class AuthenticationApi {
     // Send
     const response = await this._baseApi.send({
       method: 'POST',
-      url: '/client/auth/RegisterUser',
+      url: '/v1/auth/' + ServerAction.REST_SIGNON,
```
LucasBrazi06, line 46: Use full path

test/api/client/AuthenticationApi.ts
```diff
@@ -60,7 +61,7 @@ export default class AuthenticationApi {
     // Send
     const response = await this._baseApi.send({
       method: 'POST',
-      url: '/client/auth/Reset',
+      url: '/v1/auth/password/' + ServerAction.REST_RESET,
```
LucasBrazi06, line 64: Use full path

test/api/client/AuthenticationApi.ts
```diff
@@ -78,7 +79,7 @@ export default class AuthenticationApi {
     // Send
     const response = await this._baseApi.send({
       method: 'POST',
-      url: '/client/auth/ResendVerificationEmail',
+      url: '/v1/auth/mail/' + ServerAction.REST_RESEND,
```
LucasBrazi06, line 82: Use full path

test/api/client/AuthenticationApi.ts
```diff
@@ -96,7 +97,7 @@ export default class AuthenticationApi {
     // Send
     const response = await this._baseApi.send({
       method: 'GET',
-      url: '/client/auth/VerifyEmail',
+      url: '/v1/auth/mail/' + ServerAction.REST_CHECK,
```
LucasBrazi06, line 100: Use full path

LucasBrazi06 (changes requested):

src/server/rest/v1/router/auth/AuthRouter.ts
jerome-benoit: Well, that's not what I asked Loan to do ;)
I wanted to distinguish data structure from action on them: 

Data structures: 
- mail
- password 
- tenants
- users
- ...

Actions: 
- read
- update
- create
- delete
- reset
- resend
- signin 
- ...

It's a REST endpoint URI construction pattern: ../{data)/{action} for requests, especially on GET. The action is not given when the action type is defined by the request types: DELETE, UPDATE, .... But when it does not exist or match, the action is added at the end of the URI path.
I think the entity enum is the closest match for URI data part. But at the beginning, I've told Loan to hardcode them and after more endpoints conversion to the new API, to create or reuse an enum for the data structure name.
Loan can create a log prefix string helper for REST endpoints usage:
```
logPrefix(data, action) {
  switch (action) {
	case 'GET'
		action = ServerAction.READ
		break
    ...	 
  return `${action} on ${data}:` 
} 
```

LucasBrazi06: @jerome-benoit 
I understand your point and it makes sense but the implementation is not correct then.

As you wrote:
- ServerAction should only contain the nouns: email, signin... and not actions
- Action should contain the action: check, reset...
- Handle the Logs Action + Noun: Check Email

I see these issues:
- Action is not lowercase but we can call toLowercase() on it to register the path
- Action filter in Logs is based on ServerAction only, now if we combine both ServerAction and Action, we will may more filters than necessary.

Let me know what you think

jerome-benoit: I see the issue: there was some overlaps with Action and ServerAction. That's why I've started to use ServerAction for the new REST API URI path. But Action is really what I wanted in fact. 

So I think we should first: 

+ Untangle Action and ServerAction;
+ See if it makes sense to allow logging to use Action and combine them in logs filtering 

Or

+ Explicitly separate Action overlaps in ServerAction enum that will be used in REST endpoints and flag them with a prefix like REST_ACTION_... and a comment. It's the easiest solution.

LucasBrazi06: > I see the issue: there was some overlaps with Action and ServerAction. That's why I've started to use ServerAction for the new REST API URI path. But Action is really what I wanted in fact.
> 
> So, I think we should first:
> 
> * Untangle Action and ServerAction;
> * See if it makes sense to allow logging to use Action and combine them in logs filtering
> 
> Or
> 
> * Explicitly separate Action overlaps in ServerAction enum that will be used in REST endpoints and flag them with a prefix like REST_ACTION_... and a comment. It's the easiest solution.

So for the easiest solution will it be for 'reset password': ServerAction.REST_RESET_PASSWORD (='password/reset')?

jerome-benoit: > > * Explicitly separate Action overlaps in ServerAction enum that will be used in REST endpoints and flag them with a prefix like REST_ACTION_... and a comment. It's the easiest solution.
> 
> So for the easiest solution will it be for 'reset password': ServerAction.REST_RESET_PASSWORD (='password/reset')?

No, because you do the reset action on the password data. Just like you do the resend action on mail data. And in ServerAction, we only put the REST related actions with a prefix. 
In that case, we have to add an entity if it miss and the URL is then .../{Entity.}/{ServerAction.}.
If we add the entity handling in the REST services callback, we're Ok for proper logging: we already have the HTTP methods and if the ServerAction is provided, we can make it have precedence over HTTP methods. 
But having specific actions on specific datas is really narrowed to the auth endpoint for now. 
And I think that taking a bit of time to do it properly will help because we do not know if in the future, we will need more than CRUD actions on other endpoints, especially for charging stations that are also special . 

Just like REST endpoints URI path hierarchy is very often mapped to the data hierarchy: 

/tenants -> CRUD via HTTP methods
/tenants/organisations -> CRUD via HTTP methods
/tenants/organisations/sites -> CRUD via HTTP methods
/tenants/organisations/sites/siteareas -> CRUD via HTTP methods

Here the REST endpoints URI path do not also match one entity but we already have the express router hierarchy that will match. 

Since there's no real urgence to implement to new REST API, we can probably do meetings to discuss it to ensure we do it right. The reference on REST ressource naming is https://restfulapi.net/resource-naming/. The specific action are here named controller.

LucasBrazi06: > /tenants -> CRUD via HTTP methods
> /tenants/organisations -> CRUD via HTTP methods
> /tenants/organisations/sites -> CRUD via HTTP methods
> /tenants/organisations/sites/siteareas -> CRUD via HTTP methods

There's no organizations entity, so it's more like this I think:

In Master Tenant:
/tenants
/tenants/1234

Within the tenant:
/companies
/companies/1234/sites
/companies/1234/sites/1234/siteareas
/companies/1234/sites/1234/siteareas/1234

I agree with you to build an endpoint with the Entity and Action: 'password' + 'reset'

But if we want to keep the ServerAction's current behavior which is referencing the endpoints (entities + actions) and being used in the logs' action filter, we can have in ServerAction:
REST_PASSWORD_RESET = `${Entity.EMAIL}/${Action.CHECK}`

And we use REST_PASSWORD_RESET in the routing.

In the end there are 2 aspects:
- External usage of REST API: '/password/reset' which we both agree
- Internal routing construction of this REST API which we have to agree :-)

jerome-benoit: > 
> I agree with you to build an endpoint with the Entity and Action: 'password' + 'reset'
> 
> But if we want to keep the ServerAction's current behaviour which is referencing the endpoints (entities + actions) and being used in the logs' action filter, we can have in ServerAction:
> REST_PASSWORD_RESET = `${Entity.EMAIL}/${Action.CHECK}`
> 
> And we use REST_PASSWORD_RESET in the routing.
> 

That's also fine as long as everything is lower cased as ServerAction building. If we pass that ServerAction to the logging code, we mix the action and the entity, which is not really clean but not a big deal.

LucasBrazi06:
@Ninjeneer 
Follow this pattern below.

For instance for password reset:
Server.ts, in ServerAction enum:
REST_PASSWORD_RESET = ${Entity.EMAIL.toLowercase()}/${Action.CHECK.toLowercase()}

Use ServerAction.REST_PASSWORD_RESET in the routing.

Ninjeneer:
> For instance for password reset:
> Server.ts, in ServerAction enum:
> REST_PASSWORD_RESET = ${Entity.EMAIL.toLowercase()}/${Action.CHECK.toLowercase()}

@LucasBrazi06  As I can see, in Typescript, enums cannot be created with computed string values. Do I switch ServerAction as a class having static attributes ? Or do I switch ServerAction as a dictionnary ?
```
const serverAction: { [name: string] : string; } = {};
serverAction['Login'] = "Login";
...
serverAction['REST_PASSWORD_REST'] = `${Entity.EMAIL.toLowercase()}/${Action.CHECK.toLowercase()}`;
```

LucasBrazi06:
> dictionnary

No, forget about using Action and Entity and use a raw string (back to square one):

For instance for password reset:

Server.ts, in ServerAction enum:
REST_PASSWORD_RESET = 'password/reset'

And use ServerAction.REST_PASSWORD_RESET in the routing.

test/api/client/AuthenticationApi.ts
Ninjeneer: What do you mean by full path ?

jerome-benoit: I do not see what can be done better here at the URI path management.

jerome-benoit (approved):

PR #1869, opened 2020-10-29
OCPI: Workaround Gireve mis-interpretation of the specification

jerome-benoit:
It's ugly as hell ...


LucasBrazi06:
> It's ugly as hell ...

I assume that they won't change their backend :-(

PR #1924, opened 2020-11-23
Fix circular dependency with mocha and Configuration.

jerome-benoit:
Make Configuration self-contained


LucasBrazi06:
Merci

PR #1946, opened 2020-11-26
Not upsert charging station when updating last seen

mahdibenromdhane:


jerome-benoit (approved):

jerome-benoit:
After a discussion with Serge, the fix proposed is probably not correct. 
Could you please give us the call stack that trigger that static method? 

Thanks.

jerome-benoit (commented):
More insight is required.

jerome-benoit:
> After a discussion with Serge, the fix proposed is probably not correct.
> Could you please give us the call stack that trigger that static method?
> 
> Thanks.

I've made a change on that method that replace `global.database.getCollection<any>()` by `global.database.getCollection<ChargingStation>()`. 

If you revert that change, does it also fix the issue?

mahdibenromdhane:
> After a discussion with Serge, the fix proposed is probably not correct.
> Could you please give us the call stack that trigger that static method?
> 
> Thanks.

The ocpp test below is using a faker random charging station ID  that trigger the method :**updateChargingStationLastSeen()** in JsonWSConnection.ts , that trigger the method saveChargingStationLastSeen()

```
static readonly CHARGING_STATION_CONTEXTS: any = {
    **UNREGISTERED_OCPP16: faker.random.alphaNumeric(10),**
    INVALID_IDENTIFIER_OCPP16: 'inv@l!d:1.6',
    ASSIGNED_OCPP16: 'cs-16', // Charging Station is assigned to each site area with OCPP16
    UNASSIGNED_OCPP16: 'cs-notassigned16', // Charging station is not assigned and use OCPP16
    **UNREGISTERED_OCPP15: faker.random.alphaNumeric(10),**
    INVALID_IDENTIFIER_OCPP15: 'inv@l!d-1,5',
    ASSIGNED_OCPP15: 'cs-15', // Charging Station is assigned to each site area with OCPP15
    UNASSIGNED_OCPP15: 'cs-notassigned15' // Charging station is not assigned and use OCPP15
  };
```

```
 describe('For OCPP Version 1.6 (JSON)', () => {
      describe('With unregistered charger', () => {
        before(() => {
          testData.chargingStationContext = testData.tenantContext.getChargingStationContext(ContextDefinition.CHARGING_STATION_CONTEXTS.UNREGISTERED_OCPP16);
          testData.ocppCommonTests.setChargingStation(testData.chargingStationContext);
        });

        after(async () => {
          await testData.chargingStationContext.cleanUpCreatedData();
        });

        it('Should not be possible to register a charging station with invalid token', async () => {
          await testData.ocppCommonTests.testChargingStationRegistrationWithInvalidToken();
        });
      });
```

mahdibenromdhane:
> > After a discussion with Serge, the fix proposed is probably not correct.
> > Could you please give us the call stack that trigger that static method?
> > Thanks.
> 
> I've made a change on that method that replace `global.database.getCollection<any>()` by `global.database.getCollection<ChargingStation>()`.
> 
> If you revert that change, does it also fix the issue?

No i'v tried it and it's not fixed.

LucasBrazi06:
> > > After a discussion with Serge, the fix proposed is probably not correct.
> > > Could you please give us the call stack that trigger that static method?
> > > Thanks.
> > 
> > 
> > I've made a change on that method that replace `global.database.getCollection<any>()` by `global.database.getCollection<ChargingStation>()`.
> > If you revert that change, does it also fix the issue?
> 
> No i'v tried it and it's not fixed.

I think the fix is fine, you save only the lastSeen prop when the charger exists!
@jerome-benoit If the token is not valid or the station is not registered correctly, the method that saves the lastSeen props should not be called! So there is something on your side to fix also I assume.

jerome-benoit:
> > > > After a discussion with Serge, the fix proposed is probably not correct.
> > > > Could you please give us the call stack that trigger that static method?
> > > > Thanks.
> > > 
> > > 
> > > I've made a change on that method that replace `global.database.getCollection<any>()` by `global.database.getCollection<ChargingStation>()`.
> > > If you revert that change, does it also fix the issue?
> > 
> > 
> > No I've tried it and it's not fixed.
> 
> I think the fix is fine, you save only the lastSeen prop when the charger exists!
> @jerome-benoit If the token is not valid or the station is not registered correctly, the method that saves the lastSeen props should not be called! So there is something on your side to fix also I assume.

I think then we can remove upsert for all methods that just update one field of an existing record in a collection. 

I do not understand how a charger that fails to register go to the point where a jsonConnection instance is created.

LucasBrazi06:
> I think then we can remove upsert for all methods that just update one field of an existing record in a collection.
> 
> I do not understand how a charger that fails to register go to the point where a JsonConnection instance is created.

Normally the Json connection occurs first because the message BootNotif is sent, right?
So if the connection succeeds, you will write the lastSeen in the DB before the charger in processed.
I think that's the issue.

jerome-benoit:
> > I think then we can remove upsert for all methods that just update one field of an existing record in a collection.
> > I do not understand how a charger that fails to register go to the point where a JsonConnection instance is created.
> 
> Normally the Json connection occurs first because the message BootNotif is sent, right?
> So if the connection succeeds, you will write the lastSeen in the DB before the charger in processed.
> I think that's the issue.

You're right. The WS ping/pong server side is triggered because of that and the code path where the lastSeen update occurs on a non validated charging station. 
I think another way to fix it it to check if the charging station attached to the JsonConnection is properly registered before calling methods that will do write in the DB in that object instance. 
But we will have then a race condition between JsonConnection instance creation and charging station validation at boot notification. So the current fix is probably the easiest way to ensure no write in the DB are done from a JsonConnection object instance for a not validated charging station.

PR #1951, opened 2020-12-01
Fix ocpp pricing to only 3 decimal degits

mahdibenromdhane:
Fix bug with float sum in node.

Same issue in [this link](https://stackoverflow.com/questions/12511057/float-sum-with-javascript).

### DIFF
--- a/src/server/ocpp/utils/OCPPUtils.ts
+++ b/src/server/ocpp/utils/OCPPUtils.ts
@@ -218,9 +218,9 @@ export default class OCPPUtils {
             if (pricedConsumption.cumulatedAmount) {
               consumption.cumulatedAmount = pricedConsumption.cumulatedAmount;
             } else {
-              consumption.cumulatedAmount = Utils.truncTo(transaction.currentCumulatedPrice + consumption.amount, 6);
+              consumption.cumulatedAmount = Utils.truncTo(transaction.currentCumulatedPrice + consumption.amount, 3);
             }
-            transaction.currentCumulatedPrice = parseFloat(consumption.cumulatedAmount.toFixed(3));
+            transaction.currentCumulatedPrice = consumption.cumulatedAmount;
           }
           break;
         // Stop Transaction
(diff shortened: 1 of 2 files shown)


src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -220,7 +220,7 @@ export default class OCPPUtils {
             } else {
               consumption.cumulatedAmount = Utils.truncTo(transaction.currentCumulatedPrice + consumption.amount, 6);
             }
-            transaction.currentCumulatedPrice = consumption.cumulatedAmount;
+            transaction.currentCumulatedPrice = parseFloat(consumption.cumulatedAmount.toFixed(3));
```
jerome-benoit, line 223: You have to use the Utils.truncTo() helper function instead of toFixed() to follow the same pricing way of 'rounding' used elsewhere.

jerome-benoit (changes requested):

src/server/ocpp/utils/OCPPUtils.ts
mahdibenromdhane: If we use the: 
`Utils.truncTo(2.3 + 2.4, 3) // The result is 4.699 instead of 4.69999999;`
In this case the unit tests will fail as we expect 4.7

jerome-benoit: The unit test should not expect a rounded price to the upper first decimal value. 
If you round 100 000 transactions price up to 0.1 centimes, you get a somewhat important benefit not matching any kind of power delivered and I doubt it's a legal way of rounding the price. 
I think you have to make like Serge have done for other price calculation: never round (toFixed() is doing rounding), always truncate.

mahdibenromdhane: > The unit test should not expect a rounded price to the upper first decimal value.
> If you round 100 000 transactions price up to 0.1 centimes, you get a somewhat important benefit not matching any kind of power delivered and I doubt it's a legal way of rounding the price.
> I think you have to make like Serge have done for other price calculation: never round (toFixed() is doing rounding), always truncate.

True. 
I will change the solution.

PR #1953, opened 2020-12-02
Do not exclude cs permanetly when smart charging fails

OliveGerste:


src/integration/smart-charging/SmartChargingIntegration.ts
```diff
@@ -72,6 +74,11 @@ export default abstract class SmartChargingIntegration<T extends SmartChargingSe
     );
     if (actionsResponse.inError > 0 && retry === false) {
       await this.computeAndApplyChargingProfiles(siteArea, retry = true);
+      // Including the excluded Charging Stations for the next smart charging run again
+      for (const chargingStation of this.excludedChargingStations) {
+        chargingStation.excludeFromSmartCharging = false;
+        await ChargingStationStorage.saveChargingStation(this.tenantID, chargingStation);
+      }
```
LucasBrazi06, line 81: The charger should be excluded without saving it in the DB, so you should have removed the first one which set this flag.
After we have to switch back the charger manually in prod.

LucasBrazi06 (changes requested):

src/integration/smart-charging/SmartChargingIntegration.ts
OliveGerste: @LucasBrazi06 
Okay in this case I would pass the excludedChargingStations Array as a parameter to the buildChargingProfiles method. 
And lower the SiteAreaLimit in adjustSiteLimitation() in SapSmartCharging then. 

BTW I think I found an error in this handling, when it comes to single phased stations.
If we subtract the power of a single phased station from the site area limit, it is substracted from all 3 phases and not just from one like in real life.

LucasBrazi06: > Okay in this case I would pass the excludedChargingStations Array as a parameter to the buildChargingProfiles method.
> And lower the SiteAreaLimit in adjustSiteLimitation() in SapSmartCharging then.

Ok



> BTW I think I found an error in this handling, when it comes to single phased stations.
> If we subtract the power of a single phased station from the site area limit, it is substracted from all 3 phases and not just from one like in real life.

If we can subtract the right phase of the Site Area, okay.

PR #1959, opened 2020-12-05
Notification: Fix SMTP error handling to avoid useless sending tries

jerome-benoit:
+ update emailjs
+ convert to try catch construct with sendAsync() from emailjs
+ type email notification task

Update: 

+ Fix and simplify SMTP server switch;
+ Fix logging of missing SMTP backup configuration section
+ Rename also notifications sourceDescr field to match the new name

Update 2: 
+ Remove SMTP server in use attribute used to optimise the branch prediction


LucasBrazi06:
@jerome-benoit 
Why adding disable flag for backup smtp server instead of not providing the conf. In both case we need to update conf so disabling is just removing the backup conf.
Furthermore, if there is no backup conf, we should notify the Admins when the primary fails and if there is one, we notify  when the backup fails.
Let me know.

jerome-benoit:
> @jerome-benoit
> Why adding disable flag for backup smtp server instead of not providing the conf. In both case we need to update conf so disabling is just removing the backup conf.

In fact, we do not need to have a backup SMTP relay. The client we use already handle email sending failure correctly and properly queue for resend email according to the error returned.  

The code that will trigger the SMTP backup usage is not doing the error type filtering and retry when it should not: - email does not exist anymore on the domain; 
			   - relay under short maintenance; 
			   - ... 

The client can put the message in the queue and we can also send it via the SMTP backup and if the backup sending have a failure, we also put the message in the SMTP backup queue. And the same message can then be sent several times. 	 

Instead of removing the SMTP backup code, I've just make it conditional to begin with and disabled it by default. 
If we want to re-add the SMTP backup feature, we first have to make sure that the email have not been queued by the emailjs client.  

> Furthermore, if there is no backup conf, we should notify the Admins when the primary fails and if there is one, we notify when the backup fails.

LucasBrazi06:
Ok.
We put the backup in place at the beginning because the email service in 1&1 could be blocked due to spam detections.
When you wrote the email client is handling the smtp server error you mean the oss we are using right?
Have you checked in the logs that the retry is really occurring?
If this is the case, indeed we can disable the backup.
And what about the notification to the admin, useless too?

jerome-benoit:
> Ok.
> We put the backup in place at the beginning because the email service in 1&1 could be blocked due to spam detections.

The SAP SMTP relay is reliable. Because of the current conditional use if the backup SMTP server, emails try to go through it even if they really do not have to. 

> When you wrote the email client is handling the smtp server error you mean the oss we are using right?

All emails on a emailjs SMTPClient instance are stacked and they are removed from it if and only if sent successfully.  

> Have you checked in the logs that the retry is really occurring?

Yes, for all no more existing addresses for example, the backup is used for nothing and the mail is stacked in case of relevant errors. And if the backup allow back mails to be sent, its stack is flushed and people get 'spammed'. 
That pattern have occurred for other SMTP error types and that's why some vendor engineers have been 'spammed' on testcharger QA tenant. 

> If this is the case, indeed we can disable the backup.
> And what about the notification to the admin, useless too?

No, I'm working on it to make it generic and cover all SMTP error types. That's why the PR is now a draft.

LucasBrazi06:
> I'm working on it to make it generic and cover all SMTP error types. That's why the PR is now a draft.

Ok thanks.
So draft = I should not merge?

jerome-benoit:
> > I'm working on it to make it generic and cover all SMTP error types. That's why the PR is now a draft.
> 
> Ok thanks.
> So draft = I should not merge?

I think you can't even merge a draft PR, GitHub should not allow it.

jerome-benoit:
After some more testing, the message queueing does not seem to work as described. 
So using a backup SMTP will still be required but only when necessary, filtering on the SMTP error type will be added.

src/client/websocket/WSClient.ts
```diff
@@ -9,16 +9,16 @@ import axiosRetry from 'axios-retry';
 const MODULE_NAME = 'WSClient';
```
LucasBrazi06, line 9: What does it have to do with SMTP backup!?
Mixed PR not allowed.

src/migration/tasks/RenameUserPropertiesTask.ts
```diff
@@ -0,0 +1,44 @@
...
+import MigrationTask from '../MigrationTask';
+import { ServerAction } from '../../types/Server';
+import Tenant from '../../types/Tenant';
+import TenantStorage from '../../storage/mongodb/TenantStorage';
+import User from '../../types/User';
+import global from '../../types/GlobalType';
+
+const MODULE_NAME = 'RenameUserPropertiesTask';
+
+export default class RenameUserPropertiesTask extends MigrationTask {
```
LucasBrazi06, line 12: Was not necessary to rename this prop.
Last time you did the same with lastSeen has brought more issues than it solved.
Test it carefully!

src/notification/NotificationTask.ts
```diff
@@ -16,7 +16,7 @@ export default interface NotificationTask {
   sendChargingStationRegistered(data: ChargingStationRegisteredNotification, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
   sendUnknownUserBadged(data: UnknownUserBadgedNotification, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
   sendSessionStarted(data: TransactionStartedNotification, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
-  sendSmtpAuthError(data: SmtpAuthErrorNotification, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
+  sendSmtpError(data: SmtpErrorNotification, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
```
LucasBrazi06, line 19: Also require front-end changes.

test/api/client/utils/BaseApi.ts
```diff
@@ -26,16 +26,6 @@ export default class BaseApi {
...
-      // Debug
-      // console.log('HTTP Request ====================================');
-      // console.log(httpRequest.baseURL);
-      // console.log(httpRequest.url);
-      // console.log(httpRequest.method);
-      // console.log(httpRequest.data);
-      // console.log(httpResponse.status);
-      // console.log(httpResponse.statusText);
-      // console.log(httpResponse.data);
-      // console.log('====================================');
```
LucasBrazi06, line 38: I put that in comment for debugging, don't remove them

test/api/ocpp/json/OCPPJsonService16.ts
```diff
@@ -145,10 +145,6 @@ export default class OCPPJsonService16 extends OCPPService {
   }
 
   private async send(chargeBoxIdentity: string, message: any): Promise<any> {
-    // Debug
-    // console.log('OCPP Request ====================================');
-    // console.log({ chargeBoxIdentity, message });
-    // console.log('====================================');
```
LucasBrazi06, line 151: Same comment, don't remove.

LucasBrazi06 (commented):
I hope you have tested this PR carefully.
Too many non neceessary changes like var renaming that could impact the prod.

test/api/client/utils/BaseApi.ts
jerome-benoit: Then add a tunable to conditionally enable debug code path.

test/api/ocpp/json/OCPPJsonService16.ts
jerome-benoit: Then add a tunable to conditionally enable debug code path.

src/migration/tasks/RenameUserPropertiesTask.ts
jerome-benoit: > Was not necessary to rename this prop.

Variables must have name that reflect their purpose. Its purpose have changed so the name.  

> Last time you did the same with lastSeen has brought more issues than it solved.
> Test it carefully!

I hope you also plan to apply that rule to your code changes ;)

test/api/ocpp/json/OCPPJsonService16.ts
LucasBrazi06: No

src/migration/tasks/RenameUserPropertiesTask.ts
LucasBrazi06: > > Was not necessary to rename this prop.
> 
> Variables must have name that reflect their purpose. Its purpose have changed so the name.

Except if the impact of this renaming is vast compared to the added value of the renaming which is the case here.

LucasBrazi06: > > Last time you did the same with lastSeen has brought more issues than it solved.
> > Test it carefully!
> 
> I hope you also plan to apply that rule to your code changes ;)

It's your code which is on the table here.
Test carefully.

jerome-benoit: > > > Was not necessary to rename this prop.
> > 
> > 
> > Variables must have name that reflect their purpose. Its purpose have changed so the name.
> 
> Except if the impact of this renaming is vast compared to the added value of the renaming which is the case here.

In order to produce maintainable and readable code for the years to come, that rule must be followed. 
I can't imagine how the glibc and Linux kernel code could would have been if that rule would not have been strictly followed. 
And that's where versioning come into: a change have been done that impact an API used externally.
POSIX version IOTCL, REST requires idempotent versioned endpoints (OCPI include 3 digits versioning in endpoints URI, etc.), etc. 
Changes have been made to the E-Mobility REST API that breaks the idempotence of some endpoints  because no changes to the REST API version where done. The REST API code should handle easily version bump on the REST API. That's not the case actually and that's easy to add: 
	+ add one digit to the version to differentiate between minor and major changes
    + add the version as an argument to method handling REST endpoints requests to format the output according to the version
    + a bit of code factoring to have endpoints common code for endpoints that have not change on a version bump

LucasBrazi06 (changes requested):

src/notification/email/EMailNotificationTask.ts
```diff
@@ -141,87 +135,142 @@ export default class EMailNotificationTask implements NotificationTask {
...
           ]
         });
+      // For Unit Tests only: Tenant is deleted and email is not known thus this Logging statement is always failing with an invalid Tenant
+      // eslint-disable-next-line no-empty
+      } catch (err) { }
+      let sendSmtpError = true;
+      if (error instanceof SMTPError) {
+        const err: SMTPError = error;
+        switch (err.smtp) {
+          case 510:
```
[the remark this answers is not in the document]
jerome-benoit: Unfortunately SMTP server does not always return number type as error code, as specified in the IETF RFC, it might be a string formatted '5.1.1', it might be '511', etc. I can't know right know which format will be returned because we do not log the SMTP error deserialised. 
So an enum might not fit: having one key with several possible values to match. That data structure definition can wait for real world usage.

src/notification/email/EMailNotificationTask.ts
```diff
@@ -141,87 +135,142 @@ export default class EMailNotificationTask implements NotificationTask {
...
+        source: Utils.objectHasProperty(data, 'chargeBoxID') && data.chargeBoxID,
+        action: ServerAction.EMAIL_NOTIFICATION,
+        module: MODULE_NAME, method: 'sendEmail',
+        message: 'No suitable SMTP backup server configuration found or activated to send email',
+        actionOnUser: user
+      });
+      return;
+    }
+    // Switch SMTP server if necessary
+    if (!useBackup && this.backupInUse) {
```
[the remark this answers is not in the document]
jerome-benoit: The initialisation is only done once in any case: the object already knows which SMTP is in use and only switch to the working one via instantiation if necessary. 	
Branching on boolean is less costly than having two instanced heavy SMTPClient object that will not be garbage collected and only once used at a time.

src/notification/email/EMailNotificationTask.ts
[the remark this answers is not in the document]
LucasBrazi06: Ok

src/notification/email/EMailNotificationTask.ts
```diff
@@ -1,48 +1,51 @@
...
-      this.serverBackup = new SMTPClient({
-        user: this.emailConfig.smtpBackup.user,
-        password: this.emailConfig.smtpBackup.password,
-        host: this.emailConfig.smtpBackup.host,
-        port: this.emailConfig.smtpBackup.port,
-        tls: this.emailConfig.smtpBackup.requireTLS,
-        ssl: this.emailConfig.smtpBackup.secure
+  private static instance: EMailNotificationTask;
+  private emailConfig: EmailConfiguration = Configuration.getEmailConfig();
+  private SMTPMainClientInstance: SMTPClient;
```
LucasBrazi06, line 25: Both variables in camel case

src/notification/email/EMailNotificationTask.ts
```diff
@@ -1,48 +1,51 @@
...
+      this.SMTPMainClientInstance = new SMTPClient({
+        user: this.emailConfig.smtp.user,
+        password: this.emailConfig.smtp.password,
+        host: this.emailConfig.smtp.host,
+        port: this.emailConfig.smtp.port,
+        tls: this.emailConfig.smtp.requireTLS,
+        ssl: this.emailConfig.smtp.secure
       });
     }
+    this.backupInUse = false;
```
LucasBrazi06, line 41: Create the SMTPBackupClientInstance here also.
Remove backupInUse member variable.

src/notification/email/EMailNotificationTask.ts
```diff
@@ -141,87 +144,126 @@ export default class EMailNotificationTask implements NotificationTask {
...
+        source: Utils.objectHasProperty(data, 'chargeBoxID') && data.chargeBoxID,
+        action: ServerAction.EMAIL_NOTIFICATION,
+        module: MODULE_NAME, method: 'sendEmail',
+        message: 'No suitable main SMTP server configuration found to send email',
+        actionOnUser: user
+      });
+      return;
+    }
+    // Toggle boolean attribute to switch SMTP server when necessary
+    if (!useBackup && this.backupInUse) {
```
LucasBrazi06, line 162: Remove 'if' statement by removing this.backupInUse and use the useBackup var to get the main or backup SMTP client (check next comment).

src/notification/email/EMailNotificationTask.ts
```diff
@@ -141,87 +144,126 @@ export default class EMailNotificationTask implements NotificationTask {
...
-        }
-      } else {
-        // Email sent successfully
-        Logging.logDebug({
-          tenantID: tenant.id,
-          source: (Utils.objectHasProperty(data, 'chargeBoxID') ? data.chargeBoxID : undefined),
+    });
+    try {
+      // Send the message
+      const messageSent: Message = await this.getSMTPClient().sendAsync(messageToSend);
```
LucasBrazi06, line 180: Simply use useBackup var to get the right SMTP client.
Something like this:
`return useBackup ? SMTPBackupClientInstance : SMTPMainClientInstance`
No need this.backupInUse member var

LucasBrazi06 (changes requested):
Check my comments.

src/notification/email/EMailNotificationTask.ts
jerome-benoit: That's debatable. The class in now a singleton so having the memory of SMTP instance in use make speculative branching works at all computer architecture level and the execution a lot faster. 
Let's keep that kind of optimisation for later to make the code more easily understandable.

src/notification/email/EMailNotificationTask.ts
jerome-benoit: The implementation in the PR only create the SMTP client instance for the backup server in case of an error on the main SMTP server. 
The code that do that small optimisation is easy to read and understand and I do not see why it should be replaced by a less optimised and readable code with syntactic sugar.

src/notification/email/EMailNotificationTask.ts
LucasBrazi06: I would like it now as it'll simplify the code.

jerome-benoit: I do not see what is missing since the previous review request. 
Could you please read the PR changelog and the latest commits patch?

LucasBrazi06: Remove this.backupInUse member variable and create 2 instances of the SMTP servers in the constructor.

jerome-benoit: > Remove this.backupInUse member variable 

Already done since the previous review request. 
Could you please read the PR changelog in description and the latest commits patch?

> and create 2 instances of the SMTP servers in the constructor.

Quote of my previous response on that matter: 

'The implementation in the PR only create the SMTP client instance for the backup server in case of an error on the main SMTP server.
The code that do that small optimisation is easy to read and understand and I do not see why it should be replaced by a less optimised and readable code with syntactic sugar.'

LucasBrazi06: Already checked and read your comments!
backupInUse member variable is still there in GitHub:
image

I simply asked you to remove this handling of the SMTP servers in the first place and you created a singleton instead!
It's neither debatable nor to you to decide.
Remove it for the nth times.

jerome-benoit:
> Already checked and read your comments!
> backupInUse member variable is still there in GitHub:
>  <https://user-images.githubusercontent.com/25266945/106907687-d3702300-66fe-11eb-965c-98445da8597f.png>
> 

You then have an issue with your browser cache or GitHub cache. Checkout and pull the branch to read the up to date code.

> I simply asked you to remove this handling of the SMTP servers in the first place and you created a singleton instead!
> It's neither debatable nor to you to decide.
> 
I’m not going to make code changes that have no sense and degrade memory usage just because.
Unless it’s possible to have a constructive horizontal engineering discussion regarding e-mobility design, I’ve asked to the management to minimise my direct contributions to the e-mobility development to the strict minimum.

--
Jérôme Benoit aka fraggle
OpenPGP Key ID : 27B535D3

LucasBrazi06:
> I’m not going to make code changes that have no sense and degrade memory usage just because. Unless it’s possible to have a constructive horizontal engineering discussion regarding e-mobility design, I’ve asked to the management to minimise my direct contributions to the e-mobility development to the strict minimum.

How much memory are we talking about? kB?
It's just an SMTP instance.
I don't see the memory optimisation here.

LucasBrazi06 (approved):

PR #1968, opened 2020-12-09
Add charging station vendor integration for Shelly wallboxes

ellavas:


src/integration/charging-station-vendor/sap/SAPChargingStationVendorIntegration.ts
```diff
@@ -0,0 +1,8 @@
+import ChargingStation from '../../../types/ChargingStation';
+import ChargingStationVendorIntegration from '../ChargingStationVendorIntegration';
+
+export default class SAPChargingStationVendorIntegration extends ChargingStationVendorIntegration {
```
LucasBrazi06, line 4: Why not Shelly instead of SAP.
SAP does not make chargers.

LucasBrazi06 (changes requested):

LucasBrazi06:
To be linked to an issue for tracking reasons

src/integration/charging-station-vendor/sap/SAPChargingStationVendorIntegration.ts
ellavas: Shelly does not make the charger. It just provides a socket with a relay. The firmware is developed by us and the wallbox is built here.
The vendor name is SAP, this is the reason why I used SAPChargingStationVendorIntegration for the class. 

I can rename it if you prefer.

It would be great to have a default vendor integration which follows the standard to avoid empty integration like this one.

LucasBrazi06: Ok I understand so I would prefer that you use SAP Labs France instead ;-)
Thanks.

ellavas: Done

src/types/ChargingStation.ts
```diff
@@ -278,4 +278,5 @@ export enum ChargerVendor {
   LEGRAND = 'Legrand',
   ATESS = 'ATESS',
   MENNEKES = 'MENNEKES',
+  SAPLABSFRANCE = 'SAP Labs France Caen',
```
LucasBrazi06, line 281: SAP Labs France here too ;-)

ellavas: I can't, it's used as a mapping between the charging station and the integration.
If I change the value in the firmware, boot notification will be rejected for existing charging stations.

image

LucasBrazi06: Despite I'm not happy with your choice, now that it is productive, you're right, it cannot be changed.
Just rename the constant in snake case SAP_LABS_FRANCE.
Thanks!

LucasBrazi06: Don't forget to move the issue back to the right pipeline in Zenhub

ellavas: Renamed. The issue is in the review column, is it ok ?

LucasBrazi06: > The issue is in the review column, is it ok ?

No (check ZenHub)
image

image

ellavas: I was not on the correct board

On my side:
image

LucasBrazi06: Indeed, I removed them.
Thanks for reporting.

PR #1972, opened 2020-12-10
Migrate to per tenant DB fields cyphering

stefania-santimbrean:
Current modifications:

- Store the Cypher key from config.json in new collection default.keys in DB

Questions:

- is the correct place to do this in MongoDBStorage.ts? If not where/when?
- Is the type CryptoKey.ts needed? Currently I have not used it because I was unsure if to use it or use type CryptoConfigurations.ts so I made it with string.


LucasBrazi06:
As I wrote via email, we'll store keys in settings and we have already a storage for that.
We must have a meeting to discuss this.

jerome-benoit:
> As I wrote via email, we'll store keys in settings and we have already a storage for that.
> We must have a meeting to discuss this.

We already have started to discuss via teams, mail: 

+ At first, where the key is stored is not important, the goal at the beginning is to have generic code to handle key change migration on senstivive information. And doing global key change and migration need a temporary permanent storage to store the current key and detect key change and make the decipherng with the old and ciphering with the new. A collection in the DB in the default tenant is the best place. To make Serge happy, Stefania put the keys in one tenant settings and test the migration only for a per tenant key usage. 
+ I've asked Stephania, Kartal and Alfons weeks ago by mail to open draft PRs for their ongoing work to ease implementation to follow the coding standards in the project by doing incremental review until I think the PR is ready for your review and merge.  

That's what I've been requested to do regarding their works: doing the technical framing. So please do not close PRs that will permit me to do it ;)

src/storage/mongodb/CryptoKeyStorage.ts
```diff
@@ -0,0 +1,36 @@
...
+        const keyMDB: any = {
+            _id: new ObjectID(),
+            key: keyToSave
+        };
+
+        // Save document
+        await global.database.getCollection<any>(tenantID, 'keys').insertOne(keyMDB);
+
+    }
+}
```
jerome-benoit, line 36: That file name should be better named: CryptoStorage.ts. 

But you can move that static method in SettingsStorage.ts and code the symetric crypto keys storage handling the same way other settings are stored.

src/storage/mongodb/MongoDBStorage.ts
```diff
@@ -268,6 +269,15 @@ export default class MongoDBStorage {
     // Locks
     await this.handleIndexesInCollection(Constants.DEFAULT_TENANT, 'locks', [
     ]);
+
+    // Crypto Global key
+    await this.handleIndexesInCollection(Constants.DEFAULT_TENANT, 'keys', [
+      { fields: { key: 1 }, options: { unique: true } }
```
jerome-benoit, line 275: Keep in mind that for crypto key change, you need to have access to both keys: the old and new ones.

In order to make that migration reliable and tested for each fields, I think you need to have a data structure like that one: 

+ clear text information (deciphered with the old key)
+ ciphered information with the old key
+ ciphered information with the new key

By doing so, you can conduct tests that the migration is successful for each fields and only store the ciphered information in DB if you sure that it match the clear text once deciphered with the new key.

src/types/CryptoKey.ts
```diff
@@ -0,0 +1,3 @@
+export interface CryptoKey {
+    key: string
```
jerome-benoit, line 2: Make that data structure also able to store the old key.

jerome-benoit (changes requested):

LucasBrazi06:
This has to be fully aligned with Settings like in Settings.ts as it'll be stored here in the end

jerome-benoit:
> This has to be fully aligned with Settings like in Settings.ts as it'll be stored here in the end

My comment on `src/storage/mongodb/CryptoKeyStorage.ts` file saying basically that file have disappeared???

@stefania-santimbrean Have you received it by mail at least?

stefania-santimbrean:
> > This has to be fully aligned with Settings like in Settings.ts as it'll be stored here in the end
> 
> My comment on `src/storage/mongodb/CryptoKeyStorage.ts` file saying basically that file have disappeared???
> 
> @stefania-santimbrean Have you received it by mail at least?

Yes, received by email and I also see your comment in GitHub. I will adapt to your feedback. Thanks!

jerome-benoit:
> > > This has to be fully aligned with Settings like in Settings.ts as it'll be stored here in the end
> > 
> > 
> > My comment on `src/storage/mongodb/CryptoKeyStorage.ts` file saying basically that file have disappeared???
> > @stefania-santimbrean Have you received it by mail at least?
> 
> Yes, received by email and I also see your comment in GitHub. I will adapt to your feedback. Thanks!

Ok.

We will have a meeting to discuss that task, but:
+ the migration shall survive a server reboot/restart. That means that the state of the migration shall be tracked for each sensitive information or that you can test if each sensitive information fields have been migrated or not reliably.
+ You can generate randomly the key in the target test tenant settings to test the migration.

src/types/TenantComponents.ts
```diff
@@ -9,5 +9,6 @@ enum TenantComponents {
   ASSET = 'asset',
   SMART_CHARGING = 'smartCharging',
   CAR = 'car',
+  CRYPTO_KEY = 'cryptoKey'
```
jerome-benoit, line 12: `CRYPTO = 'crypto'`

src/utils/Cypher.ts
```diff
@@ -26,19 +32,79 @@ export default class Cypher {
...
+  public static async detectConfigurationKey(tenantID: string): Promise<void> {
+
+    const configCryptoKey: string = Configuration.getCryptoConfig().key;
+    const keySettings = await SettingStorage.getCryptoKeySettings(tenantID);
+    let cryptoKeySetting: CryptoKeySetting;
+
+    // Check if Crypto Key settings exists
+    if (keySettings) {
+      // Detect new config Cypher key
+      if (keySettings.cryptoKey.newKey && keySettings.cryptoKey.newKey !== configCryptoKey) {
```
jerome-benoit, line 44: You can shorten the test like that: 
`keySettings.cryptoKey?.newKey !== configCryptoKey`

src/scheduler/tasks/MigrateSensitiveDataTask.ts
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable @typescript-eslint/indent */
+/* eslint-disable linebreak-style */
```
jerome-benoit, line 2: Please never disable eslint rules at a file level. Just fix the code ;)

src/scheduler/tasks/MigrateSensitiveDataTask.ts
```diff
@@ -0,0 +1,37 @@
...
+        await Cypher.detectConfigurationKey(tenant.id);
+
+        // Get all settings per tenant
+        const settings = await SettingStorage.getSettings(tenant.id, {},
+            Constants.DB_PARAMS_MAX_LIMIT);
+
+        // Filter settings with sensitiveData
+        const reducedSettings = settings.result.filter((
+            value: SettingDB) => {
+            if (value.sensitiveData && value.sensitiveData.length > 0) {
```
jerome-benoit, line 25: You can shorten the test like that:

`value?.sensitiveData.length`

And to test if an array is empty or not, we have an helper static method in Utils.ts: `Utils.isEmptyArray()`

src/scheduler/tasks/MigrateSensitiveDataTask.ts
```diff
@@ -0,0 +1,37 @@
...
+                return true;
+            }
+        });
+
+        // Save sensitiveData from settings in Migration collection per Tenant
+
+        // Migrate if cryptoKey changed?
+        if (reducedSettings && reducedSettings.length > 0) {
```
jerome-benoit, line 33: Same here.

src/scheduler/tasks/MigrateSensitiveDataTask.ts
```diff
@@ -0,0 +1,37 @@
...
+import { SettingDB } from '../../types/Setting';
+import SettingStorage from '../../storage/mongodb/SettingStorage';
+import { TaskConfig } from '../../types/TaskConfig';
+import Tenant from '../../types/Tenant';
+
+export default class MigrateSensitiveDataTask extends SchedulerTask {
+
+    public async processTenant(tenant: Tenant, config: TaskConfig): Promise<void> {
+        // Detect if Crypto Key changed
+        await Cypher.detectConfigurationKey(tenant.id);
```
jerome-benoit, line 16: It's an initial code sketch but you have to only trigger the migration code path if the key have been changed. 
So returning a boolean here would help to take that decision.

src/storage/mongodb/MongoDBStorage.ts
```diff
@@ -293,6 +296,9 @@ export default class MongoDBStorage {
           await LockingManager.release(createDatabaseLock);
         }
       }
+
+      // Detect changed Config Crypto Key per tenant
+      await Cypher.detectConfigurationKey(tenantId);
```
jerome-benoit, line 301: I think it's gonna be removed at some point. Could you please add a FIXME or TODO comment stating its removal?

src/utils/Cypher.ts
```diff
@@ -26,19 +32,79 @@ export default class Cypher {
...
+
+    if (cryptoKeySetting) {
+      const keySettingToSave = {
+        identifier: TenantComponents.CRYPTO_KEY,
+        type: KeySettingsType.CRYPTO_KEY,
+        cryptoKey: cryptoKeySetting
+      } as KeySettings;
+      await SettingStorage.saveCryptoKeySettings(tenantID, keySettingToSave);
+    }
+  }
```
jerome-benoit, line 77: That method looks really overcomplicated to me: 
+ ciphering bootstrap: only one key per tenant and initial sensitive data cyphering will use it;
+ key updated: two keys key, formerKey and no migration done: sensitive data cyphering will still use the formerKey until the migration is done successfully. No more key update is allowed since the migration is done. So the crypto settings data structure is more like :

```
{
	key: string;
	formerKey?: string;
	migrationDone?: boolean;
}
```

With that kind of data structure, that function becomes a test on a boolean attribute. It then just has to be properly updated at key change and ciphering migration done.

src/utils/Cypher.ts
```diff
@@ -10,6 +15,7 @@ const MODULE_NAME = 'Cypher';
 
 export default class Cypher {
   private static configuration: CryptoConfiguration;
+  private static cryptoKey: CryptoKeySetting;
```
jerome-benoit, line 18: In the end, only that attribute will remain with a more sensible name like `cryptoSetting`, the key in the configuration file will be used to initially populate all tenants crypto settings with a migration.

src/utils/Cypher.ts
```diff
@@ -26,19 +32,79 @@ export default class Cypher {
...
+      await SettingStorage.saveCryptoKeySettings(tenantID, keySettingToSave);
+    }
+  }
+
+  public static async migrateSensitiveData(tenantID: string, setting: SettingDB): Promise<void> {
+    if (this.cryptoKey) {
+      this.decryptSensitiveDataInJSON(setting, this.cryptoKey.oldKey);
+      this.encryptSensitiveDataInJSON(setting, this.cryptoKey.newKey);
+      await SettingStorage.saveSettings(tenantID, setting);
+    }
```
jerome-benoit, line 84: I would really prefer to also have a code path in that method that test if the ciphered value with the key changed really match the 'clear' value deciphered with the formerKey for safety.

src/utils/Cypher.ts
```diff
@@ -26,19 +32,79 @@ export default class Cypher {
...
+  }
+
+  public static async migrateAllSensitiveData(tenantID: string, settings: SettingDB[]): Promise<void> {
+    for (const setting of settings) {
+      await this.migrateSensitiveData(tenantID, setting);
+    }
+  }
+
+  public static encrypt(data: string, key?: string): string {
```
jerome-benoit, line 93: I think you should make the key argument required here. 
That way we know from reading the code calling that method which key is gonna be used.

src/utils/Cypher.ts
```diff
@@ -26,19 +32,79 @@ export default class Cypher {
...
     const iv = crypto.randomBytes(IV_LENGTH);
-    const cipher = crypto.createCipheriv(Cypher.getConfiguration().algorithm, Buffer.from(Cypher.getConfiguration().key), iv);
+    const cryptoKey = key ? key : Cypher.getConfiguration().key;
```
jerome-benoit, line 95: Not needed if the key argument is required.

src/utils/Cypher.ts
```diff
@@ -26,19 +32,79 @@ export default class Cypher {
...
+    const cipher = crypto.createCipheriv(Cypher.getConfiguration().algorithm, Buffer.from(cryptoKey), iv);
     let encryptedData = cipher.update(data);
     encryptedData = Buffer.concat([encryptedData, cipher.final()]);
     return iv.toString('hex') + ':' + encryptedData.toString('hex');
   }
 
-  public static decrypt(data: string): string {
+  public static decrypt(data: string, key?: string): string {
```
jerome-benoit, line 102: Same remark here.

src/utils/Cypher.ts
```diff
@@ -48,7 +114,7 @@ export default class Cypher {
     return crypto.createHash('sha256').update(data).digest('hex');
   }
 
-  public static encryptSensitiveDataInJSON(obj: Record<string, any>): void {
+  public static encryptSensitiveDataInJSON(obj: Record<string, any>, key?: string): void {
```
jerome-benoit, line 117: Same remark here.

src/utils/Cypher.ts
```diff
@@ -82,7 +148,7 @@ export default class Cypher {
     }
   }
 
-  public static decryptSensitiveDataInJSON(obj: Record<string, any>): void {
+  public static decryptSensitiveDataInJSON(obj: Record<string, any>, key?: string): void {
```
jerome-benoit, line 151: Same remark here.

src/types/SensitiveData.ts
```diff
@@ -0,0 +1,21 @@
+export interface SensitiveDataMigration {
```
jerome-benoit, line 1: SensitiveDataMigrationState?

src/scheduler/tasks/MigrateSensitiveDataTask.ts
```diff
@@ -0,0 +1,37 @@
...
+            if (value.sensitiveData && value.sensitiveData.length > 0) {
+                return true;
+            }
+        });
+
+        // Save sensitiveData from settings in Migration collection per Tenant
+
+        // Migrate if cryptoKey changed?
+        if (reducedSettings && reducedSettings.length > 0) {
+            await Cypher.migrateAllSensitiveData(tenant.id, reducedSettings);
```
jerome-benoit, line 34: The migration must be protected by a mutex lock at the tenant level.

jerome-benoit (changes requested):

src/utils/Cypher.ts
stefania-santimbrean: Here I need to check if newKey is undefined because if keySetting is defined, cryptoKey (corrected to crypto) will always be defined (from the way we save the settings). And here I separate the logic if in the DB we have only oldKey (corrected to formerKey) or we have oldKey and newKey (formerKey and key). The test `keySettings.cryptoKey?.newKey !== configCryptoKey` will return true while `keySettings.cryptoKey.newKey && keySettings.cryptoKey.newKey !== configCryptoKey` will return false. I tested now in the debug console. Reading about the safe navigation operator, I don't think I can use it to test that newKey is defined. Or can I?

src/storage/mongodb/MongoDBStorage.ts
stefania-santimbrean: I can remove it right now, because I'm doing it in the Scheduler Task

src/utils/Cypher.ts
stefania-santimbrean: decrypt is also used in StripeBillingIntegration.ts and making key required throws error from there. Should I keep it optional and keep the logic for getting key from Config if not in CryptoSetting?

src/utils/Cypher.ts
stefania-santimbrean: There are quite a few dependencies which had to be modified if we make key required instead of optional. I will commit changes and you can look to see if it's fine or if to keep key optional to not affect all of the other places.

src/utils/Cypher.ts
jerome-benoit: I saw an email about that topic but do not have the time to answer before today. 
It seems that the code now reflect the design recommandation. 
Do you still have issue with it?

src/utils/Cypher.ts
```diff
@@ -26,19 +31,75 @@ export default class Cypher {
...
+
+    const configCryptoKey: string = Configuration.getCryptoConfig().key;
+    const keySettings = await SettingStorage.getCryptoSettings(tenantID);
+    let cryptoSettingToSave: CryptoSetting;
+
+    // Check if Crypto Settings exist
+    if (keySettings) {
+      // Detect new config Cypher key
+      if (keySettings.crypto?.key !== configCryptoKey) {
+        await SettingStorage.deleteSetting(tenantID, keySettings.id);
```
jerome-benoit, line 48: It's probably not needed. The saveCryptoSettings() method is doing an update MongoDB request. 
If you remove it, it should work.

src/utils/Cypher.ts
stefania-santimbrean: I tried to simplify the method. Is it okay now? I still have to integrate the migrationDone part

src/utils/Cypher.ts
stefania-santimbrean: Actually it does not work without it. It's not clear to me why, because I see that saveSettings() is using findOneAndUpdate(). But it blocks there if I don't delete first. The findOneAndUpdate() does not resolve.

stefania-santimbrean: I was not passing the id of the setting, that's why it was throwing an error. Now it works without the deleteSetting. Thanks :)

src/utils/Cypher.ts
```diff
@@ -26,19 +35,176 @@ export default class Cypher {
...
+  }
+
+  public static async detectConfigurationKey(tenantID: string): Promise<boolean> {
+    let isKeyChanged = false;
+    const configCryptoKey: string = Configuration.getCryptoConfig().key;
+    const keySettings = await SettingStorage.getCryptoSettings(tenantID);
+    let cryptoSettingToSave: CryptoSetting;
+    // TODO check for migrationDone
+
+    // Check if Crypto Settings exist
```
jerome-benoit, line 60: Initial crypto settings shall be added at tenant creation. 
A default crypto settings configuration setup code must be added and called at tenant creation and here for safety. 
A random secure key shall be generated.

src/utils/Cypher.ts
```diff
@@ -26,19 +35,176 @@ export default class Cypher {
...
+    const keySettingToSave = {
+      id: this.keySetting.id,
+      identifier: TenantComponents.CRYPTO,
+      type: CryptoSettingsType.CRYPTO,
+      crypto: this.cryptoSetting
+    } as KeySettings;
+    await SettingStorage.saveCryptoSettings(tenantID, keySettingToSave);
+  }
+
+  public static getSensitiveDatainSetting(setting: SettingDB): string[] {
```
jerome-benoit, line 115: There's might be sensitive data elsewhere.

src/utils/Cypher.ts
```diff
@@ -26,19 +35,176 @@ export default class Cypher {
...
+        for (const property of setting.sensitiveData) {
+          const sensitiveData = {
+            identifier: setting.identifier,
+            path: property,
+            initialValue: {
+              encryptedValue: valuebeforeDecrypt[property],
+              key: this.cryptoSetting.formerKey
+            },
+            clearValue: valueAfterDecrypt[property],
+            migratedValue: {
```
jerome-benoit, line 154: To test if a migration should be re-started, you can: 
+ add a reference in the settings to the array record used for current migration in sensitivedatamigration; 
+ test first for the migrationDone boolean;
+ check inside the sensitivedatamigration collection the matching array record if all array elements have initialValue and migratedValue fields filled with correct values.; 
+ run the migration again on missing array elements or not correctly migrated ones: test for migratedValue existence;
+ flag the migration has done.

jerome-benoit (changes requested):

jerome-benoit:
> Questions:
> 
> * is the correct place to do this in MongoDBStorage.ts? If not where/when?

No but you've already created a dedicated file to handle the collection for sensitive data migration. You've answered the question yourself ;)

> * Is the type CryptoKey.ts needed? Currently I have not used it because I was unsure if to use it or use type CryptoConfigurations.ts so I made it with string.

You've also found the solution by yourself ;)

src/types/Setting.ts
```diff
@@ -235,3 +236,23 @@ export interface AssetUserPasswordConnectionType {
...
+
+export interface KeySettings extends Setting {
+  identifier: TenantComponents.CRYPTO;
+  type: CryptoSettingsType;
+  crypto: CryptoSetting;
+}
+
+export interface KeySetting {
+
+}
```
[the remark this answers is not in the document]
stefania-santimbrean: The question is about the empty KeySetting ?

src/utils/Cypher.ts
stefania-santimbrean: elsewhere outside the settings collection?

jerome-benoit: In the chargingstations collection for example.

stefania-santimbrean: But I don't see any sensitiveData field in chargingstations so how can I migrate sensitiveData from other collections?

src/types/Setting.ts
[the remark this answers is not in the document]
stefania-santimbrean: Actually the KeySetting might not be needed because there is only one type of crypto key. I followed the example from other Setting types but maybe in this case I can keep just CryptoSettingsType, KeySettings which I can rename to KeyCryptoSettings and CryptoSetting. What do you think?

src/utils/Cypher.ts
jerome-benoit: I think it has been missed somehow then ... 
OCPP-S can require to save user/password for HTTP authentication to query the charging station.
Could you please have a look the soap charging station client code to confirm it has been missed?

src/types/Setting.ts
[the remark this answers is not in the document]
jerome-benoit: > Actually the KeySetting might not be needed because there is only one type of crypto key. I followed the example from other Setting types but maybe in this case I can keep just CryptoSettingsType, KeySettings which I can rename to KeyCryptoSettings and CryptoSetting. What do you think?

I think the cryptographic settings related to a key shall really be saved. 
Ideally the migration should also be able to handle no key change and algorithm and block cypher change. Let's keep it as a future improvement.

Could you please default the migration to 

blockCypher: AES
type: symmetric
keySize: 256
operationMode: GCM

?

stefania-santimbrean:
Some notes:
- in config.json add under Scheduler->tasks the following:
`{
  "name": "MigrateSensitiveDataTask",
  "active": true,
  "periodicity": "* * * * *",
  "config": {}
}`
- MigrateSensitiveDataTask -> detects when migation is not done and does it
- SchneiderAssetIntegration, StripeBillingIntegration, AddSensitiveDataInSettingsTask were dependencies that needed to be changed because the crypto key is now in db -> don't know how to test this part
- in start.ts and LockingStorage.ts - I've added the part that deletes all database locks related to migration of sensitive data at server start
- In MongoDBStorage we call migrateCryptoKey that migrates the crypto key from config file to DB -> called once

src/scheduler/tasks/MigrateSensitiveDataTask.ts
```diff
@@ -0,0 +1,56 @@
...
+          // If tenant has settings with sensitive data, migrate them
+          if (reducedSettings && !Utils.isEmptyArray(reducedSettings)) {
+            // Set a migration Id per tenant
+            await SensitiveDataMigrationStorage.setMigrationId(migrationId);
+            // Migrate
+            await Cypher.migrateAllSensitiveData(tenant.id, reducedSettings);
+          }
+
+          // Set migration done in Crypto Settings
+          await Cypher.setMigrationDone(tenant.id);
```
jerome-benoit, line 48: The migration can only be considered as done: 
+ if no error are thrown from Cypher.migrateAllSensitiveData: a try {} catch (error) {} construct is missing; 
+ lock have be released successfully.

src/utils/Cypher.ts
```diff
@@ -26,19 +34,168 @@ export default class Cypher {
...
+  public static getCryptoKeySync(): string {
+    return this.cryptoSetting?.migrationDone ? this.cryptoSetting?.key : this.cryptoSetting?.formerKey;
+  }
+
+  public static async getCryptoKey(tenantID: string): Promise<string> {
+    if (!this.cryptoSetting) {
+      this.keySetting = await SettingStorage.getCryptoSettings(tenantID);
+      this.cryptoSetting = this.keySetting.crypto;
+    }
+    return this.cryptoSetting?.migrationDone ? this.cryptoSetting?.key : this.cryptoSetting?.formerKey;
```
jerome-benoit, line 68: You can also use the helper method here.

src/utils/Cypher.ts
jerome-benoit: @stefania-santimbrean unless I've missed something, you've implemented that review comment.

jerome-benoit (changes requested):

src/utils/Cypher.ts
stefania-santimbrean: which helper method you mean?

stefania-santimbrean:
For the first migration of the Crypto Key from the config file to the database, the config file has to contain;
`
"Crypto": {
  "key": "6pKSrPeSgE6ZNwQXvJPxpsH1E6JWCUyy",
  "blockCypher": "AES",
  "keySize": 256,
  "operationMode": "CTR"
}
`

stefania-santimbrean:
**To test this PR:**

1. Make sure you have in config.json the new **Crypto** and **MigrateSensitiveDataTask** information
2. First start the server so that the initial key is migrated to the database
2. To trigger a migration of sensitive data -> In Robo3T change the crypto Setting document to be like this:
```javascript
{
    "_id" : ObjectId("5ff80f4e3b067644a031dace"),
    "content" : {
        "crypto" : {
            "key" : "{{new-crypto-key-256-bits-length}}",
            "keySetting" : {
                "blockCypher" : "AES",
                "keySize" : 256,
                "operationMode" : "GCM"
            },
            "formerKey" : "6pKSrPeSgE6ZNwQXvJPxpsH1E6JWCUyy",
            "formerKeySetting" : {
                "blockCypher" : "AES",
                "keySize" : 256,
                "operationMode" : "CTR"
            },
            "migrationDone" : false
        }
    },
    "createdBy" : null,
    "identifier" : "crypto",
    "lastChangedBy" : null,
    "lastChangedOn" : ISODate("2021-01-08T07:52:46.665Z"),
    "sensitiveData" : null
}
```

LucasBrazi06:
@stefania-santimbrean 
Do not provide any security information in github!!!
I removed the key value from your comment!

LucasBrazi06:
> * in start.ts and LockingStorage.ts - I've added the part that deletes all database locks related to migration of sensitive data at server start

We may have several several batch server instances and only the first one that starts up should remove the locks.
For the time being we have only one but we should find another solution here.

stefania-santimbrean:
> @stefania-santimbrean
> Do not provide any security information in github!!!
> I removed the key value from your comment!

Sorry, I did not realize :(

stefania-santimbrean:
> > * in start.ts and LockingStorage.ts - I've added the part that deletes all database locks related to migration of sensitive data at server start
> 
> We may have several several batch server instances and only the first one that starts up should remove the locks.
> For the time being we have only one but we should find another solution here.

Should this handling now or?

stefania-santimbrean:
@LucasBrazi06 could you review the PR please?

jerome-benoit:
> > > * in start.ts and LockingStorage.ts - I've added the part that deletes all database locks related to migration of sensitive data at server start
> > 
> > 
> > We may have several several batch server instances and only the first one that starts up should remove the locks.
> > For the time being we have only one but we should find another solution here.
> 
> Should this handling now or?

Yes but only on SCP: you have to delete the lock that only have been the same hostname than the current one.

You also have to ensure the deletion is only run within the master processus.

src/utils/Cypher.ts
jerome-benoit: The one to get the correct crypto key.

LucasBrazi06:
> @LucasBrazi06 could you review the PR please?

All conversations should be closed and Jerome should acknowledge so I can start to avoid feeedback conflicts.
Could you check?

src/utils/Cypher.ts
stefania-santimbrean: But this is the function which gets the correct crypto key. I created the getCryptoKeySync() only because in StripeBillingIntegration.ts I need to get the key in the constructor and cannot use the asyncronous getCryptoKey()

src/utils/Cypher.ts
stefania-santimbrean: In ev-server/srv/client/ocpp/soap/SoapChargingStationClient.ts I do not see Cypher being used, nor any user/password. I am looking in the wrong place?

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
+    }
+    return algorithm;
+  }
+
+  public static getCryptoKeySync(): string {
+    return this.cryptoSetting?.migrationDone ? this.cryptoSetting?.key : this.cryptoSetting?.formerKey;
+  }
+
+  public static async getCryptoKey(tenantID: string): Promise<string> {
+    if (!this.cryptoSetting) {
```
LucasBrazi06, line 82: Do not use 'this' in static method.

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
+      this.keySetting = await SettingStorage.getCryptoSettings(tenantID);
+      this.cryptoSetting = this.keySetting.crypto;
+    }
+    return this.cryptoSetting?.migrationDone ? this.cryptoSetting?.key : this.cryptoSetting?.formerKey;
```
LucasBrazi06, line 86: This is wrong, during the migration you can have settings migrated and other not.
To review the full algo with me and to simplify.

src/utils/Cypher.ts
LucasBrazi06: No sensitive date outside settings.

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
+          allSensitiveData[`${property}`] = value;
+        }
+      }
+    }
+    return allSensitiveData;
+
+  }
+
+  public static prepareSaveSensitiveDataMigration(setting: SettingDB,
+    valueBeforeDecrypt: string[], valueAfterDecrypt: string[], valueAfterEncrypt: string[]): SettingSensitiveData {
```
LucasBrazi06, line 140: I should have only one method in this class to migrate sensitive data of a tenant.

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
+
+    return {
+      id: setting.id,
+      identifier: setting.identifier,
+      path: setting.sensitiveData,
+      sensitiveData: allSensitiveData
+    };
+  }
+
+  public static async migrateSensitiveData(tenantID: string, setting: SettingDB): Promise<void> {
```
LucasBrazi06, line 167: To review the algo of the migration with me.

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
+
+    // Prepare sensitive data migration
+    const settingSensitiveData = this.prepareSaveSensitiveDataMigration(setting, valueBeforeDecrypt, valueAfterDecrypt, valueAfterEncrypt);
+    // Save sensitive data migration of setting
+    await SensitiveDataMigrationStorage.saveSensitiveData(tenantID, settingSensitiveData);
+  }
+
+  public static async getMigrateSettingsIdentifiers(tenantID: string): Promise<string[]> {
+    const migration = await SensitiveDataMigrationStorage.getSensitiveDataMigration(tenantID);
+    return migration?.settingSensitiveData?.map((setting) => setting.identifier);
```
LucasBrazi06, line 190: Why should we have settingSensitiveData with no identifier?

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
+  }
+
+  public static async migrateAllSensitiveData(tenantID: string, settings: SettingDB[]): Promise<void> {
```
LucasBrazi06, line 193: Tenant ID is sufficient here, retrieve all the settings to migrate from the DB.

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
+    try {
+      const identifiers = await this.getMigrateSettingsIdentifiers(tenantID);
+      // For each setting that contains sensitive data, migrate that data
+      for (const setting of settings) {
+        if (!identifiers?.includes(setting.identifier)) {
+          await this.migrateSensitiveData(tenantID, setting);
```
LucasBrazi06, line 199: You can put the content of this method here.
Moreover migrateSensitiveData is static and you use this and it's not public but should be private.

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
+        source: Constants.CENTRAL_SERVER,
+        module: MODULE_NAME,
+        method: 'migrateAllSensitiveData',
+        message: 'Migration of sensitive data failed'
+      });
+    }
+  }
+
+  public static encrypt(data: string, key: string): string {
     const iv = crypto.randomBytes(IV_LENGTH);
```
LucasBrazi06, line 213: Why the tenantID is not passed here and why the setting is not retrieved?

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
-    const cipher = crypto.createCipheriv(Cypher.getConfiguration().algorithm, Buffer.from(Cypher.getConfiguration().key), iv);
+    const cipher = crypto.createCipheriv(this.getAlgorithm(key), Buffer.from(key), iv);
     let encryptedData = cipher.update(data);
     encryptedData = Buffer.concat([encryptedData, cipher.final()]);
     return iv.toString('hex') + ':' + encryptedData.toString('hex');
   }
 
-  public static decrypt(data: string): string {
+  public static decrypt(data: string, key: string): string {
```
LucasBrazi06, line 220: Tenant ID is expected also here.

LucasBrazi06 (changes requested):
Globally this topic is complex to handle and you need more assistance.
Code is sometimes not at the right place, you made object oriented basic errors...
We have to review the logic of the migration together in a call.
We will also start by a normal use case like: tenant creation, crypto key generation, change setting, check the unit tests.
Next, we'll change of the key in the settings: migration of the sensitive data + provide unit tests
Finally we'll put in place the migration of existing data.
And we'll keep all of that as simple as possible.

LucasBrazi06:
Error in sub-module SAP Convergeant Charging:



Maybe my sub-modules was not in sync.

src/types/SensitiveData.ts
```diff
@@ -0,0 +1,28 @@
...
+  identifier: string;
+  path: string[];
+  sensitiveData: SensitiveData[]
+}
+
+export interface SensitiveData {
+  identifier: string;
+  path: string;
+  initialValue: EncryptedValue;
+  clearValue: string;
```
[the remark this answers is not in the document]
jerome-benoit: It was a control record field.
@stefania-santimbrean you can remove it from the DB but you have to keep the check that the decryption of the same value cyphered match to handle ciphering error. That way, the storage of the value is not needed while still handling ciphering error.

src/types/Setting.ts
```diff
@@ -238,3 +239,28 @@ export interface AssetUserPasswordConnectionType {
...
+export interface KeyCryptoSetting {
+  blockCypher: string;
+  keySize: number;
+  operationMode: string;
+}
+
+export interface CryptoSetting {
+  key: string;
+  keySetting: KeyCryptoSetting;
+  formerKey?: string;
```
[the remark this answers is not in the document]
jerome-benoit: To handle migration restart at any states. 
You have no guarantee that a crash will not happen before any migration state write in the DB.

src/scheduler/tasks/MigrateSensitiveDataTask.ts
```diff
@@ -0,0 +1,58 @@
...
+            value: SettingDB) => {
+            if (value?.sensitiveData && !Utils.isEmptyArray(value?.sensitiveData)) {
+              return true;
+            }
+          });
+
+          // If tenant has settings with sensitive data, migrate them
+          if (reducedSettings && !Utils.isEmptyArray(reducedSettings)) {
+            // Set a migration Id per tenant
+            await SensitiveDataMigrationStorage.setMigrationId(migrationId);
```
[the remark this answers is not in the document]
jerome-benoit: The migration shall be done that way:
+ a migration task put the crypto settings in the configuration file in each tenant crypto settings (no value change);
+ a scheduler task run once a day detect if a key change has been made in a tenant and run the cyphering migration.

That's why the crypto settings in the tenant keep a record of the two keys, their crypto settings and the status of the migration. 

And why an helper method to get the currently used key is helpful.

src/storage/mongodb/MongoDBStorage.ts
```diff
@@ -316,6 +332,9 @@ export default class MongoDBStorage {
     if (!foundCollection) {
       try {
         await this.db.createCollection(tenantCollectionName);
+        if (name === 'settings') {
+          await SettingStorage.initSettings(tenantID);
```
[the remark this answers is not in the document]
jerome-benoit: It's the right place. At tenant creation, an initial random key shall be created.

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
+      id: this.keySetting.id,
+      identifier: TenantComponents.CRYPTO,
+      type: CryptoSettingsType.CRYPTO,
+      crypto: this.cryptoSetting
+    } as KeySettings;
+    await SettingStorage.saveCryptoSettings(tenantID, keySettingToSave);
+  }
+
+  public static async saveMigrationId(tenantID:string, migrationId: string): Promise<void> {
+    this.cryptoSetting.sensitiveDataMigrationId = migrationId;
```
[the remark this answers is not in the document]
jerome-benoit: > Error in sub-module SAP Convergeant Charging:
> 
> 
> 
> Maybe my sub-modules was not in sync.

Most developers do not have access to private sub repos. So if they change an API used in a submodule, it does not build anymore and they are not aware of it.

src/types/Setting.ts
```diff
@@ -238,3 +239,28 @@ export interface AssetUserPasswordConnectionType {
...
 export interface AssetSchneiderConnectionType extends AssetUserPasswordConnectionType {
 }
+
+export enum CryptoSettingsType {
+  CRYPTO = 'crypto'
+}
+
+export interface KeySettings extends Setting {
+  identifier: TenantComponents.CRYPTO;
+  type: CryptoSettingsType;
```
jerome-benoit, line 249: No need for an 's' here.

src/utils/Cypher.ts
```diff
@@ -26,19 +35,193 @@ export default class Cypher {
...
+    if (key === this.cryptoSetting?.formerKey) {
+      algorithm = `${this.cryptoSetting.formerKeySetting.blockCypher.toLowerCase()}-${this.cryptoSetting.formerKeySetting.keySize}-${this.cryptoSetting.formerKeySetting.operationMode.toLowerCase()}`;
+    } else if (key === this.cryptoSetting?.key) {
+      algorithm = `${this.cryptoSetting.keySetting.blockCypher.toLowerCase()}-${this.cryptoSetting.keySetting.keySize}-${this.cryptoSetting.keySetting.operationMode.toLowerCase()}`;
+    }
+    return algorithm;
+  }
+
+  public static getCryptoKeySync(): string {
+    return this.cryptoSetting?.migrationDone ? this.cryptoSetting?.key : this.cryptoSetting?.formerKey;
```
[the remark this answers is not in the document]
jerome-benoit: Debatable. Given that the migration is done asynchronously, an helper method to ensure you get the current and valid key is helpful.

The old cypher API was not tied to a multi-tenant design, only wrapper around node.js cyphering library. 

I think we can make it tenant aware but I would keep the wrapper and make them private. 
That way public method of the cypher class always have , tenantID: string, in their signature and the low level methods are hidden.

src/utils/Cypher.ts
jerome-benoit: > I would expect that you first try to access the DB and next that you fallback to the value in the config.json file.
> Do not access 'this' in a static method!
> Moreover 'cryptoSetting' is per tenant, why do you keep it as a member class variable!?

Given the review, making the Cypher class multi tenant aware should solve most of the comments. 

@stefania-santimbrean : Cypher can stay mostly has an helpers class of static methods but the all public methods shall be tenant aware.

src/types/configuration/CryptoConfiguration.ts
```diff
@@ -1,4 +1,7 @@
 export default interface CryptoConfiguration {
   key: string;
   algorithm: string;
+  blockCypher?: string;
+  keySize?: number;
+  operationMode?: string;
```
jerome-benoit, line 6: These 3 fields are key dependent settings. They need to be attached to a key, not the global configuration.

src/types/Setting.ts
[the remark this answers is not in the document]
LucasBrazi06: Then only tracking the migrated key (passwords, ...) which is the setting should be sufficient.

src/scheduler/tasks/MigrateSensitiveDataTask.ts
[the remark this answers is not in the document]
LucasBrazi06: This approach is ok and any changes of the key in the settings is not allowed until the scheduled task runs and migrates all the data.
I think that the code org could be greatly simplified.

src/storage/mongodb/MongoDBStorage.ts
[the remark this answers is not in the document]
LucasBrazi06: This method is called handleIndexesInCollection() and not createInitialSettingsForTenant() ;-)
Should be moved in a new method.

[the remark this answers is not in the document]
LucasBrazi06: And SettingStorage should be used to deal with the DB only, no business logic inside.

src/utils/Cypher.ts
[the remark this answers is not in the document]
LucasBrazi06: This method has been created only for Stripe which has to be refactored so this is why we don't need it anymore.

src/utils/Cypher.ts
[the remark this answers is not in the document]
LucasBrazi06: Yes right, my bad!
I just added her in these repos.
@stefania-santimbrean resync the backend to get the sub-modules.

src/scheduler/tasks/MigrateSensitiveDataTask.ts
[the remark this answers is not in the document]
jerome-benoit: > This approach is ok and any changes of the key in the settings is not allowed until the scheduled task runs and migrates all the data.
> I think that the code org could be greatly simplified. 

I think making the Cypher class multi tenant aware will simplify it: The key handling complexity is hidden and the API change worth it. 

The way to do initial migration have been misunderstood and not put as a migration task. And I've not reviewed the code between the request and the review ;)

src/storage/mongodb/MongoDBStorage.ts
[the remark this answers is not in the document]
jerome-benoit: It can be moved to the tenant creation REST endpoint, at the end of the DB creation process;

src/storage/mongodb/SensitiveDataMigrationStorage.ts
```diff
@@ -0,0 +1,87 @@
...
+import Cypher from '../../utils/Cypher';
+import DatabaseUtils from './DatabaseUtils';
+import { ObjectID } from 'mongodb';
+import Utils from '../../utils/Utils';
+import global from './../../types/GlobalType';
+
+const MODULE_NAME = 'SensitiveDataMigrationStorage';
+
+export default class SensitiveDataMigrationStorage {
+  private static migrationId: ObjectID;
```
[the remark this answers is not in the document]
jerome-benoit: It's not really a migration task like the others. It do not used the current migration framework made to run migration only at startup. 
It's a scheduler task that use it's own collection to track the migration state that permit to survive to a migration stopped in the middle.

The id is required to have a reference to the current migration run but it should follow the same naming used everywhere: id in the object and _id in the DB.

src/types/Setting.ts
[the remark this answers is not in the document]
jerome-benoit: Not sure I understand the sentence :): 
+ The migration is asynchronous, both keys and their according crypto settings need to be stored. We can change the way (data structure) or/and where (collection) are stored but it make sense to have them in the settings collection since we do not have a settings history. And I would prefer the settings endpoint inputs treatment to not play with other collections. 
+ The migration state data structure can be simplified but it still has to permit to keep track of the migration of each records. The state record have been made self-contained also by design, at least at the beginning for debug purpose: with one state record, you should be able to do the migration of a non migrated record and see if everything have worked as expected. I think now we can remove some fields like clear text, the key and its settings. 
+ The state of one migration record is given by the presence of both ciphered string, it can also be the former ciphered string with a boolean stating its migration status. As long as the status can be tracked at the record level and that we can check that no memory corruption have make the migration fail, that's fine by me.

src/storage/mongodb/SensitiveDataMigrationStorage.ts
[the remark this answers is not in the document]
LucasBrazi06: If there is a dedicated collection, the write in the 'settings' and in 'sensitivedatamigrations' must be done in a transaction otherwise you can manage writing the settings and fail writing the sensitivedatamigrations.
I think tracking the changes only in settings collection should sufficient and would spare the code to handle the sensitivedatamigrations.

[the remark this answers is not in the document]
jerome-benoit: > If there is a dedicated collection, the write in the 'settings' and in 'sensitivedatamigrations' must be done in a transaction otherwise you can manage writing the settings and fail writing the sensitivedatamigrations.

Both actions can be considered independents: 
+ in one collection, you only store the settings
+ in other collection, you keep track of the migration statuses in a transactional manner (at at the beginning: self contained data structure to restart a transaction for debugging)   

> I think tracking the changes only in settings collection should sufficient and would spare the code to handle the sensitivedatamigrations.

It can help to have a separate collection to permit to keep a migration history per tenant for debug purpose. I've just took the safe design road on that migration. I don't mind to only have one record history in settings but since the code is already there, not sure it worth it. 

You've scheduled a meeting with Stefania. So I assume you took over its development framing.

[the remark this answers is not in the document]
LucasBrazi06: Even if the code is already there, it doesn't mean we have to keep it only for that reason.
As usual, I'm the final reviewer but I need your feedback on the choices that have been taken which Stefania knows I think, thus the meeting.
I'll get back to you to keep you in sync.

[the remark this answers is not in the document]
jerome-benoit: > Even if the code is already there, it doesn't mean we have to keep it only for that reason.

The main reason is the separation of the data used by different process: one for the cyphering settings handling and one for the migration states. The settings determine if a migration shall be done and then the migration process starts and write its own data separately. The historisation is a design bonus. As well as not keeping sensitive informations ciphered strings close to their keys and settings once the migration states are expunged from the transactional fields used for debug. The deletion of the migration states was planned at the beginning once the migration was successful to permit to keep the transactional extra fields in case of errors or issue while not opening severe security issues as a tradeoff.    

I think you have the informations needed.

stefania-santimbrean:
Just a quick note: Please keep this PR here as I'm looking at all the comments when doing the other PR for each step agreed in the meeting and sent on email.

PR #1975, opened 2020-12-12
Update non-existing OCPP parameter + Fixed typos

LucasBrazi06:


src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -1776,7 +1772,7 @@ export default class OCPPUtils {
           source: chargingStation.id,
           action: ServerAction.CHARGING_STATION_CHANGE_CONFIGURATION,
           module: MODULE_NAME, method: 'updateChargingStationTemplateOcppParameters',
-          message: `Error in changing OCPP Parameter '${ocppParameter.key}' from '${currentOcppParam.value}' to '${ocppParameter.value}'`,
+          message: `Error in changing OCPP Parameter '${ocppParameter.key}' from '${currentOcppParam?.value}' to '${ocppParameter.value}'`,
           detailedMessages: { error: error.message, stack: error.stack }
         });
       }
```
jerome-benoit, line 1778: I do not understand or see here how the unknown OCPP parameter in DB is forced to be pushed and  flagged as a custom parameter in that PR. 
I think it's only force pushed here, right?

jerome-benoit (commented):
Just need more explanation on how to make unknown OCPP parameter custom.

src/server/ocpp/utils/OCPPUtils.ts
LucasBrazi06: Yep, I force the push only.
I consider 'custom' only the OCPP params set by the end-user using the UI that we want to keep visible.
When applied by the templates, I just want the value to be set, and it can be kept invisible (goal is to have a charger well configured).
If the user wants these params public, he can follow the Delta procedure.
Gerald asked them to make these params on Delta visible by default.

jerome-benoit: > Yep, I force the push only.

If the error of that kind of OCPP parameter configuration change is properly handled and reported, that's then fine. 
Tracking of obsolete or deprecated OCPP parameters in templates is somewhat less easy with that PR. 
Could we add a special logging for non existent paramater that also fails to be force pushed?  
That way, deprecated parameter tracking will be more easy: non existing + change configuration success != non existing + change configuration failure.

LucasBrazi06: > Tracking of obsolete or deprecated OCPP parameters in templates is somewhat less easy with that PR.
> Could we add a special logging for non existent paramater that also fails to be force pushed?
> That way, deprecated parameter tracking will be more easy: non existing + change configuration success != non existing + change configuration failure.

Yes, I logged a warning when the param is not found in the DB and if a failure occurs, then the usual error is logged. So we can recognize failure on non existing OCPP param: warning + error.

jerome-benoit: But the change configuration error log message does not tell if it happens on an non existing parameter or not. 
I thought the log warning on non existing parameter can be removed and the log message could only differ a bit when the change configuration fails or success by adding  information on the parameter existence, like: 
...
```
let occpParameterExist = true;
if (!currentOcppParam) {
	occpParameterExist = false;
```
...
```
`message: `${!occpParameterExist && 'Non existent'} OCPP Parameter '${ocppParameter.key}' has been successfully set from '${currentOcppParam?.value}' to '${ocppParameter.value}'`
```
... 

That way, less log entries and more template debug information in them. 
You see?

LucasBrazi06: Yes, agree, there are too many logs in the DB ;-).
I'll proceed with the changes.

jerome-benoit (approved):
I have no more comments to add and the PR already looked Ok anyway, just small nitpick.

PR #1989, opened 2020-12-22
Smart Charging - set sticky buffer to 20 percent

OliveGerste:


LucasBrazi06:
@OliveGerste Maybe this would be good also to provide this value in the Smart Charging's settings along the checkbox to activate this feature ;-).

OliveGerste:
> @OliveGerste Maybe this would be good also to provide this value in the Smart Charging's settings along the checkbox to activate this feature ;-).

Good idea! 
I am on it :)

PR #2009, opened 2021-01-08
Assets add missing properties

OliveGerste:


src/types/Asset.ts
```diff
@@ -18,6 +20,12 @@ export default interface Asset extends CreatedUpdatedProps, AbstractCurrentConsu
   values: Consumption[],
 }
 
+export enum AssetType {
+  CO = 'CO',
+  PR = 'PR',
+  CO_PR = 'CO-PR',
+}
```
LucasBrazi06, line 27: It would be nice also to have full string keys for readability?

  CONSUME = 'CO',
  PRODUCE = 'PR',
  CONSUME_PRODUCE = 'CO-PR',

As you want as I had already accepted your proposal ;-)

LucasBrazi06 (approved):

LucasBrazi06:
@OliveGerste One remark is that variable name 'fallbackValue' is only valid if you have a dynamic asset but not when you have a static one (no connection to a backend to get the value).
I would call it more like 'staticValueWatt' (with the Unit also ;-))
Also 'fluctuation' -> 'fluctuationPercent'?
Let me know.

PR #2015, opened 2021-01-11
Clear the locks by hostname at server startup

stefania-santimbrean:


src/utils/Utils.ts
```diff
@@ -1345,4 +1347,8 @@ export default class Utils {
   public static isPlateIDValid(plateID): boolean {
     return /^[A-Z0-9- ]*$/.test(plateID);
   }
+
+  public static getHostname(): string {
```
jerome-benoit, line 1351: That helper can also be used in the following files: 
+ LockingManager
+ LoggingStorage

src/locking/LockingManager.ts
```diff
@@ -82,6 +82,11 @@ export default class LockingManager {
     return true;
   }
 
+  public static async cleanUpLocks(): Promise<void> {
```
jerome-benoit, line 85: To ensure that code will only be run on SCP, you can for example pass a boolean defaulting to true and guard the locks cleanup with it. Or guard the cleanups with `if (Configuration.isCloudFoundry() || Utils.isDevelopmentEnv())`. The latter have side effects but since it's on boolean variables, that's not an issue. I prefer the boolean as argument, so the code readers know the condition without have to read the method.  
Nitpick: cleanup is one word, so its use as camel case is more cleanupLocks.

jerome-benoit (changes requested):

PR #2032, opened 2021-01-13
Type and fix OCPP error handling.

jerome-benoit:

### DIFF
--- a/src/exception/OcppError.ts
+++ b/src/exception/OcppError.ts
@@ -3,6 +3,9 @@
  * Attributes name are part of the OCPP specification
  */
 export default class OCPPError extends Error {
+  code: string;
+  details?: any;
+
   public constructor(readonly params: {
     source: string;
     module: string;
@@ -12,6 +15,8 @@ export default class OCPPError extends Error {
     details?: any;
   }) {
     super(params.message);
+    this.code = params.code;
+    this.details = params.details;
 
     Object.setPrototypeOf(this, OCPPError.prototype); // For instanceof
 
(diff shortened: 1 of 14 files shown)


src/exception/OcppError.ts
```diff
@@ -3,6 +3,9 @@
  * Attributes name are part of the OCPP specification
  */
 export default class OCPPError extends Error {
+  code: string;
```
LucasBrazi06, line 6: Visibility private?

LucasBrazi06 (approved):

src/exception/OcppError.ts
jerome-benoit: It can't, it's used to properly build the error message to send: 

`messageToSend = JSON.stringify([messageType, messageId, commandParams.code ? commandParams.code : OCPPErrorType.GENERIC_ERROR, commandParams.message ? commandParams.message : '', commandParams.details ? commandParams.details : {}]);`

The code was assigning default values that was hiding the issue: the code never sent the exact error. 
That's because we still have any type usage. It avoids to catch missing but required attributes like here. I've tried to reduce it in important code paths lately like the WebSockets handling or OCPP messages handling, but that's still not enough to catch mistyping that reveal some bugs.

LucasBrazi06: Then public ? :-)
Visibility must be po

LucasBrazi06: Then public ? :-)
Visibility must be po

LucasBrazi06: I don't see where OCPPError is used in your example.
If you can access instance members in class OCPPError, make them public.
Otherwise encapsulation must be applied with getters and make instance members private.

jerome-benoit: sendMessage() signature: 

`public async sendMessage(messageId: string, commandParams: Record<string, unknown> | OCPPError, ... `

and commandParams is used to build the message string to send. 

I can make the OCPP attributes explicitly public instead of implicitly.

LucasBrazi06: > I can make the OCPP attributes explicitly public instead of implicitly.

YA :-)
I think there is a tslint setting for that right? :-)

jerome-benoit:
> I think there is a tslint setting for that right? :-)
> 
It was enabled but was making the code difficult to read because of the highlights. I think you have removed the setting or asked to remove it.

--
Jérôme Benoit aka fraggle
OpenPGP Key ID : 27B535D3

LucasBrazi06:
> It was enabled but was making the code difficult to read because of the highlights. I think you have removed the setting or asked to remove it.

Ok but I don't think this was for the visibility. You can put it back :-).

PR #2033, opened 2021-01-14
01. Crypto - Migrate crypto key from config file to database

stefania-santimbrean:
Questions:
- Unit Test corrected fine? (createContext needs to be run before test so the crypto key gets generated for new tenants)
- All dependencies of Cypher.ts covered?
- To check about key generation length: 16 or 32?
- operation mode: CTR changed to GCM?

(Please ignore previous PR #2030, I went back few commits and created new branch)


src/integration/billing/stripe/StripeBillingIntegration.ts
```diff
@@ -694,14 +691,14 @@ export default class StripeBillingIntegration extends BillingIntegration<StripeB
...
-        source: Constants.CENTRAL_SERVER,
-        module: MODULE_NAME, method: 'checkIfStripeIsInitialized',
-        action: ServerAction.CHECK_CONNECTION,
-        message: 'No connection to Stripe available'
-      });
+  private async initializeStripe() {
+    this.settings.currency = this.stripeSettings.currency;
+    // Get Crypto Setting
+    const cryptoSetting = await Cypher.getCryptoSetting(this.tenantID);
+    if (this.settings.secretKey) {
```
jerome-benoit, line 698: I would prefer that you trap the missing key a lower level and throw an error. It's a requirement to have a key set at the tenant level for e-mobility to work.

src/migration/tasks/MigrateCryptoSettingsFromConfigToDBTask.ts
```diff
@@ -0,0 +1,50 @@
...
+    for (const tenant of tenants.result) {
+      await this.migrateTenant(tenant);
+    }
+  }
+
+  public async migrateTenant(tenant: Tenant): Promise<void> {
+    // Crypto Key from config file
+    const configCryptoKey: string = Configuration.getCryptoConfig().key;
+    // Crypto Key Properties from config file
+    const confiCryptoKeyProperties: CryptoKeyProperties = Utils.separateCryptoAlgo(Configuration.getCryptoConfig().algorithm);
```
jerome-benoit, line 25: confi'g'CryptoKeyProperties

src/utils/Utils.ts
```diff
@@ -1341,6 +1341,32 @@ export default class Utils {
     return /^[A-Z0-9- ]*$/.test(plateID);
   }
 
+  public static separateCryptoAlgo(algo: string): CryptoKeyProperties {
+    const regExp = /(.*?)-(.*?)-(.*)/.exec(algo);
```
jerome-benoit, line 1345: Just split the string with `algo.split('-')`. With object destructuring to array you can have in one line: 

`[blockCypher, blockSize, operationMode] = algo.split('-')`

Regexp is overkill here and have performance drawbacks than simple string handling methods do not exhibit.

src/utils/Utils.ts
```diff
@@ -1341,6 +1341,32 @@ export default class Utils {
     return /^[A-Z0-9- ]*$/.test(plateID);
   }
 
+  public static separateCryptoAlgo(algo: string): CryptoKeyProperties {
```
jerome-benoit, line 1344: parseCryptoSettings?
parseConfigCryptoSettings?

src/utils/Utils.ts
```diff
@@ -1341,6 +1341,32 @@ export default class Utils {
...
+  public static buildAlgorithm(properties: CryptoKeyProperties): string {
+    return `${properties.blockCypher.toLowerCase()}-${properties.blockSize}-${properties.operationMode.toLowerCase()}`;
+  }
+
+  public static generateKey(): string {
+    // TODO change 16 to 32 and test on Mac
+    return crypto.randomBytes(16).toString('hex');
+  }
+
+  public static getKeyProperties(): CryptoKeyProperties {
```
jerome-benoit, line 1362: It's the default properties. I think the method name should reflect it.

test/api/CypherTest.ts
```diff
@@ -16,15 +17,24 @@ chai.use(responseHelper);
 
 const FAKE_JSON: CypherJSON = { 'sensitiveData': ['content.secret1', 'content.secret2'], 'content': { 'secret1': 'Harry', 'secret2': 'Potter' } };
 const FAKE_WORD = 'Wingardium Leviosa';
-const FAKE_WORD_ENCRYPTED = '73d49673e8f70de6e16ba99bc0c273da:8b9cb41f960d2b50604921a68d7d0382472d';
+const FAKE_WORD_ENCRYPTED = '04480e7ab47eb5729faea20f0844dffa:81927d0bc660c4548b27cc0e3427d6874ea3';
```
jerome-benoit, line 20: I think that unit test is not relevant anymore: its purpose was to detect a key change and the goal of your work is to allow key to change. Newly created tenant already have a random key. 
So I think you can just delete it. 
Unit tests is relevant to test that the key migration code work properly.

src/utils/Cypher.ts
```diff
@@ -2,43 +2,34 @@ import BackendError from '../exception/BackendError';
...
-          source: Constants.CENTRAL_SERVER,
-          module: MODULE_NAME,
-          method: 'getConfiguration',
-          message: 'Crypto configuration is missing'
-        });
-      }
-    }
-    return this.configuration;
+  public static async getCryptoSetting(tenantID: string): Promise<CryptoSetting> {
+    return (await SettingStorage.getCryptoSettings(tenantID)).crypto;
```
jerome-benoit, line 17: Please keep the error throwing if no settings are found. 
E-mobility can't work otherwise and it shall not happen.

jerome-benoit (changes requested):
I think you can already make Cypher class tenant aware: 

+ Cypher: encrypt/decrypt(this.connection.connection.password, tenantID);
+ Cypher: getCryptoSetting(this.tenantID) becomes a private method called inside the class.

src/utils/Utils.ts
stefania-santimbrean: parseConfigCryptoAlgorithm seems more appropiate since I give as input the algorithm from the Crypto Config file and return Crypto Key Properties and not the whole Crypto Setting

test/api/CypherTest.ts
stefania-santimbrean: I delete it then and create add in SettingTest.ts a condition if crypto setting is in DB

src/utils/Cypher.ts
```diff
@@ -13,22 +11,20 @@ const MODULE_NAME = 'Cypher';
...
 export default class Cypher {
 
-  public static async getCryptoSetting(tenantID: string): Promise<CryptoSetting> {
-    return (await SettingStorage.getCryptoSettings(tenantID)).crypto;
-  }
-
-  public static encrypt(data: string, cryptoSetting: CryptoSetting): string {
+  public static async encrypt(data: string, tenantID: string): Promise<string> {
     const iv = crypto.randomBytes(IV_LENGTH);
+    const cryptoSetting = await this.getCryptoSetting(tenantID);
```
jerome-benoit, line 16: It's a static method, use class name instead of this.

src/integration/asset/schneider/SchneiderAssetIntegration.ts
```diff
@@ -127,9 +127,7 @@ export default class SchneiderAssetIntegration extends AssetIntegration<AssetSet
     const params = new URLSearchParams();
     params.append('grant_type', 'password');
     params.append('username', this.connection.connection.user);
-    // Get Crypto Setting
-    const cryptoSetting = await Cypher.getCryptoSetting(this.tenantID);
-    params.append('password', Cypher.decrypt(this.connection.connection.password, cryptoSetting));
+    params.append('password', (await Cypher.decrypt(this.connection.connection.password, this.tenantID)));
```
jerome-benoit, line 130: You like parenthesis ;) 
They are not needed here, there's already an argument separator. And also in other changes in that PR.

src/utils/Cypher.ts
```diff
@@ -13,22 +11,20 @@ const MODULE_NAME = 'Cypher';
...
     encryptedData = Buffer.concat([encryptedData, cipher.final()]);
     return iv.toString('hex') + ':' + encryptedData.toString('hex');
   }
 
-  public static decrypt(data: string, cryptoSetting: CryptoSetting): string {
+  public static async decrypt(data: string, tenantID: string): Promise<string> {
     const dataParts = data.split(':');
     const iv = Buffer.from(dataParts.shift(), 'hex');
     const encryptedData = Buffer.from(dataParts.join(':'), 'hex');
+    const cryptoSetting = await this.getCryptoSetting(tenantID);
```
jerome-benoit, line 27: It's a static method, use class name instead of this.

src/utils/Utils.ts
```diff
@@ -1341,12 +1341,12 @@ export default class Utils {
...
-  public static separateCryptoAlgo(algo: string): CryptoKeyProperties {
-    const regExp = /(.*?)-(.*?)-(.*)/.exec(algo);
+  public static parseConfigCryptoAlgorithm(algo: string): CryptoKeyProperties {
+    const [blockCypher, blockSize, operationMode] = algo.split('-');
     return {
-      blockCypher: regExp[1].toUpperCase(),
-      blockSize: parseInt(regExp[2]),
-      operationMode: regExp[3].toUpperCase()
+      blockCypher: blockCypher,
+      blockSize: parseInt(blockSize),
```
jerome-benoit, line 1348: Please use Utils.ConvertToInt() helper instead.

src/utils/Cypher.ts
```diff
@@ -136,4 +132,17 @@ export default class Cypher {
...
   }
+
+  private static async getCryptoSetting(tenantID: string): Promise<CryptoSetting> {
+    const cryptoSettings = (await SettingStorage.getCryptoSettings(tenantID)).crypto;
+    if (!cryptoSettings) {
+      throw new BackendError({
+        source: Constants.CENTRAL_SERVER,
+        module: MODULE_NAME,
+        method: 'getCryptoSetting',
+        message: 'No Crypto Settings found in the database'
```
jerome-benoit, line 143: Please use the tenant ID information to build a more informative error message with the tenant name and/or its ID.

src/utils/Utils.ts
```diff
@@ -1341,12 +1341,12 @@ export default class Utils {
...
 
-  public static separateCryptoAlgo(algo: string): CryptoKeyProperties {
-    const regExp = /(.*?)-(.*?)-(.*)/.exec(algo);
+  public static parseConfigCryptoAlgorithm(algo: string): CryptoKeyProperties {
+    const [blockCypher, blockSize, operationMode] = algo.split('-');
     return {
-      blockCypher: regExp[1].toUpperCase(),
-      blockSize: parseInt(regExp[2]),
-      operationMode: regExp[3].toUpperCase()
+      blockCypher: blockCypher,
```
jerome-benoit, line 1347: You have to be consistent regarding the ciphering settings casing.

src/utils/Utils.ts
```diff
@@ -1359,7 +1359,7 @@ export default class Utils {
     return crypto.randomBytes(16).toString('hex');
   }
 
-  public static getKeyProperties(): CryptoKeyProperties {
+  public static getDefaultKeyProperties(): CryptoKeyProperties {
     return {
       blockCypher: 'AES',
```
jerome-benoit, line 1364: You have to be consistent regarding the ciphering settings casing. Store them uppercased or lowercased. You can display them uppercased on the dashboard only and store them lowercased.

jerome-benoit (changes requested):

jerome-benoit:
> Questions:

> * operation mode: CTR changed to GCM?

Not yet. That's not the goal of that migration.

src/utils/Cypher.ts
```diff
@@ -145,4 +132,17 @@ export default class Cypher {
...
   }
+
+  private static async getCryptoSetting(tenantID: string): Promise<CryptoSetting> {
+    const cryptoSettings = (await SettingStorage.getCryptoSettings(tenantID)).crypto;
+    if (!cryptoSettings) {
+      throw new BackendError({
+        source: Constants.CENTRAL_SERVER,
+        module: MODULE_NAME,
+        method: 'getCryptoSetting',
+        message: `Tenant with ID: ${tenantID} does not have crypto settings.`
```
jerome-benoit, line 143: You can get the name with `TenantStorage.getTenant(tenantID).name`. The tenantID is not really a human readable output to log.

src/storage/mongodb/TenantStorage.ts
```diff
@@ -121,6 +121,8 @@ export default class TenantStorage {
     const uniqueTimerID = Logging.traceStart(tenantID, MODULE_NAME, 'createTenantDB');
     // Create tenant collections
     await global.database.checkAndCreateTenantDatabase(tenantID);
+    // Create initial settings for tenant
+    await global.database.createInitialSettingsForTenant(tenantID);
```
jerome-benoit, line 125: I think putting data inside the DB is not targeted to the DB helpers. 
The right place is in TenantService after `await TenantStorage.createTenantDB(filteredRequest.id);`.

src/integration/billing/stripe/StripeBillingIntegration.ts
```diff
@@ -29,21 +29,17 @@ export default class StripeBillingIntegration extends BillingIntegration<StripeB
...
-    if (this.settings.secretKey) {
-      this.settings.secretKey = Cypher.decrypt(settings.secretKey);
-    }
-    // Currently the public key is not encrypted
-    this.stripe = new Stripe(this.settings.secretKey);
+    this.stripeSettings = settings;
   }
 
   public async checkConnection(): Promise<void> {
     // Check Stripe
```
jerome-benoit, line 41: Nitpick: you can change also the comment to match the method purpose.

jerome-benoit (changes requested):
It's probably the latest review before my approval. 

Thanks.

src/utils/Cypher.ts
LucasBrazi06: Yep but a call in the DB just to get the name...
Type to pass the tenant object instead the ID.

stefania-santimbrean: @LucasBrazi06 not sure what you mean by passing the tenant object?

jerome-benoit:
> +
> +  private static async getCryptoSetting(tenantID: string): Promise<CryptoSetting> {
> +    const cryptoSettings = (await SettingStorage.getCryptoSettings(tenantID)).crypto;
> +    if (!cryptoSettings) {
> +      throw new BackendError({
> +        source: Constants.CENTRAL_SERVER,
> +        module: MODULE_NAME,
> +        method: 'getCryptoSetting',
> +        message: `Tenant with ID: ${tenantID} does not have crypto settings.`
> Yep but a call in the DB just to get the name...
> Type to pass the tenant object instead the ID.
> 
> 

Since the information we have before calling Cypher class methods is the tenantID, requiring a Tenant instance will just require similar calls to the DB that could have been avoided before ciphering operations.
The code path here is an error path that should never happen. And if it happens, I think having the tenant name along with the tenantID in the message will ease the debug.  But we can stay with the tenantID alone, that’s fine.


--
Jérôme Benoit aka fraggle
OpenPGP Key ID : 27B535D3

src/utils/Cypher.ts
```diff
@@ -136,11 +137,12 @@ export default class Cypher {
   private static async getCryptoSetting(tenantID: string): Promise<CryptoSetting> {
     const cryptoSettings = (await SettingStorage.getCryptoSettings(tenantID)).crypto;
     if (!cryptoSettings) {
+      const tenantName = (await TenantStorage.getTenant(tenantID)).name;
       throw new BackendError({
         source: Constants.CENTRAL_SERVER,
         module: MODULE_NAME,
         method: 'getCryptoSetting',
-        message: `Tenant with ID: ${tenantID} does not have crypto settings.`
+        message: `Tenant with ID: ${tenantName} does not have crypto settings.`
```
jerome-benoit, line 145: Please read log message you change and ensure the information given are understandable : if you switch to tenant name, it's no more the tenantID. 
I think you can log both `Tenant ${tenantName} (tenandID: ${tenantID}) does not have crypto settings`.

src/utils/Cypher.ts
LucasBrazi06: To replace this:
`private static async getCryptoSetting(tenantID: string): Promise<CryptoSetting> {`
by this:
`private static async getCryptoSetting(tenant: Tenant): Promise<CryptoSetting> {`

But this method is only containing DB access so it's useless, DB method should be called right away.

stefania-santimbrean: But there's no need to call tenant storage unless there's an error thrown. If the crypto setting is there and we call tenant storage to get tenant it is an unnecessary call to db.

LucasBrazi06: This can be handled outside this method I think.

jerome-benoit: > But there's no need to call tenant storage unless there's an error thrown. If the crypto setting is there and we call tenant storage to get tenant it is an unnecessary call to db.

Just revert the change I've requested (for nothing) to the previous state: tenantID in the error message. That will be enough and do not require extra DB query.

stefania-santimbrean: Okay, back to just tenant id now

jerome-benoit (approved):

src/storage/mongodb/SettingStorage.ts
```diff
@@ -304,6 +304,50 @@ export default class SettingStorage {
...
+  }
+
+  public static async getCryptoSettings(tenantID: string): Promise<KeySetting> {
+    // Get the Crypto Key settings
+    const settings = await SettingStorage.getSettings(tenantID,
+      { identifier: TenantComponents.CRYPTO },
+      Constants.DB_PARAMS_MAX_LIMIT);
+    if (settings.count > 0) {
+      const cryptoSetting = {
+        formerKey: settings.result[0].content.crypto.formerKey,
```
jerome-benoit, line 328: You should test for the formerKey attribute existence in the result before assignin: 

`settings.result[0].content.crypto.formerKey ? settings.result[0].content.crypto.formerKey : null`

And do the same for other formerKey related settings.

jerome-benoit (changes requested):

src/storage/mongodb/SettingStorage.ts
```diff
@@ -304,6 +304,50 @@ export default class SettingStorage {
...
+  public static async getCryptoSettings(tenantID: string): Promise<KeySetting> {
+    // Get the Crypto Key settings
+    const settings = await SettingStorage.getSettings(tenantID,
+      { identifier: TenantComponents.CRYPTO },
+      Constants.DB_PARAMS_MAX_LIMIT);
+    if (settings.count > 0) {
+      const cryptoSetting = {
+        formerKey: settings.result[0].content.crypto.formerKey && settings.result[0].content.crypto.formerKey,
+        formerKeyProperties: {
+          blockCypher: settings.result[0].content.crypto.formerKeyProperties && settings.result[0].content.crypto.formerKeyProperties?.blockCypher,
```
jerome-benoit, line 330: You test for `settings.result[0].content.crypto.formerKeyProperties` before the assignation, so `settings.result[0].content.crypto.formerKeyProperties?.blockCypher` does not need to test it again (the ? is a typescript syntactic sugar for that test). 

But the correct test before that variable value assignation is: 

`settings.result[0].content.crypto.formerKeyProperties?.blockCypher && settings.result[0].content.crypto.formerKeyProperties.blockCypher`

The same goes for all variable value assignation here: the interface that define the data structure make some fields optional, so testing for their presence is a requirement before using it.

jerome-benoit (changes requested):

LucasBrazi06 (approved):
- This branch was not kept in sync with master-qa -> Merge the other branches with latest version of master-qa 
- operationMode with wrong value -> Cannot decypher data! (occurs when creating a new tenant)! -> Have you tested this scenario?
- We try to set the tenantID at first pos in input params of methods (Cypher class)
- Miscs fixes

PR #2037, opened 2021-01-18
REST Authentication documentation + schemas

Ninjeneer:


src/assets/server/rest/v1/schemas/auth/auth-check-eula.json
```diff
@@ -0,0 +1,16 @@
+{
```
LucasBrazi06, line 1: File not used I think: to remove?

src/assets/server/rest/v1/schemas/auth/auth-eula.json
```diff
@@ -0,0 +1,12 @@
+{
```
LucasBrazi06, line 1: File not used I think: to remove?

src/assets/server/rest/v1/schemas/auth/auth-resend-verification-email.json
```diff
@@ -0,0 +1,20 @@
+{
```
LucasBrazi06, line 1: File not used I think: to remove?

src/assets/server/rest/v1/schemas/auth/auth-reset-password.json
```diff
@@ -0,0 +1,38 @@
+{
```
LucasBrazi06, line 1: File not used I think: to remove?

src/assets/server/rest/v1/schemas/auth/auth-signin.json
```diff
@@ -0,0 +1,23 @@
+{
```
LucasBrazi06, line 1: File not used I think: to remove?

src/assets/server/rest/v1/schemas/auth/auth-signon.json
```diff
@@ -0,0 +1,48 @@
+{
```
LucasBrazi06, line 1: File not used I think: to remove?

src/assets/server/rest/v1/schemas/auth/auth-verify-email.json
```diff
@@ -0,0 +1,21 @@
+{
```
LucasBrazi06, line 1: File not used I think: to remove?

LucasBrazi06 (changes requested):
Lots of files to remove I think.
The OAS is fine

LucasBrazi06:
@ClaudeROSSI You can also have a look at this PR as you know OAS also :-)

src/assets/server/rest/v1/schemas/auth/auth-check-eula.json
Ninjeneer: It is

AuthValidtion.ts
```ts
this._authCheckEula = JSON.parse(fs.readFileSync(`${global.appRoot}/assets/server/rest/v1/schemas/auth/auth-check-eula.json`, 'utf8'));
```

AuthService.ts
```ts
filteredRequest = AuthValidator.getInstance().validateAuthCheckEula(req.query);
```

Same for the others below

LucasBrazi06: Ah ok, it for JSon schema validator, nothing to do with the OAS in that case, right?

Ninjeneer: Right

src/assets/server/rest/v1/schemas/auth/auth-eula.json
```diff
@@ -0,0 +1,12 @@
...
+  "title": "Eula Request",
+  "type": "object",
+  "properties": {
+    "Language": {
+      "$ref": "common.json#/definitions/language"
+    },
+    "tenant": {
```
LucasBrazi06, line 8: Why do we have the param tenant's first letter in lower case here and not the language?

LucasBrazi06 (changes requested):
Typo on one param?

src/assets/server/rest/v1/schemas/auth/auth-eula.json
Ninjeneer: I followed the way AuthSecurtiy gather informations
`Language` param starts with an uppercase letter because it comes from Query, and `tenant` param comes from the header of the request. I guess there is a technical choice behind this ? Otherwise, I could just move tenant param to Query, tell me

LucasBrazi06: @jerome-benoit Could you have a look?

jerome-benoit: > @jerome-benoit Could you have a look?

If it's a POST request on an endpoint, all data must come in the body or the headers.
If it's a GET request on an endpoint, query string should be avoided in favor of URI with parameters parsed: they then can be case sensitive, unlike in the query string part, among one of numerous advantages. 
restapi.com/authors/:id -> author with id
restapi.com/authors/:name -> authors with name
restapi.com/authors/:town -> authors in the town
restapi.com/authors/:id/town -> authors town with id
etc.

src/assets/server/rest/v1/docs/e-mobility-oas.json
```diff
@@ -277,32 +292,35 @@
         "description": "Register",
         "requestBody": {
           "content": {
-            "application/x-www-form-urlencoded": {
```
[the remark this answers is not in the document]
jerome-benoit: > I put the type to 'x-www-form-urlencoded' because it was more user friendly in Swagger to provide data.

If the data input expect JSON format, it's not a correct change: the API documentation must give the exact data type and format expected on an endpoint, not be user friendly. It's main target is developers and the data expected must the same as the one they have to implement.

src/assets/server/rest/v1/schemas/auth/auth-eula.json
LucasBrazi06: If it's related to the GET Eula, the tenant is deduced from the URL and should be passed in the Query string and named Tenant.

src/assets/server/rest/v1/docs/e-mobility-oas.json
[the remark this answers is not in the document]
LucasBrazi06: Yes you're right, so let's stick to the standard then.
@Ninjeneer Replace all application/x-www-form-urlencoded by application/json.
Thanks!

src/server/rest/v1/validator/AuthValidation.ts
```diff
@@ -0,0 +1,76 @@
+import { HttpCheckEulaRequest, HttpEulaRequest, HttpLoginRequest, HttpRegisterUserRequest, HttpResendVerificationMailRequest, HttpResetPasswordRequest, HttpVerifyEmailRequest } from '../../../../types/requests/HttpUserRequest';
+
+import SchemaValidator from './SchemaValidator';
+import fs from 'fs';
+import global from '../../../../types/GlobalType';
+
+export default class AuthValidator extends SchemaValidator {
+  private static _instance: AuthValidator | undefined;
+  private _authSignIn: any;
```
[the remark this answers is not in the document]
Ninjeneer: >  And starting attributes name with underscore is redundant with the private scope.

Indeed, I removed it

> Is it hard to define a proper type here?

Yes it is, as schemas can have multiple complex structures, with custom keywords at any level. Even AJV module itself doesn't type schemas. However, in order to remove the any type, I created a Schema that can handle barely any type of schemas. Not really relevant, but better than `any` type.
```ts
export default interface Schema {
  [properties: string]: unknown;
  id: string;
  type: string;
}
```

src/assets/server/rest/v1/docs/e-mobility-oas.json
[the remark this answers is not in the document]
Ninjeneer: Done

src/assets/server/rest/v1/schemas/auth/auth-eula.json
Ninjeneer: Done

jerome-benoit: > If it's related to the GET Eula, the tenant is deduced from the URL and should be passed in the Query string and named Tenant.

Query string are case insensitive. A more correct GET URI is /eula/.../:tenantID

The new REST API shall never ever use query string, only well constructed URIs with parameters inside, each of them matching an express router. That's the whole purpose of having made a hierarchical express routers tree: being able to build meaningful URI to query in a REST standardised manner.

LucasBrazi06: Just realized that the EULA is not tenant relevant, so why are we talking about passing it in the end :) !?
Only the Language is relevant here and can be a query string as this is not an identifier.

Ninjeneer: Therefore, should we only use eulas stored in the default tenant ? Currently, each tenant has its own `eula` collection, that's why we are using tenant identifier here

Ninjeneer: Done, removed tenant related parameters/attributes

LucasBrazi06 (approved):

PR #2040, opened 2021-01-18
Translate every csv headers

cabch:
changed every csv headers to be translated
erased all template strings for csv related content
added userLocal to OCPPParam type


src/assets/i18n/fr.json
```diff
@@ -1,4 +1,60 @@
 {
+  "billing": {
+    "generatedUser": "Utilisateur généré pour '{{email}}'",
+    "chargingStopSiteArea": "Charging Station {{totalConsumption}} kWh à la borne {{siteArea}} (terminé à {{time}})",
```
ClaudeROSSI, line 4: Message does not match the english version
I would suggest:

"Chargement de {{totalConsumption}} kWh sur le site {{siteArea}} ........

src/assets/i18n/fr.json
```diff
@@ -1,4 +1,60 @@
...
+    "ocppProtocol": "Protocol OCPP",
+    "ocppVersion": "Version OCPP",
+    "powerLimitUnit": "Unité de limite de puissance",
+    "timezone" : "Timezone de la Borne",
+    "vendor": "Vendeur"
+  },
+  "general": {
+    "changedBy": "Changé Par",
+    "changedOn": "Changé Le",
+    "createdOn": "Créé le",
```
ClaudeROSSI, line 27: In french, we never have characters un Uppercase!

sauf pour les nom propres et des trucs comme Roi, Président, etc

bref:
Changé par
Changé le

src/assets/i18n/en.json
```diff
@@ -1,4 +1,60 @@
...
+    "chargeBoxSN": "Charge Box S/N",
+    "chargePointSN": "Charge Point S/N",
+    "chargingStation": "Charging Station",
+    "connector": "Connector",
+    "firmwareVersion": "Firmware Version",
+    "lastReboot": "Last Reboot",
+    "lastSeen": "Last Seen",
+    "maxPower": "Maximum Power (Watt)",
+    "model": "Model",
+    "numberOfConnectors": "Number Of Connectors",
```
ClaudeROSSI, line 17: Rules in English, 
The rules are fairly standard for title case: Capitalize the first and the last word. Capitalize nouns, pronouns, adjectives, verbs, adverbs, and subordinate conjunctions. Lowercase articles (a, an, the), coordinating conjunctions, and prepositions.

so: of has to be lowercase!

src/server/rest/v1/service/ChargingStationService.ts
```diff
@@ -1331,31 +1337,51 @@ export default class ChargingStationService {
...
+      csv += (chargingStation.connectors ? chargingStation.connectors.length : '0') + Constants.CSV_SEPARATOR;
+      csv += (chargingStation?.siteArea?.name ? chargingStation.siteArea.name : '') + Constants.CSV_SEPARATOR;
       if (chargingStation.coordinates && chargingStation.coordinates.length === 2) {
-        csv += `${chargingStation.coordinates[1]}` + Constants.CSV_SEPARATOR;
-        csv += `${chargingStation.coordinates[0]}` + Constants.CSV_SEPARATOR;
+        csv += chargingStation.coordinates[1] + Constants.CSV_SEPARATOR;
+        csv += chargingStation.coordinates[0] + Constants.CSV_SEPARATOR;
       } else {
-        csv += `''${Constants.CSV_SEPARATOR}''`;
+        csv += '' + Constants.CSV_SEPARATOR + '';
```
ClaudeROSSI, line 1372: two values if we enter the "if" case - and only one value in the "else" case? 
Is this normal?

It should be sufficient to do the following. right?
csv +=  Constants.CSV_SEPARATOR + Constants.CSV_SEPARATOR;

src/types/ChargingStation.ts
```diff
@@ -280,6 +280,7 @@ export type OCPPParams = {
   siteAreaName: string;
   chargingStationName: string;
   params: OcppParameter[];
+  userLocale: string;
```
ClaudeROSSI, line 283: it sounds weird to have the user locale part of the OCPP params!

I would suggest to pass the request as parameter and then just rely on 
- req.user.locale

src/server/rest/v1/service/StatisticService.ts
```diff
@@ -473,36 +474,38 @@ export default class StatisticService {
   static convertToCSV(loggedUser: UserToken, transactionStatsMDB: any[], dataCategory: string, dataType: string, year: number | string, dataScope?: string): string {
     let user: User;
     let unknownUser = Utils.buildUserFullName(user, false, false);
+    const i18nManager = new I18nManager(loggedUser.locale);
```
ClaudeROSSI, line 477: can't we just use the same logic everywhere ?
Pass the request and use:
- req.user.locale

I would even suggest to have a request.getI18nManager() to avoid creating a new instance each time we need some translation! What do you think?

src/types/ChargingStation.ts
cabch: Yes wanted to add a comment. 

I don't know how to do differently yet, need to show you where it's called as I can't access the req at that place. 
You suggest to add the entire req as param ? 

Needed userLocale only to make it work temporarly, we definitely need to have a look to this.

src/server/rest/v1/service/ChargingStationService.ts
jerome-benoit: Please use string literal, it's more readable.

src/server/rest/v1/service/StatisticService.ts
jerome-benoit: > I would even suggest to have a request.getI18nManager() to avoid creating a new instance each time we need some translation! What do you think?

i18nManager class shall use the singleton design pattern, in the whole e-mobility codebase. I do know if that can be part of that PR for ev-server. 
It's a pretty easy change to do that will likely have a measurable impact on the memory consumption, especially on the dashboard.

src/types/ChargingStation.ts
LucasBrazi06: You can pass it to the method, OCPP is not treated with the same logic ;-)
image

src/server/rest/v1/service/StatisticService.ts
LucasBrazi06: > > I would even suggest to have a request.getI18nManager() to avoid creating a new instance each time we need some translation! What do you think?
> 
> i18nManager class shall use the singleton design pattern, in the whole e-mobility codebase. I do know if that can be part of that PR for ev-server.
> It's a pretty easy change to do that will likely have a measurable impact on the memory consumption, especially on the dashboard.

This was what Claude proposed :-)
Ok with that.

jerome-benoit:
> 
> i18nManager class shall use the singleton design pattern, in the whole e-mobility codebase. I do know if that can be part of that PR for ev-server.
> It's a pretty easy change to do that will likely have a measurable impact on the memory consumption, especially on the dashboard.
> 
> This was what Claude proposed :-)
> 

It could also be the object factory design pattern that might not always give you the same object instance ;)

--
Jérôme Benoit aka fraggle
OpenPGP Key ID : 27B535D3

LucasBrazi06:
@cabch  Put back to requested changes pipeline

src/assets/i18n/fr.json
cabch: changed :)

src/assets/i18n/en.json
cabch: changed :)

src/server/rest/v1/service/ChargingStationService.ts
jerome-benoit: > It should be sufficient to do the following. right?
> csv += Constants.CSV_SEPARATOR + Constants.CSV_SEPARATOR;

Otherwise the number of columns will be off by one if GPS coordinates are empty and the import in other tools will likely fail.

src/assets/i18n/fr.json
cabch: done :)

src/types/ChargingStation.ts
cabch: of ourse !! I changed it

src/server/rest/v1/service/ChargingStationService.ts
cabch: I let it as it is as said Jérôme we need 2 empty columns so the CSV is correctly build

src/server/rest/v1/service/ChargingStationService.ts
```diff
@@ -1310,9 +1309,9 @@ export default class ChargingStationService {
     return chargingStations;
   }
 
-  private static convertOCPPParamsToCSV(ocppParams: OCPPParams, writeHeader = true): string {
+  private static convertOCPPParamsToCSV(ocppParams: OCPPParams, userLocale: string, writeHeader = true): string {
     let csv = '';
-    const i18nManager = new I18nManager(ocppParams.userLocale);
```
cabch: @LucasBrazi06 is that what you meant by "You can pass it to the method" ?

src/utils/I18nManager.ts
```diff
@@ -26,6 +27,16 @@ export default class I18nManager {
...
+  // Get the already existing instance of I18nManager depending on the locale if it exists otherwise we create it
+  public static getInstanceForLocale(locale: string): I18nManager {
+    if (!I18nManager.instances.has(locale)) {
+      I18nManager.instances.set(locale, new I18nManager(locale));
+    }
+
+    return I18nManager.instances.get(locale);
+  }
+
+
```
cabch, line 39: @ClaudeROSSI @jerome-benoit @LucasBrazi06 here is how it ends up (now we call I18nManager. getInstanceForLocale(locale) and it either chose the existing instance or it creates it)

ClaudeROSSI (changes requested):
minor comments - some text to change

src/assets/i18n/fr.json
```diff
@@ -1,44 +1,100 @@
...
       "body": "Vous avez badgé sur la borne '{{chargeBoxID}}', mais aucune session n'a été démarrée"
     },
     "sessionStarted": {
-      "title": "Session Démarrée",
+      "title": "Session démarrée",
       "body": "Votre session sur la borne '{{chargeBoxID}}', connecteur '{{connectorId}}', a été démarrée avec succès dans l'organisation '{{tenantName}}'"
     },
     "ocpiPatchChargingStationsStatusesError": {
-      "title": "Itinérance Envoie des Statuts",
+      "title": "Itinérance envoie des statuts",
```
cabch, line 68: @ClaudeROSSI Traduction de "Roaming Send Statuses" pas sure que ce soit vraiment compréhensible ??

src/server/rest/v1/service/ChargingStationService.ts
```diff
@@ -1270,10 +1270,10 @@ export default class ChargingStationService {
         // TODO: To remove the 'lastHeartBeat' when new version of Mobile App will be released (> V1.3.22)
         'id', 'inactive', 'connectorsStatus', 'connectorsConsumption', 'public', 'firmwareVersion', 'chargePointVendor', 'chargePointModel',
         'ocppVersion', 'ocppProtocol', 'lastSeen', 'lastHeartBeat', 'firmwareUpdateStatus', 'coordinates', 'issuer', 'voltage',
-        'siteAreaID', 'siteArea.id', 'siteArea.name', 'siteArea.siteID', 'siteArea.site.name', 'siteArea.address', 'maximumPower',
-        'connectors.connectorId', 'connectors.status', 'connectors.type', 'connectors.power', 'connectors.errorCode',
+        'siteAreaID', 'siteArea.id', 'siteArea.name', 'siteArea.siteID', 'siteArea.site.name', 'siteArea.address', 'maximumPower', 'powerLimitUnit',
```
cabch, line 1273: These columns weren't retrieved from db so some values were empty in the exported csv

ClaudeROSSI (approved):

LucasBrazi06:
@ClaudeROSSI You can mark the conversation resolved once you checked them :)

LucasBrazi06:
@cabch Haven't we said that perf tests needed to be done to check both solutions you sent me ;-) (+= vs join)

LucasBrazi06 (approved):

PR #2045, opened 2021-01-20
03. Crypto - Change crypto setting in UI and trigger migration in backend

stefania-santimbrean:
Connected to ev-dashboard [PR #2215](https://github.com/LucasBrazi06/ev-dashboard/pull/2215)


src/server/rest/v1/service/security/SettingSecurity.ts
```diff
@@ -186,6 +186,23 @@ export default class SettingSecurity {
...
+              blockSize: sanitize(request.content.crypto.keyProperties?.blockSize),
+              operationMode: sanitize(request.content.crypto.keyProperties?.operationMode)
+            },
+            formerKey: sanitize(request.content.crypto.formerKey),
+            formerKeyProperties: {
+              blockCypher: sanitize(request.content.crypto.formerKeyProperties?.blockCypher),
+              blockSize: sanitize(request.content.crypto.formerKeyProperties?.blockSize),
+              operationMode: sanitize(request.content.crypto.formerKeyProperties?.operationMode)
+            },
+            migrationToBeDone: sanitize(request.content.crypto.migrationToBeDone)
```
jerome-benoit, line 203: That value shall not be set by a REST endpoint call, the server shall be the only responsible that set it by detecting if a key change have occurred or not in the call. 

Since it seems the key are hashed in the request payload, you can compare the hashed value of the key and the formerKey to know if the have changed or not.

src/utils/Cypher.ts
```diff
@@ -145,4 +139,110 @@ export default class Cypher {
       }
     }
   }
+
+  // This method will be reused in a Scheduler task that resumes migation
+  public static async handleCryptoSettingsChange(tenantID: string): Promise<void> {
+    const createDatabaseLock = LockingManager.createExclusiveLock(tenantID, LockEntity.DATABASE, 'migrate-settings-sensitive-data');
+    if (await LockingManager.acquire(createDatabaseLock)) {
+      try {
+        await Cypher.migrate(tenantID);
```
jerome-benoit, line 148: The migration is synchronous right now? And will be make asynchronous after, right?

jerome-benoit (changes requested):

src/utils/Cypher.ts
stefania-santimbrean: When changing a key from the dashboard and calling the REST API yes we wait for the migration to be executed and the error is thrown and shown in dashboard if migration is not completed. Is this not what the expected behavior should be? When should it be asynchronous?

jerome-benoit: The migration shall be run only by the crypto migration scheduler task at night, asynchronously. When a crypto settings occurs on a tenant, it just flag the boolean migrationDone to false. The REST settings endpoint will then also forbid any further crypto settings change and the dashboard behaviour will change regarding that value (grey input, etc.)

stefania-santimbrean: @LucasBrazi06 I understood from our meeting and the email with the minutes that when we chance a crypto key in the dashboard we also trigger a migration of the sensitive data. Should I change to only set migrationToBeDone to true and let the scheduler task do the migration? In the current implementation both the REST endpoint and the scheduler task call the same method which executes the migration

LucasBrazi06: I'm ok to have a synchronous migration call in the REST endpoint as this is not time consuming.
If this one fails, the scheduler will take over (checks once every minute) and re-run it.
So the current implementation is ok.

src/utils/Cypher.ts
```diff
@@ -161,4 +167,82 @@ export default class Cypher {
...
+    const settingsToMigrate = await Cypher.getSettingsWithSensitiveData(tenantID);
+    // If tenant has settings with sensitive data, migrate them
+    if (!Utils.isEmptyArray(settingsToMigrate)) {
+      // Migrate
+      for (const settingToMigrate of settingsToMigrate) {
+        // Migration already done?
+        if (Utils.isEmptyArray(settingToMigrate.formerSensitiveData)) {
+          // Save former senitive data in setting
+          const formerSensitiveData = Cypher.backupSensitiveData(settingToMigrate);
+          formerSensitiveData.push(Cypher.hash(cryptoSetting.formerKey));
```
LucasBrazi06, line 194: No need to push the former key in the backed up data

src/utils/Cypher.ts
```diff
@@ -161,4 +167,82 @@ export default class Cypher {
...
+
+  private static backupSensitiveData(setting: SettingDB): string[] {
+    const formerSensitiveData: string[] = [];
+    for (const property of setting.sensitiveData) {
+    // Check that the property does exist otherwise skip to the next property
+      if (Utils.objectHasProperty(setting, property)) {
+        const value = _.get(setting, property) as string;
+        // If the value is undefined, null or empty then do nothing and skip to the next property
+        if (value && value.length > 0) {
+          formerSensitiveData.push(value);
```
LucasBrazi06, line 227: Not only the value but also the key and in a Json doc:
`formerSensitiveData[property] = value;`

LucasBrazi06 (changes requested):
Check my comments

src/server/rest/v1/service/SettingService.ts
```diff
@@ -212,8 +205,22 @@ export default class SettingService {
     // Update timestamp
     setting.lastChangedBy = { 'id': req.user.id };
     setting.lastChangedOn = new Date();
+
+    if (settingUpdate.identifier === TenantComponents.CRYPTO) {
+      if (Cypher.hash(settingUpdate.content.crypto.key) !== Cypher.hash(setting.content.crypto.key)) {
+        settingUpdate.content.crypto.migrationToBeDone = true;
+      }
+      settingUpdate.content.crypto.formerKey = setting.content.crypto.key;
```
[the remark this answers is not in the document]
stefania-santimbrean: No, in dashboard I assign formerKeyProperties from keyProperties. The reason why I do the key here is beause it is hashed in the frontend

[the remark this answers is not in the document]
LucasBrazi06: I think it has to be done in the backend, not in the front end.
In the front-end you change the key and in the backend you handle the change.

src/utils/Cypher.ts
```diff
@@ -135,17 +140,82 @@ export default class Cypher {
...
-      throw new BackendError({
-        source: Constants.CENTRAL_SERVER,
-        module: MODULE_NAME,
-        method: 'getCryptoSetting',
-        message: `Tenant ID '${tenantID}' does not have crypto settings.`
-      });
+  // This method will be reused in a Scheduler task that resumes migation
+  public static async handleCryptoSettingsChange(tenantID: string): Promise<void> {
+    const createDatabaseLock = LockingManager.createExclusiveLock(tenantID, LockEntity.DATABASE, 'migrate-settings-sensitive-data');
+    if (await LockingManager.acquire(createDatabaseLock)) {
```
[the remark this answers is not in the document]
stefania-santimbrean: I would rather say that I test the migrationToBeDone boolean from the db setting and throw an error that migation is in progress like you also commented in ev-dashboard. I can also add an error for the lock if it cannot be acquired but I'm not sure if that case will ever exist.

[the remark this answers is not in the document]
LucasBrazi06: If I call this method from another place, nothing tells me if it has been run successfully or not.
Thus the exception.

src/utils/Cypher.ts
```diff
@@ -161,4 +174,82 @@ export default class Cypher {
...
+    const settingsToMigrate = await Cypher.getSettingsWithSensitiveData(tenantID);
+    // If tenant has settings with sensitive data, migrate them
+    if (!Utils.isEmptyArray(settingsToMigrate)) {
+      // Migrate
+      for (const settingToMigrate of settingsToMigrate) {
+        // Migration already done?
+        if (Utils.isEmptyArray(settingToMigrate.backupSensitiveData)) {
+          // Save former senitive data in setting
+          const backupSensitiveData = Cypher.prepareBackupSensitiveData(settingToMigrate);
+          backupSensitiveData.push(Cypher.hash(cryptoSetting.formerKey));
```
LucasBrazi06: No need to store the key and it is already stored in crypto settings.
Just removed it.

LucasBrazi06 (changes requested):
I made some changes.
You can double check and let me know.
In the end, it would be nice if we could have the unit tests in this PR also.
Could you add them and close the Unit Test PR + Issue?
I would help me to test the migration.
Thanks!

src/utils/Cypher.ts
stefania-santimbrean: Sorry I removed it, but then forgot to pull and when merging I lost it :(

LucasBrazi06 (commented):
I would need more unit tests to test the crypto key changes

PR #2049, opened 2021-01-21
04. Crypto - Scheduler task to continue migration in case of server crash

stefania-santimbrean:
Add in config file:
```
	"Scheduler": {
		"active": true,
		"tasks": [
			{...},
			{
				"name": "MigrateSensitiveDataTask",
				"active": true,
				"periodicity": "* * * * *",
				"config": {}
			}
		]
	}
```


src/utils/Cypher.ts
```diff
@@ -145,4 +139,111 @@ export default class Cypher {
...
+    // Filter settings with sensitiveData
+    return settings.result.filter((value: SettingDB) => {
+      if (value?.sensitiveData && !Utils.isEmptyArray(value?.sensitiveData)) {
+        return true;
+      }
+    });
+  }
+
+  public static async migrateSettings(tenantID: string, cryptoSetting: CryptoSetting): Promise<void> {
+
```
jerome-benoit, line 185: You can remove blank line.

src/utils/Cypher.ts
```diff
@@ -145,4 +139,111 @@ export default class Cypher {
...
+    const settingsToMigrate = await Cypher.getSettingsWithSensitiveData(tenantID);
+    // If tenant has settings with sensitive data, migrate them
+    if (!Utils.isEmptyArray(settingsToMigrate)) {
+      // Migrate
+      for (const setting of settingsToMigrate) {
+        if (!setting.formerSensitiveData && Utils.isEmptyArray(setting.formerSensitiveData)) {
+          delete setting.formerSensitiveData;
+          // Save former senitive data in setting
```
jerome-benoit, line 193: typo: sensitive

src/utils/Cypher.ts
```diff
@@ -1,44 +1,38 @@
...
-          method: 'getConfiguration',
-          message: 'Crypto configuration is missing'
-        });
-      }
-    }
-    return this.configuration;
-  }
 
-  public static encrypt(data: string): string {
+  public static async encrypt(data: string, tenantID: string, former = false): Promise<string> {
```
jerome-benoit, line 17: I do not really like the signature change here. A better signature would have been to just pass the key to use as an argument.

src/utils/Cypher.ts
```diff
@@ -145,4 +139,111 @@ export default class Cypher {
...
+  }
+
+  public static async migrateSettings(tenantID: string, cryptoSetting: CryptoSetting): Promise<void> {
+
+    const settingsToMigrate = await Cypher.getSettingsWithSensitiveData(tenantID);
+    // If tenant has settings with sensitive data, migrate them
+    if (!Utils.isEmptyArray(settingsToMigrate)) {
+      // Migrate
+      for (const setting of settingsToMigrate) {
+        if (!setting.formerSensitiveData && Utils.isEmptyArray(setting.formerSensitiveData)) {
```
jerome-benoit, line 191: I do not understand that test: if there's no formerSensitiveData (or an empty array of them), why launching the migration?

The sanity check is that at key change, if there's a previous former settings, they shall be replaced: current -> former, previous former -> deleted and new key is added.

jerome-benoit (changes requested):

src/utils/Cypher.ts
stefania-santimbrean: Yes, I agree. But in a previous review you suggested putting the tenantID in the encrypt/decrypt and making a getCryptoKey private method that we only call in Cypher.  But even if I pass the cryptoKey (which contains both key, keyPropeties, formerKey, formerKeyProperties), I still have the same problem. Is it better to create a new type which would contain only key and keyProperties and pass that to the encrypt/decrypt?

src/utils/Cypher.ts
stefania-santimbrean: After a migration is finished the formerSensitiveData are cleaned up from settings. Which means that if the setting does not have formerSensitiveData (and migrationToBeDone is true), that setting has to be migrated - and the opposite if it has formerSensitiveData (and migrationToBeDone is true), that settings was migrated. After a migration is completed, the formerSensitiveData is cleaned and migrationToBeDone is set to false

src/utils/Cypher.ts
jerome-benoit: Fair enough. Then just rename the boolean argument to something more meaningful like `useFormerKey`.

src/utils/Cypher.ts
jerome-benoit: > After a migration is finished the formerSensitiveData are cleaned up from settings. Which means that if the setting does not have formerSensitiveData (and migrationToBeDone is true), that setting has to be migrated - and the opposite if it has formerSensitiveData (and migrationToBeDone is true), that settings was migrated. After a migration is completed, the formerSensitiveData is cleaned and migrationToBeDone is set to false

The migration code shall be reentrant, with reentrant code, you they only need to test for migrationToBeDone and then run again the code. Testing for the migration status is then not needed.

src/utils/Cypher.ts
stefania-santimbrean: I renamed in PR 03. Crypto

src/utils/Cypher.ts
stefania-santimbrean: I don't understand what should I change? Testing migrationToBeDone is needed to know if migration is finished or not. Testing for formerSensitiveData is needed to know if a particular setting what migrated or not.

LucasBrazi06 (approved):

LucasBrazi06:
@jerome-benoit 
Could you please update AWS and SCP config with this new job?
```
{
  name": "MigrateSensitiveDataTask",
  "active": true,
  "periodicity": "* * * * *",
  "config": {}
}
```
Thanks.

PR #2062, opened 2021-01-29
Track CS registration status (boot notification accepted)

jerome-benoit:
And use it to update lastSeen only on CS that have successfully been registered: only registered ones will now be seen as online. 

That PR contains the minimal changes needed to deal with registration and online status


src/server/ocpp/json/JsonRestWSConnection.ts
```diff
@@ -83,7 +83,7 @@ export default class JsonRestWSConnection extends WSConnection {
         source: this.getChargingStationID(),
         module: MODULE_NAME,
         method: 'handleRequest',
-        message: 'Charging Station is not connected to the backend',
+        message: `Charging Station '${this.getChargingStationID()}' is not connected to the backend`,
```
LucasBrazi06, line 86: No need to add ${this.getChargingStationID()} as it is already provided in the 'source'.

LucasBrazi06 (changes requested):
Why the flag 'registered'?
If a charging station is in the DB it means that it's registered.
Moreover, too many files with formatting, should be in another PR.

jerome-benoit:
> Why the flag 'registered'?
> If a charging station is in the DB it means that it's registered.

In OCPP, a charging station is considered registered if the backend reply to a boot notification with the 'accepted' status. Otherwise it's either 'pending' (not implemented on e-mobility) or 'rejected'. 
Currently, the code can reject boot notification of charging station is the DB and after still continue to consider them as registered charging station, which is a bug, revealed by schneider lab stations latest reconnection tries to the backend. 

The OCPP boot notification command design is a charging station registration process (highly insecure in the spec, but that not the point). If the process fails, that station shall not be seen as online on e-mobility.

> Moreover, too many files with formatting, should be in another PR.

You're right but given our workload lately and the low amount of code cleanups on that PR, when I've done the fix, I've also done cleanups while reading code.

LucasBrazi06:
> The OCPP boot notification command design is a charging station registration process (highly insecure in the spec, but that not the point). If the process fails, matching stations shall not be seen as online on e-mobility.

This use case should not happen:
- If a new station is rejected, it is not created in the DB.
- If the station is accepted, it is created in the DB.

There is no other use case to handle here.
Do you confirm?

jerome-benoit:
> > The OCPP boot notification command design is a charging station registration process (highly insecure in the spec, but that not the point). If the process fails, matching stations shall not be seen as online on e-mobility.
> 
> This use case should not happen:
> 
> * If a new station is rejected, it is not created in the DB.
> * If the station is accepted, it is created in the DB.
> 
> There is no other use case to handle here.
> Do you confirm?

No ;)

+ The station is already in the DB and the boot notification is rejected because of changes in its technical attributes and it's not recognized anymore as the same charging station: 

```
        // Check if same vendor and model
        if ((chargingStation.chargePointVendor !== bootNotification.chargePointVendor ||
          chargingStation.chargePointModel !== bootNotification.chargePointModel) ||
          (chargingStation.chargePointSerialNumber && bootNotification.chargePointSerialNumber &&
            chargingStation.chargePointSerialNumber !== bootNotification.chargePointSerialNumber)) {
          // Not the same Charging Station!   

```

The backend is already trying to enforce connected station unicity per tenant. 
The end result is here a charging station with a rejected boot notification, not deleted and seen as online on e-mobility, but unusable because the station retry to register only by sending boot notifications on a regular basis. Deleting it logically (or the record in DB) is not an acceptable solution, it's still here and just fails to register at boot notification, hence the boolean to track also that status, ensure such the station are not seen as usable on e-mobility and log the reason for admins. 
And as I said, the boot notification process in OCPP is a registration process expected to be run at each boot with three states, the implementation here just handle two states to cope with an existing code path.

LucasBrazi06:
> The backend is already trying to enforce connected station unicity per tenant.
> The end result is here a charging station with a rejected boot notification, not deleted and seen as online on e-mobility, but unusable because the station retry to register only by sending boot notifications on a regular basis. 

Expected behavior!

> Deleting it logically (or the record in DB) is not an acceptable solution, it's still here and just fails to register at boot notification, hence the boolean to track also that status, ensure such station are not seen as usable on e-mobility and log the reason for admins.

It's not up to you to decide if it's acceptable or not.

> And as I said, the boot notification process in OCPP is a registration process expected to be run at each boot with three states, the implementation here just handle two states to cope with an existing code path that reject boot notification of existing stations.

2 states is ok, no need a new flag.
Moreover you haven't discussed this feature with me and I've never told you to implement it.
I think you have enough to do today than implementing on your own some unwanted features as usual.
I'll close this PR.

jerome-benoit:
> > The backend is already trying to enforce connected station unicity per tenant.
> > The end result is here a charging station with a rejected boot notification, not deleted and seen as online on e-mobility, but unusable because the station retry to register only by sending boot notifications on a regular basis.
> 
> Expected behavior!

Having a station station seen usable but unusable is an expected behaviour of e-mobility? 
Not having a proper handling of boot notification error (a basic feature) in e-mobility is an expected behaviour?
 
> > Deleting it logically (or the record in DB) is not an acceptable solution, it's still here and just fails to register at boot notification, hence the boolean to track also that status, ensure such station are not seen as usable on e-mobility and log the reason for admins.
> 
> It's not up to you to decide if it's acceptable or not.

Do you think having a buggy error handling of in the boot notification process is acceptable? 
Do you think not respecting the OCPP spec regarding charging station registration at boot notification is acceptable?
 
> > And as I said, the boot notification process in OCPP is a registration process expected to be run at each boot with three states, the implementation here just handle two states to cope with an existing code path that reject boot notification of existing stations.
> 
> 2 states is ok, no need a new flag.
> Moreover you haven't discussed this feature with me and I've never told you to implement it.
> I think you have enough to do today than implementing on your own some unwanted features as usual.
> I'll close this PR.

That PR does not implement a feature (and is not even close to that), that PR fix a bug regarding the current charging station registration process in e-mobility: 
  + its implementation shows that the OCPP spec are clearly misinterpreted;
  + its implementation does not handle boot notification error properly; 
  + its implementation misleads tenant admins about charging station usability status;
  + its implementation misleads users about charging station usability status (people at the Lab have tried and failed to use two of north side stations in boot notification rejection state);  

I reopen that PR until an agreement is found regarding that bug fixing: the station can be logically deleted instead but that remains a semantic error of the OCPP spec interpretation to consider the charging station deletion flag as a registration status flag. 
Station specs changes (firmware upgrade can change them) handling are still left buggy because it's not a 15min patch to do: no way to make the charging station usable again without doing manual changes in DB.

LucasBrazi06:
I already decided, stop discussing via GitHub.
You still can come in my office for further discussion.
You should have done this BEFORE YOU decide to go on that topic.

PR #2065, opened 2021-01-31
config scripts for mta deployment

captaineaglex:
I have added some config scripts which copies necessary configurations for respective server type on the MTA build runtime.


package.json
```diff
@@ -28,6 +28,13 @@
     "components:export": "node src/componentsExport.js",
     "start": "npm run start:prod",
     "prestart": "npm run build:prod",
+    "config:batch": "copy .\\src\\assets\\ev-scp-config\\e-Mobility-config\\config_batch_server.json .\\src\\assets\\config.json",
+    "config:oData": "copy .\\src\\assets\\ev-scp-config\\e-Mobility-config\\config_odata_server.json .\\src\\assets\\config.json",
+    "config:fakerest": "copy .\\src\\assets\\ev-scp-config\\e-Mobility-config\\config_fake_rest_server.json .\\src\\assets\\config.json",
+    "config:ocpi": "copy .\\src\\assets\\ev-scp-config\\e-Mobility-config\\config_ocpi_server.json .\\src\\assets\\config.json",
+    "config:rest": "copy .\\src\\assets\\ev-scp-config\\e-Mobility-config\\config_rest_server.json .\\src\\assets\\config.json",
+    "config:chargebox-json": "copy .\\src\\assets\\ev-scp-config\\e-Mobility-config\\config_chargebox_json_server.json .\\src\\assets\\config.json",
+    "config:chargebox-soap": "copy .\\src\\assets\\ev-scp-config\\e-Mobility-config\\config_chargebox_soap_server.json .\\src\\assets\\config.json",
```
jerome-benoit, line 37: The issue here is that the command used to make the copy is not OS agnostic. 

It's better to rely on a npm package installed as a dev dep like https://www.npmjs.com/package/copyfiles for example.

You probably also need to install mbt package as a dev dep to allow generation of the MTA archive from a script.

captaineaglex: You are right about copy methodology. I also had doubts about it but wasn't sure. Better to do as you suggest for copying files.

Do you want to build MTA archive for the server separately as well? We have another repository for MTA building where we build an MTA archive for the whole application where the server is a module of it. If we build a separate MTA definition for the server, when it is deployed to the same environment, it will act as a separate application (from the perspective of Diego architecture)(this might not cause apps to crash or disfunction). 

But it is possible to deploy only server components from the ev-sap-emobility-mta repository where we have mbt dependency added there.

jerome-benoit: > Do you want to build MTA archive for the server separately as well? We have another repository for MTA building where we build an MTA archive for the whole application where the server is a module of it. If we build a separate MTA definition for the server, when it is deployed to the same environment, it will act as a separate application (from the perspective of Diego architecture)(this might not cause apps to crash or disfunction).

I've forgot about the main repo for deployment ;) Forget about that remark.

captaineaglex: I will apply one change requested by Jerome. I will ensure the scripts are OS independent.

captaineaglex: I have added ncp module as a dev dependency and revised scripts with ncp function instead of using copy. backslash and function incompatibility between different OS should be dealt with by this change.

jerome-benoit (approved):
There should have changes in package-lock.json but that's not a showstopper for the merge. 
You have probably forgot to run npm install on that branch.

PR #2068, opened 2021-02-02
handling updateRegistrationToken from frontend

cabch:
need to add the unit tests


LucasBrazi06 (approved):
Gut !

src/server/rest/v1/service/RegistrationTokenService.ts
```diff
@@ -81,21 +81,8 @@ export default class RegistrationTokenService {
...
-      MODULE_NAME, 'handleUpdateRegistrationToken', req.user);
-    if (Utils.isComponentActiveFromToken(req.user, TenantComponents.ORGANIZATION)) {
-      // Check Site Area
-      const siteArea = await SiteAreaStorage.getSiteArea(req.user.tenantID, filteredRequest.siteAreaID);
-      UtilsService.assertObjectExists(action, siteArea, `Site Area ID '${filteredRequest.siteAreaID}' does not exist`,
-        MODULE_NAME, 'handleUpdateRegistrationToken', req.user);
-    }
     // Check Auth
-    if (!Authorizations.canUpdateRegistrationToken(req.user, registrationToken.siteArea?.siteID)) {
+    if (!Authorizations.canUpdateRegistrationToken(req.user, req.body.siteAreaID)) {
```
LucasBrazi06, line 85: What if the organization is not active in the tenant, should we pass the site area ID?
We cannot forbid the usage of token because the org is inactive.
Have you tested it with a tenant without an Org? Also check in the UI that Site Area is not showing up in that case.

src/server/rest/v1/service/RegistrationTokenService.ts
```diff
@@ -249,4 +251,47 @@ export default class RegistrationTokenService {
...
+    }
+    const filteredRequest = RegistrationTokenSecurity.filterRegistrationTokenRequest(req.query);
+    // Check User
+    let userProject: string[] = [];
+    if (Authorizations.canListUsers(req.user)) {
+      userProject = [ 'createdBy.name', 'createdBy.firstName', 'lastChangedBy.name', 'lastChangedBy.firstName' ];
+    }
+    // Get the token
+    const registrationToken = await RegistrationTokenStorage.getRegistrationToken(req.user.tenantID,
+      filteredRequest.id,
```
LucasBrazi06, line 274: If it's a GET rename id -> ID

src/server/rest/v1/service/security/RegistrationTokenSecurity.ts
```diff
@@ -23,6 +23,7 @@ export default class RegistrationTokenSecurity {
 
   public static filterRegistrationTokenRequest(request: any): Partial<RegistrationToken> {
```
LucasBrazi06, line 24: Create a new type here like HttpSiteAreaRequest
And GET params are in PascalCase: ID, Description, SiteAreaID.
Moreover for this get I would expect only the ID to get one token.

LucasBrazi06 (changes requested):
Check my comments

src/server/rest/v1/service/RegistrationTokenService.ts
cabch: I changed the `filterRegistrationTokenRequest` with `filterRegistrationTokenByIDRequest` so only the id is returned. it's already handled as "ID" param

src/server/rest/v1/service/security/RegistrationTokenSecurity.ts
cabch: Modified as `filterRegistrationTokenByIDRequest` as we only need the id in the end

src/server/rest/v1/service/RegistrationTokenService.ts
cabch: I let it as is as we will need to handle it soon after refactoring authorizations. NB: in the UI we don't see site area if org component is not active

jerome-benoit (changes requested):
You have added on that PR git submodules that duplicate the already existing ones. 
Please delete all directories referencing them.

test/api/RegistrationTokenTest.ts
```diff
@@ -0,0 +1,175 @@
...
+      // Create
+      it('Should not be able to create a new registration token', async () => {
+        // Check cannot create
+        testData.newRegistrationToken = Factory.registrationToken.build();
+        const response = await testData.basicCentralService.createEntity(
+          testData.basicCentralService.registrationApi,
+          testData.newRegistrationToken,
+          false
+        );
+        expect(response.status).to.be.eq(HTTPAuthError.ERROR);
```
jerome-benoit, line 124: Ideally, you should return a HTTP 403 code.

test/api/RegistrationTokenTest.ts
```diff
@@ -0,0 +1,175 @@
...
+
+      // Check readById
+      it('Should not be able to get registration token by id', async () => {
+        // Check cannot access
+        const response = await testData.basicCentralService.getEntityById(
+          testData.basicCentralService.registrationApi,
+          testData.newRegistrationToken,
+          false
+        );
+        expect(response.status).to.be.eq(HTTPAuthError.ERROR);
```
jerome-benoit, line 135: Same here.

test/api/RegistrationTokenTest.ts
```diff
@@ -0,0 +1,175 @@
...
+
+      // Check readAll
+      it('Should not be able to retrieve registration tokens list', async () => {
+        // Check cannot access
+        const response = await testData.basicCentralService.checkEntityInList(
+          testData.basicCentralService.registrationApi,
+          testData.newRegistrationToken,
+          false
+        );
+        expect(response.status).to.equal(HTTPAuthError.ERROR);
```
jerome-benoit, line 146: Same here.

test/api/RegistrationTokenTest.ts
```diff
@@ -0,0 +1,175 @@
...
+        // Check cannot update
+        testData.newRegistrationToken.description = faker.random.word();
+        testData.newRegistrationToken.expirationDate = faker.date.future();
+
+        const response = await testData.basicCentralService.updateEntity(
+          testData.basicCentralService.registrationApi,
+          testData.newRegistrationToken,
+          false
+        );
+        expect(response.status).to.equal(HTTPAuthError.ERROR);
```
jerome-benoit, line 160: Same here.

src/authorization/Authorizations.ts
```diff
@@ -389,11 +389,9 @@ export default class Authorizations {
     });
   }
 
-  public static canReadRegistrationToken(loggedUser: UserToken, siteID: string): boolean {
-    return Authorizations.canPerformAction(loggedUser, Entity.TOKEN, Action.READ, {
-      site: siteID,
-      sites: loggedUser.sitesAdmin
```
jerome-benoit, line 395: I think that will remove some token handling rights to site admin. Is it intentional?
Have you pushed down that authorization in code paths below in the call stack?

jerome-benoit (changes requested):

test/api/RegistrationTokenTest.ts
LucasBrazi06: Done. They all return 403 now :-)

src/authorization/Authorizations.ts
LucasBrazi06: @cabch You have to handle the role Site Admin in here + in the Unit Test.
Nevertheless I will merge you PR.
We can have a meeting for this special role.

cabch: Ok, yes it would be great to clarify that :) thanks
I did it that way as i didn't get the point why I would need sites here

PR #2074, opened 2021-02-04
Add GPS coordinates to sensitive data to obfuscate in logs.

jerome-benoit:
And name firstName. 
The first might also match valid name fields that should be displayed. Let's see if it really makes the logs less usable.  

SAP Security requirement.

### DIFF
--- a/src/utils/Constants.ts
+++ b/src/utils/Constants.ts
@@ -32,7 +32,7 @@ export default class Constants {
   public static readonly EXPORT_RECORD_MAX_COUNT = 100000;
 
   public static readonly DEFAULT_TENANT = 'default';
-  public static readonly DEFAULT_TENANT_OBJECT= Object.freeze({
+  public static readonly DEFAULT_TENANT_OBJECT = Object.freeze({
     id: Constants.DEFAULT_TENANT,
     name: Constants.DEFAULT_TENANT
   } as Tenant);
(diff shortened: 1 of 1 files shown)


LucasBrazi06:
@jerome-benoit 
Unit test failed.
I provided a fix 931b730f1db9f493d9d62303b9a1164203d3da00
Again another blind commit without testing?

jerome-benoit:
> Again another blind commit without testing?

Again another blind unpleasant and unconstructive remark on collaborator work without having a clue about the related work done and issue at stake?  
The fix is incorrect, master-qa package-lock.json is broken by latest merges, I'm on vacation and I will not work anymore outside my working hours on e-mobility for obvious reason.

jerome-benoit:
The 'fix' you've made introduced an information disclosure security hole in the logging code: cloning the message before the obfuscation make the sensitive information obfuscation of no-op. 

Yet another non reviewed commit pushed blindly which can have important impact on the production.

LucasBrazi06:
> The 'fix' you've made introduced an information disclosure security hole in the logging code: cloning the message before the obfuscation make the sensitive information obfuscation of no-op.

This fix was confirmed by Tobias.

jerome-benoit:
> > The 'fix' you've made introduced an information disclosure security hole in the logging code: cloning the message before the obfuscation make the sensitive information obfuscation of no-op.
> 
> This fix was confirmed by Tobias.

Confirmed to disclose sensitive information: it's node a shallow clone, it's a deep clone, the obfuscation is done on the cloned object and the cloned object is not returned by the method.

LucasBrazi06:
> Confirmed to disclose sensitive informations: it's not a shallow clone, it's a deep clone, the obfuscation is done on the cloned object and the cloned object is not returned by the method

Then a unit test that checks this scenario must be provided and a proper fix also.

jerome-benoit:
> > Confirmed to disclose sensitive informations: it's not a shallow clone, it's a deep clone, the obfuscation is done on the cloned object and the cloned object is not returned by the method
> 
> Then a unit test that checks this scenario must be provided and a proper fix also.

All commits must reviewed, especially if the code paths are security sensitive, period. 

And it's illegal to delete a message without the user consent. It's the second time it happened and that's already way beyond acceptable good manners. If the second time it happened, I will not accept a third time.

PR #2075, opened 2021-02-04
Add cars migration task

mahdibenromdhane:
- Data in cars-definition.json to be verified.
- If resizing images to thumb using sharp module is validated, I will create a migration task to align ev-database car thumb images.


LucasBrazi06 (approved):
Gut !

PR #2083, opened 2021-02-08
Proposal: Use backup email in constructor + remove singleton as notif…

LucasBrazi06:
…icationSources is a static var in NotificationHandler.


LucasBrazi06:
@jerome-benoit My proposal to simplify the code in smtp email handling.

PR #2086, opened 2021-02-08
Stripe 1st integration

ClaudeROSSI:
Two endpoints have been added:
- one allowing the current user to link his card to his account (as a default payment method)
- one to trigger the invoice payment (using the the default payment method)

No unit tests so far ... existing billing tests are failing!


src/server/rest/CentralRestServerService.ts
```diff
@@ -84,6 +84,7 @@ class RequestMapper {
           [ServerAction.BILLING_SYNCHRONIZE_INVOICES]: BillingService.handleSynchronizeInvoices.bind(this),
           [ServerAction.BILLING_FORCE_SYNCHRONIZE_USER_INVOICES]: BillingService.handleForceSynchronizeUserInvoices.bind(this),
           [ServerAction.BILLING_CREATE_TRANSACTION_INVOICE]: BillingService.handleCreateTransactionInvoice.bind(this),
+          [ServerAction.BILLING_ATTACH_PAYMENT_METHOD]: BillingService.handleBillingAttachPaymentMethod.bind(this),
```
LucasBrazi06, line 87: You can also use the new RESTful API by adding BillingRouter.ts class in server/rest/v1/router/api
And declare the billing class in the GlobalRoute.ts
Check TenantRouter class, it should work the same.
So no need changes in that class anymore.

src/integration/billing/stripe/StripeBillingIntegration.ts
```diff
@@ -394,6 +395,106 @@ export default class StripeBillingIntegration extends BillingIntegration<StripeB
...
+      console.log('💳 Attached ' + event.data.object.id + ' to customer');
+    } else {
+      console.log(`❌ unexpected event : ${event.type}`);
+    }
+    return true;
+  }
+
+  public async chargeInvoice(invoice: BillingInvoice): Promise<unknown> {
+    await this.checkConnection();
+
```
LucasBrazi06, line 441: C'est un test la ligne blanche ? :-)

LucasBrazi06 (changes requested):
Use the new RESTful endpoint.

src/integration/billing/stripe/StripeBillingIntegration.ts
ClaudeROSSI: ooops ... removed ;)

src/server/rest/CentralRestServerService.ts
ClaudeROSSI: BillingRouter added ...
But I'm not sure to understand what happens if we have it twice (once in new RestFul API and still  declared the old way!)

LucasBrazi06: You can remove the old ones.
The old REST interface will be replaced by the new RESTful one.

ClaudeROSSI:
Closing this one because a second branch will be used to provide the same coding after switching to the latest STRIPE API - version 8.*.*

PR #2090, opened 2021-02-09
billing-ut - database initialize properly

ClaudeROSSI:
minor changes - to make it possible to start billing tests

### DIFF
--- a/test/api/BillingTest.ts
+++ b/test/api/BillingTest.ts
@@ -11,6 +11,7 @@ import ContextProvider from './context/ContextProvider';
 import Cypher from '../../src/utils/Cypher';
 import Factory from '../factories/Factory';
 import { HTTPAuthError } from '../../src/types/HTTPError';
+import MongoDBStorage from '../../src/storage/mongodb/MongoDBStorage';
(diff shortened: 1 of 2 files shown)


test/api/BillingTest.ts
```diff
@@ -112,6 +114,8 @@ describe('Billing Service', function() {
   this.timeout(1000000);
   describe('With component Billing (tenant utbilling)', () => {
     before(async () => {
+      global.database = new MongoDBStorage(config.get('storage'));
```
jerome-benoit, line 117: Ideally you should test that global.database is falsy, but nobody does that in the unit tests ;)

ClaudeROSSI: Would it be possible to get an approval for that  change as is?
I need it in other PR to be able to simply trigger the billing tests. thanks!

jerome-benoit: I can approuve it but the merge will be done by Serge when his time permits, unless the rules have changed and now you also have the rights to do merge after one approval by me.

jerome-benoit (approved):

PR #2097, opened 2021-02-12
Activate user account manually back

cabch:
- Migration task to add the user setting with account activation by default set to false
- New tab `user` in `Technical settings` menu with checkbox to enable/disable auto activation of user accounts
- email verification set user to inactive if user's `don't activate by default` setting is set to true / active otherwise

N.B: cannot test locally validation by clicking the verification email without adding hardcoded data - need to be tested


src/migration/tasks/MigrateUserSettingsTask.ts
```diff
@@ -0,0 +1,45 @@
...
+    const userSetting = await SettingStorage.getUserSettings(tenant.id);
+    // If no user setting exists, initialize it
+    if (Utils.isEmptyObject(userSetting)) {
+      // Create new user setting with account activation param
+      const settingsToSave = {
+        identifier: UserSettingsType.USER,
+        content: {
+          type: UserSettingsContentType.ACCOUNT_ACTIVATION,
+          accountActivation: {
+            doNotActivateByDefault: false
```
jerome-benoit, line 28: I'm not very fond of that name: `enabled: true/false` is enough

cabch: I agree, I didn't find something clear enough and forgot to ask before PR @LucasBrazi06 is 'enabled' ok ?

LucasBrazi06: If we follow the same Settings structure we should have something like this:
```
{ 
    "_id" : ObjectId("5c6c8e8ee7fd060008215e30"), 
    "identifier" : "user", 
    "content" : {
        "type" : "user", 
        "user" : {
            "doNotActivateByDefault": true
        }
    }, 
    "createdBy" : null, 
    "createdOn" : ISODate("2019-02-19T23:17:34.189+0000"), 
    "lastChangedBy" : null, 
    "lastChangedOn" : ISODate("2019-06-14T09:38:08.048+0000"), 
    "sensitiveData" : []
}
```
@cabch I thought this was already discussed last Thursday ;-)

jerome-benoit: >     "content" : {
>         "type" : "user", 
>         "user" : {
>             "doNotActivateByDefault": true
>         }

accountActivationByDefault: true/false is more understandable.

No other settings data structure use a negation for feature enablement on boolean. From a user experience perpective, it makes no sense: 

export interface BillingSetting {
```
  usersLastSynchronizedOn?: Date;
  invoicesLastSynchronizedOn?: Date;
}

export interface StripeBillingSetting extends BillingSetting {
  url: string;
  secretKey: string;
  publicKey: string;
  noCardAllowed: boolean;
  immediateBillingAllowed: boolean;
  periodicBillingAllowed: boolean;
  advanceBillingAllowed: boolean;
  currency: string;
  taxID: string;
}
```

LucasBrazi06: > No other settings data structure use a negation for feature enablement on boolean. From a user experience perpective, it makes no sense:

I agree.
@cabch Could you please rename doNotActivateByDefault to autoActivateAccountAfterValidation or something equivalent and proceed with the changes in the front-end also?
Thanks.

ClaudeROSSI: Can't we just name it "Manual Activation" ON/OFF?

LucasBrazi06:
@cabch Not linked to an issue

LucasBrazi06:
@cabch I fixed issues in verifying user's account.
Could you merge with master-qa and check that your code still works?
Thanks!

cabch:
> @cabch I fixed issues in verifying user's account.
> Could you merge with master-qa and check that your code still works?
> Thanks!

still works on my side 👌🏻

src/migration/tasks/MigrateUserSettingsTask.ts
```diff
@@ -0,0 +1,45 @@
...
+    const userSetting = await SettingStorage.getUserSettings(tenant.id);
+    // If no user setting exists, initialize it
+    if (Utils.isEmptyObject(userSetting)) {
+      // Create new user setting with account activation param
+      const settingsToSave = {
+        identifier: UserSettingsType.USER,
+        content: {
+          type: UserSettingsContentType.USER,
+          user: {
+            manualAccountActivation: false
```
ClaudeROSSI, line 28: Which value is the default one when creating a new tenant?

ClaudeROSSI (changes requested):

src/server/rest/v1/service/TenantService.ts
```diff
@@ -187,6 +188,25 @@ export default class TenantService {
...
+    // Check for settings in db
+    const userSettings = await SettingStorage.getUserSettings(tenantID);
+    if (Utils.isEmptyObject(userSettings)) {
+      // Create new user setting with account activation param
+      const settingsToSave = {
+        identifier: UserSettingsType.USER,
+        content: {
+          type: UserSettingsContentType.USER,
+          user: {
+            manualAccountActivation: false
```
cabch, line 201: @ClaudeROSSI here is the new tenant user setting

I added "false" otherwise I will need to add some other logic to handle the fact the first user (the one that created the tenant) will be blocked as its accoount will be in "manual activation mode" == inactive == he won't be able to even log to the app.

src/migration/tasks/MigrateUserSettingsTask.ts
cabch: I added it in TenantService.ts :)

src/server/rest/v1/service/security/SettingSecurity.ts
```diff
@@ -188,6 +188,12 @@ export default class SettingSecurity {
             settings.content.asset.connections.push(sanitizedConnection);
           }
           break;
+        case UserSettingsContentType.USER:
+          settings.content.user =
+          {
```
ClaudeROSSI, line 193: let's put the opening bracket on the previous line! (Just for you Serge) ;)

src/server/rest/v1/service/security/SettingSecurity.ts
```diff
@@ -188,6 +188,12 @@ export default class SettingSecurity {
...
+            manualAccountActivation: request.content.user.manualAccountActivation
```
ClaudeROSSI, line 194: please sanitize all values coming from the request.

like: 
sanitize(request.content.stripe.immediateBillingAllowed),

src/storage/mongodb/SettingStorage.ts
```diff
@@ -335,6 +335,18 @@ export default class SettingStorage {
     }
   }
 
+  public static async getUserSettings(tenantID: string): Promise<UserSetting> {
+    // Get the user settings
+    const settings = await SettingStorage.getSettings(tenantID,
+      { identifier: UserSettingsType.USER },
+      Constants.DB_PARAMS_MAX_LIMIT);
```
ClaudeROSSI, line 342: I do not understand the DB_PARAMS_MAX_LIMIT. 
We always have a single user config. right?
so would should ask for only one config. using  Constants.DB_PARAMS_SINGLE_RECORD.

src/storage/mongodb/SettingStorage.ts
```diff
@@ -335,6 +335,18 @@ export default class SettingStorage {
...
+    const userSettings = {} as UserSetting;
```
ClaudeROSSI, line 343: I do not understand why we have this intermediate {} object.
Can't you just use and return settings.result[0] or null when nothing is found ????

src/server/rest/v1/service/TenantService.ts
```diff
@@ -187,6 +188,25 @@ export default class TenantService {
     }
   }
 
+  public static async createInitialUserSetting(tenantID: string): Promise<void> {
+    // Check for settings in db
+    const userSettings = await SettingStorage.getUserSettings(tenantID);
+    if (Utils.isEmptyObject(userSettings)) {
```
ClaudeROSSI, line 194: do we really need to check here ... user settings are for sure empty while creating a new  tenant !
Anyway, I would suggest to simply check:

if (userSettings) {
...
}

and make sure getUserSettings() method returns null when nothing is found!

src/types/Setting.ts
```diff
@@ -3,7 +3,7 @@ import TenantComponents from './TenantComponents';
 
 export interface Setting {
   id?: string;
-  identifier: TenantComponents;
+  identifier: TenantComponents | UserSettingsType;
```
ClaudeROSSI, line 6: franchement ... j'aime pas le ou " | User..."

on ne pouvait pas juste enrichir l'enum TenantComponents ... meme si c'est pas vraiment un component?
@Serge! Help!!!

ClaudeROSSI (commented):
Franchement pas facile pour moi les PR reviews ...

J'ai mis des commentaires ... mais je ne suis pas sur de moi!

src/server/rest/v1/service/security/SettingSecurity.ts
jerome-benoit: It's the current practice but Serge wants to use schema validation in the future on all REST endpoints.

src/storage/mongodb/SettingStorage.ts
jerome-benoit: It's a const variables definition requirement, you have to initialise them.

src/types/Setting.ts
jerome-benoit: Yes, it's like the crypto settings: it shall reuse the same code for components but it's in the sub category of technical settings.

src/storage/mongodb/SettingStorage.ts
cabch: it makes sense - done

src/server/rest/v1/service/TenantService.ts
cabch: we don't really need to check here this was just a security - I replaced with if (!userSettings) 
please tell if I need to delete this check

src/types/Setting.ts
cabch: modified to TechnicalSettingsType as it can contain user, crypto and other that are not related to components - crypto shouldn't be in components

src/storage/mongodb/SettingStorage.ts
cabch: modified with @ClaudeROSSI

src/server/rest/v1/service/security/SettingSecurity.ts
cabch: sanitize added

src/migration/tasks/MigrateUserSettingsTask.ts
cabch: replaced manualAccountActivation to autoActivateAccountAfterValidation and all the true/false resulting from it

cabch: true by default (= users are auto activated for every new tenants and for all the existing tenants so the behaviour doesn't change from today's)

src/storage/mongodb/SettingStorage.ts
```diff
@@ -335,6 +335,16 @@ export default class SettingStorage {
...
+  public static async getUserSettings(tenantID: string): Promise<UserSetting> {
+    let userSetting : UserSetting = null;
+    // Get the user settings
+    const settings = await SettingStorage.getSettings(tenantID, { identifier: TechnicalSettingsType.USER }, Constants.DB_PARAMS_SINGLE_RECORD);
+    if (settings && settings.count > 0 && settings.result[0]) {
+      userSetting = settings.result[0] as UserSetting;
+    }
+    return userSetting;
+  }
+
```
ClaudeROSSI, line 347: @Serge, @Jerome: I'm confused! When looking at the others 'getSettings' method it seems that for some of them we want to get null when no settings are found (like the cryptoSettings) and for the others we return somehow a copy (clone) of the current settings.

Also the code is very verbose for most of them. A lot of lines for (in the end) simply copying the values.
Why don't we just use lodash and conditional chaining?

Something like:
image

Let me know what you think!

src/server/rest/v1/service/AuthService.ts
```diff
@@ -571,11 +575,11 @@ export default class AuthService {
       tenantID: tenantID,
       user: user, action: action,
       module: MODULE_NAME, method: 'handleVerifyEmail',
-      message: 'User account has been successfully verified and activated',
+      message: userStatus === UserStatus.ACTIVE ? 'User account has been successfully verified and activated' : 'User account has been successfully verified but needs an admin to activate it',
       detailedMessages: { params: req.query }
     });
     // Ok
-    res.json(Constants.REST_RESPONSE_SUCCESS);
+    res.json(userStatus === UserStatus.ACTIVE ? Constants.REST_RESPONSE_SUCCESS : { status: 'Success', userStatus: UserStatus.INACTIVE });
```
ClaudeROSSI, line 582: I would suggest to send back the userStatus in both cases.
Why should we only have the information when it is inactive?
By the way, is this information consumed by the caller? if not, please remove it!

cabch: yes will modify it

src/migration/tasks/MigrateUserSettingsTask.ts
ClaudeROSSI: ok ... autoActiavted by default ... we can anyway change it later!

src/storage/mongodb/SettingStorage.ts
ClaudeROSSI: after discussion with serge ... let's open an issue for this!

ClaudeROSSI (approved):
well done!

LucasBrazi06 (approved):

PR #2104, opened 2021-02-15
Sensitive data test

roxiciobotea210:
Added tests for checking if sensitive data items are obfuscated in logs.
Made log methods async.
Return and propagate log id in saveLog method.
Update anonymizeSensitiveData method logic - it was not working for strings & the RegExp was not working for all scenarios.


src/storage/mongodb/LoggingStorage.ts
```diff
@@ -80,6 +80,7 @@ export default class LoggingStorage {
     // Insert
     if (global.database) {
       await global.database.getCollection<Log>(tenantID, 'logs').insertOne(logMDB);
+      return logMDB._id.toHexString();
```
jerome-benoit, line 83: _id already have an ObjectID type. No need to convert it, just return it and reuse. 
It's a hot code path, it should be as light as possible.

src/utils/Logging.ts
```diff
@@ -719,6 +721,9 @@ export default class Logging {
       for (const sensitiveData of Constants.SENSITIVE_DATA) {
         // Anonymize
         message.replace(new RegExp(sensitiveData, 'gi'), Constants.ANONYMIZED_VALUE);
+        // Should be
+        // message=message.replace(new RegExp(sensitiveData, 'gi'), Constants.ANONYMIZED_VALUE); //in order to anonymize the FULL string
```
jerome-benoit, line 725: `message.replace(new RegExp(sensitiveData, 'gi') ... = message.replace(new RegExp(sensitiveData, 'gi') ... `
The fix proposal does not fix anything ;) 
If you want to fix the regexp, you should look for the string after the senvisite data key and obfuscated only that part: 
+ password: blah blah -> password: ####
+ password=blah blah -> password=###
+ ...

src/utils/Logging.ts
```diff
@@ -730,7 +735,7 @@ export default class Logging {
         if (typeof message[key] === 'string') {
           for (const sensitiveData of Constants.SENSITIVE_DATA) {
             if (key.toLocaleLowerCase() === sensitiveData.toLocaleLowerCase()) {
-              // Anonymize
+            // Anonymize
```
jerome-benoit, line 738: Incorrect reindentation.

test/api/SecurityTest.ts
```diff
@@ -0,0 +1,224 @@
...
+  after(async function() {
+    // Housekeeping
+    // Reset components before leaving
+    const res = await testData.superCentralService.updateEntity(
+      testData.centralService.tenantApi, initialTenant);
+    expect(res.status).to.equal(200);
+  });
+
+  describe('Success cases (tenant utall)', () => {
+    it('Check that sensitive data string is anonymized', async () => { // Will fail
```
jerome-benoit, line 82: If it fails, the code must be fixed to make it works.

test/api/SecurityTest.ts
```diff
@@ -0,0 +1,224 @@
...
+    });
+    it('Check that everything goes well if detailed message is number', async () => { // Will fail
+      const logId:string = await Logging.logDebug({
+        source: 'test',
+        tenantID: testData.credentials.tenantId,
+        action: ServerAction.HTTP_REQUEST,
+        message: 'Just a test',
+        module: 'test',
+        method: 'test',
+        detailedMessages: 123 // Is this a valid usecase?
```
jerome-benoit, line 199: Not really, it just tests that the recursion stop if the data type cannot be obfuscated.
Please remove it for now.
We will need to status first on sensitive data handling in number type.

test/api/SecurityTest.ts
```diff
@@ -0,0 +1,224 @@
...
+    });
+    it('Check that everything goes well if detailed message is object with sensitive data field number', async () => { // Should fail, but doesn't
+      const logId:string = await Logging.logDebug({
+        source: 'test',
+        tenantID: testData.credentials.tenantId,
+        action: ServerAction.HTTP_REQUEST,
+        message: 'Just a test',
+        module: 'test',
+        method: 'test',
+        detailedMessages: { password: 123 } // Is this a valid usecase?
```
jerome-benoit, line 215: Same comment here.

test/api/SecurityTest.ts
```diff
@@ -0,0 +1,224 @@
...
+    });
+    it('Check that everything goes well if detailed message is boolean', async () => { // Will fail
+      const logId:string = await Logging.logDebug({
+        source: 'test',
+        tenantID: testData.credentials.tenantId,
+        action: ServerAction.HTTP_REQUEST,
+        message: 'Just a test',
+        module: 'test',
+        method: 'test',
+        detailedMessages: true // Is this a valid usecase?
```
jerome-benoit, line 183: Not really, it just tests that the recursion stop if the data type cannot be obfuscated.
Please remove.

jerome-benoit (changes requested):

src/utils/Logging.ts
roxiciobotea210: Will work on this, thanks. Is it ok if I do it in this PR? Since it's not related to tests.

jerome-benoit: > Will work on this, thanks. Is it ok if I do it in this PR? Since it's not related to tests.

That's fine. Update the PR changelog accordingly.

jerome-benoit:
@roxiciobotea210 Could please update the Jira ticket relative to that task? 
Time spent, status, etc.

test/api/SecurityTest.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+  } else if (typeof message === 'object') { // In case of object, check every field
+    for (const key of Object.keys(message)) {
+      if (typeof message[key] === 'string') { // If the field value is a string and the key is part of sensitive data, check if the value is anonymized
+        for (const sensitiveData of Constants.SENSITIVE_DATA) {
+          if (key.toLocaleLowerCase() === sensitiveData.toLocaleLowerCase()) {
+            expect(message[key]).to.equal(Constants.ANONYMIZED_VALUE);
+          }
+        }
+
```
jerome-benoit, line 36: Remove unneeded blank line.

src/utils/Logging.ts
jerome-benoit: Needless to say having sensitive data that might not be properly obfuscated is a security issue that have a very high priority and must be fixed ASAP ;)

roxiciobotea210: I know, it is my main focus. :)

roxiciobotea210:
> @roxiciobotea210 Could please update the Jira ticket relative to that task?
> Time spent, status, etc.

@jerome-benoit Are you referring to the Zenhub issue? I looked but I don't see where i can add these details in there. Or do we also have a Jira board? If so, can you please point me to it? Many thanks.

jerome-benoit:
> > @roxiciobotea210 Could please update the Jira ticket relative to that task?
> > Time spent, status, etc.
> 
> @jerome-benoit Are you referring to the Zenhub issue? I looked but I don't see where i can add these details in there. Or do we also have a Jira board? If so, can you please point me to it? Many thanks.

No Jira ticket != ZenHub issue. 
You should have attended to a meeting initiated by Frederic Coutellier that have presented Jira and its usage within e-mobility. Now you have to use it ;)

roxiciobotea210:
> > > @roxiciobotea210 Could please update the Jira ticket relative to that task?
> > > Time spent, status, etc.
> > 
> > 
> > @jerome-benoit Are you referring to the Zenhub issue? I looked but I don't see where i can add these details in there. Or do we also have a Jira board? If so, can you please point me to it? Many thanks.
> 
> No Jira ticket != ZenHub issue.
> You should have attended to a meeting initiated by Frederic Coutellier that have presented Jira and its usage within e-mobility. Now you have to use it ;)

I don't know when this meeting took place, I joined the project in the beginning of this month and I did not receive an invitation for it. Anyways :) I hope I know how to use Jira well enough to update a ticket, I just don't have a link to the board and can't find it anywhere. So would you please be kind to send me a link? Or if there is a recording of that meeting, I'd be happy to watch it if you tell me where to find it. Many thanks! :)

jerome-benoit:
> > > > @roxiciobotea210 Could please update the Jira ticket relative to that task?
> > > > Time spent, status, etc.
> > > 
> > > 
> > > @jerome-benoit Are you referring to the Zenhub issue? I looked but I don't see where i can add these details in there. Or do we also have a Jira board? If so, can you please point me to it? Many thanks.
> > 
> > 
> > No Jira ticket != ZenHub issue.
> > You should have attended to a meeting initiated by Frederic Coutellier that have presented Jira and its usage within e-mobility. Now you have to use it ;)
> 
> I don't know when this meeting took place, I joined the project in the beginning of this month and I did not receive an invitation for it. 

Previous week. The link to Jira will be sent by mail later on.

LucasBrazi06:
> Previous week. The link to Jira will be sent by mail later on.

Fred forgot to invite Roxana :-(

src/storage/mongodb/LoggingStorage.ts
roxiciobotea210: I tried but it seems to fail without this. Also - this is how I saw id-s being used throughout the code.

jerome-benoit: Read carefully my comment: `_id` already have type defined in the mongodb package types declaration. No need to do an heavy type conversion in a hot code path, just use the type already defined for the `_id` field.

src/storage/mongodb/LoggingStorage.ts
```diff
@@ -80,6 +81,7 @@ export default class LoggingStorage {
     // Insert
     if (global.database) {
       await global.database.getCollection<Log>(tenantID, 'logs').insertOne(logMDB);
+      return logMDB._id;
```
jerome-benoit, line 84: Could you please add a comment on the reason with pushConvertObjectIDToString() helper is not here? 
`// For performance reason, the ObjectID type is returned and used.`
Or something similar.

src/utils/Logging.ts
```diff
@@ -703,67 +704,70 @@ export default class Logging {
...
+            if (dataPart.toLowerCase().startsWith(sensitiveData.toLocaleLowerCase())) {
+              // Anonymize each query string part
+              dataParts[i] = dataPart.substring(0, sensitiveData.length + 1) + Constants.ANONYMIZED_VALUE;
+            }
+          }
+        }
+        message = dataParts.join('&');
+        return message;
+      }
+      // If it is a string, but not a query string
```
jerome-benoit, line 736: // Check if ...

src/utils/Logging.ts
```diff
@@ -703,67 +704,70 @@ export default class Logging {
...
       for (const sensitiveData of Constants.SENSITIVE_DATA) {
-        // Anonymize
-        message.replace(new RegExp(sensitiveData, 'gi'), Constants.ANONYMIZED_VALUE);
+        if (new RegExp(sensitiveData + '[:=](.*)', 'gi').test(message)) { // If it contains sensitive data and it matches the RegExp, Anonymize
```
jerome-benoit, line 738: You do not handle here if there might have space before : or =

I would also add , or ; for string that might come from CSV.

src/utils/Logging.ts
```diff
@@ -703,67 +704,70 @@ export default class Logging {
...
+          message = message.substring(0,sensitiveData.length + 1) + Constants.ANONYMIZED_VALUE;
```
jerome-benoit, line 739: Missing space after comma. 

Please install the vscode plugin recommended for the repo and give them the rights they need to format and lint the code added.

src/utils/Logging.ts
```diff
@@ -703,67 +704,70 @@ export default class Logging {
...
-      Logging.logError({
-        tenantID: Constants.DEFAULT_TENANT,
-        type: LogType.SECURITY,
-        module: MODULE_NAME,
-        method: 'anonymizeSensitiveData',
-        action: ServerAction.LOGGING,
-        message: 'No matching object type for log message anonymisation',
-        detailedMessages: { message: message }
-      });
+      return message;
```
jerome-benoit, line 759: Why do you need to return the message? The object instance is cloned and the recursion do change directly the instance.

src/utils/Logging.ts
```diff
@@ -703,67 +704,70 @@ export default class Logging {
...
+        message = dataParts.join('&');
+        return message;
+      }
+      // If it is a string, but not a query string
       for (const sensitiveData of Constants.SENSITIVE_DATA) {
-        // Anonymize
-        message.replace(new RegExp(sensitiveData, 'gi'), Constants.ANONYMIZED_VALUE);
+        if (new RegExp(sensitiveData + '[:=](.*)', 'gi').test(message)) { // If it contains sensitive data and it matches the RegExp, Anonymize
+          message = message.substring(0,sensitiveData.length + 1) + Constants.ANONYMIZED_VALUE;
+          return message;
```
jerome-benoit, line 740: Same as below why returning the message?

jerome-benoit (changes requested):

src/utils/Logging.ts
roxiciobotea210: I modified the function to use returns everywhere because I noticed that it was not doing anything when the message was of type string, meaning that -  the string that was being saved in the Log object was always the original string without any modification (data was not anonymized).
I believe this has to do with the fact that in JS/TS strings are passed by value, not by reference, so the reference to the string in the original object was not updated to the modified value.

jerome-benoit: > I modified the function to use returns everywhere because I noticed that it was not doing anything when the message was of type string, meaning that - the string that was being saved in the Log object was always the original string without any modification (data was not anonymized).
> I believe this has to do with the fact that in JS/TS strings are passed by value, not by reference, so the reference to the string in the original object was not updated to the modified value.

Fair enough then.

src/utils/Logging.ts
roxiciobotea210: I have all extensions installed, I have to check what permissions are missing.

src/utils/Logging.ts
roxiciobotea210: Updated RegExp to take into consideration , and ; .
Also takes into account now potential space before/after : = , ; .

jerome-benoit: Also add some string examples in the unit tests with spaces, ';', ...

roxiciobotea210: I added more tests for this. Also - the last test is for an array of strings and each string is a different combination of sensitive data with spaces and ; , : =
I am out of office today for personal stuff, but in case something else is missing I will add it on Monday. Same for resolving other comments which were not resolved yet.

src/storage/mongodb/LoggingStorage.ts
roxiciobotea210: I checked most of the project code and noticed that `pushConvertObjectIDToString()` is only used in the **get** methods, where an aggregation is built. 
This is the `saveLog()` method and I noticed that for the **save** methods that return an id, this is being done as I initially wrote it - `return logMDB._id.toHexString()` in this case. (You can check saveSite, saveAsset, saveInvoice etc. ; they all use `.toHexString()` on the _id for the return value.) 
Is it ok if I revert to this approach here too, since it is how the code is wrote in most of the other save methods? I need the id as a string anyway so I am doing this conversion sooner or later.
Or is this a special case?

jerome-benoit: Fine, keep it as it's done currently in that PR. 

The ObjectID conversion to string in ev-server should be audit later on because: 
+ information are lost and it made debugging a lot less efficient: https://docs.mongodb.com/manual/reference/method/ObjectId/
+ the conversion is costly and should only be done if needed to avoid a global information loss on the object instance.
That could be part of another issue and PR.

roxiciobotea210: Understood, thanks.

jerome-benoit (approved):

jerome-benoit (approved):

src/storage/mongodb/LoggingStorage.ts
LucasBrazi06: I think that the _id should be returned by the insertOne() method, right?

src/utils/Logging.ts
```diff
@@ -702,68 +702,71 @@ export default class Logging {
...
+        message = dataParts.join('&');
+        return message;
+      }
+      // Check if it is a string, but not a query string
       for (const sensitiveData of Constants.SENSITIVE_DATA) {
-        // Anonymize
-        message.replace(new RegExp(sensitiveData, 'gi'), Constants.ANONYMIZED_VALUE);
+        if (new RegExp(sensitiveData + '(\\s*)[:=,;](\\s*)(.*)', 'gi').test(message)) { // If it contains sensitive data and it matches the RegExp, Anonymize
+          message = message.substring(0, sensitiveData.length + 1) + Constants.ANONYMIZED_VALUE;
+          return message;
```
LucasBrazi06, line 739: If you have a string like:
'password=mypass+name=serge'
It'll end up like this if I'm not wrong (name is before password in the list of anon fields):
'password=mypass+name=####'
To avoid this, I would simply anonymize the whole string.
Let me know.

src/storage/mongodb/LoggingStorage.ts
jerome-benoit: > I think that the _id should be returned by the insertOne() method, right?

It's returned also as part of the data structure returned `InsertOneWriteOpResult<WithId<TSchema>>` but it has the same value as the one generated by ObjectId() instantiation in the method.

src/utils/Logging.ts
jerome-benoit: The code should anonymize everything after password= and name=, the regexp is using 'gi' flag. 
But strings combination of several replacement at once shall be added to the unit tests.

LucasBrazi06: Then also fixed in the code above which returns the string after the first replacement I think

src/storage/mongodb/LoggingStorage.ts
LucasBrazi06: Where is the ObjectId() instantiation in the method?

image

Does the MongoDB insertOne() method adds it in the logMDB object?

src/utils/Logging.ts
roxiciobotea210: The string you are mentioning ('password=mypass+name=serge') is indeed badly obfuscated at the moment. That is because '+' is not treated as a separator in strings (like '&' is). Should I treat it as one and implement that logic + add corresponding tests?

roxiciobotea210: Also - yes, the code you are mentioning is returning the string after the first replacement, because it does not contain '&' (is not a query string).

LucasBrazi06: I think that you just have to replace the code below:
```
if (new RegExp(sensitiveData + '(\\s*)[:=,;](\\s*)(.*)', 'gi').test(message)) { // If it contains sensitive data and it matches the RegExp, Anonymize
          message = message.substring(0, sensitiveData.length + 1) + Constants.ANONYMIZED_VALUE;
          return message;
```

by this one:

```
if (new RegExp(sensitiveData + '(\\s*)[:=,;](\\s*)(.*)', 'gi').test(message)) { // If it contains sensitive data and it matches the RegExp, Anonymize
          return Constants.ANONYMIZED_VALUE;
```

src/storage/mongodb/LoggingStorage.ts
jerome-benoit: ObjectId() is called inside insertOne() and the object saved is passed as a reference, not value.

src/utils/Logging.ts
roxiciobotea210: Done + added specific test for checking query string obfuscation is done properly.

PR #2114, opened 2021-02-16
Check OCPI PushLocationsTask

Jakob8:
I created this Pull Request to investigate if the PushLocationsTask only pushes the EVSE Statuses and not all EVSEs / Locations

### DIFF
--- a/src/scheduler/tasks/ocpi/OCPIPushLocationsTask.ts
+++ b/src/scheduler/tasks/ocpi/OCPIPushLocationsTask.ts
@@ -67,6 +67,9 @@ export default class OCPIPushLocationsTask extends SchedulerTask {
         });
         // Build OCPI Client
         const ocpiClient = await OCPIClientFactory.getCpoOcpiClient(tenant, ocpiEndpoint);
+        // TODO: Investigate if the PushLocationsTask only pushes the EVSE Statuses and not all EVSEs / Locations
+        // CpoOcpiClient has no method sendLocations or pushLocation
+        // sendEVSEStatuses() -> patchEVSEStatus() only contains the EVSE status in the payload and no location
         // Send EVSE statuses
         const sendResult = await ocpiClient.sendEVSEStatuses(!Utils.isUndefined(config.processAllEVSEs) ? config.processAllEVSEs : false);
         Logging.logInfo({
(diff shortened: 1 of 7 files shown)


src/scheduler/tasks/ocpi/OCPIPushLocationsTask.ts
```diff
@@ -67,6 +67,9 @@ export default class OCPIPushLocationsTask extends SchedulerTask {
         });
         // Build OCPI Client
         const ocpiClient = await OCPIClientFactory.getCpoOcpiClient(tenant, ocpiEndpoint);
+        // TODO: Investigate if the PushLocationsTask only pushes the EVSE Statuses and not all EVSEs / Locations
+        // CpoOcpiClient has no method sendLocations or pushLocation
+        // sendEVSEStatuses() -> patchEVSEStatus() only contains the EVSE status in the payload and no location
         // Send EVSE statuses
         const sendResult = await ocpiClient.sendEVSEStatuses(!Utils.isUndefined(config.processAllEVSEs) ? config.processAllEVSEs : false);
```
Jakob8, line 74: pushLocationsTask only pushes the EVSE Statuses

jerome-benoit: The locations handling is in the ocpiClient.checkLocations() and is used in the matching task OCPICheckLocationsTask. 
That task is just not correctly named and shall be renamed to OCPIPushEVSESStatusesTask.ts. 

Could you please do the renaming and audit all code paths using ocpiClient.sendEVSEStatuses() to ensure it's used with the right naming? (update also configuration files) 

The PUT on the location endpoint to add/update locations is not implemented and currently Gireve fetch the locations on a regular basis by using the CPO endpoint on our OCPI server.

jerome-benoit (changes requested):
Please have a look at the comment made.

src/scheduler/tasks/ocpi/OCPIPushLocationsTask.ts
Jakob8: done. renamed the objects from pushLocation to pushEVSEStatuses

jerome-benoit: Thanks. 

I think you can undraft the PR and put in the review pipeline.

PR #2123, opened 2021-02-22
OCPI - get Charging Station by ChargeBoxSerialNumber

Jakob8:
EMSP Implementation:
location_id and evse_uid are received from Gireve and stored as chargeBoxSerialNumber = (location_id + ‘-‘ + evse_uids)

Fixed getChargingStation: option to filter by chargeBoxSerialNumber

___________________________________________________________________________
Problem: 
Hi Jakob,

As discussed, we have an issue in the current eMSP implementation.
The token will never be allowed by the eMSP tenant because we can’t retrieve the charging station.

We receive the following body when Gireve asks us to authorize a token
{
     "location_id": "37534",
     "evse_uids": [
             "121225"
     ]
}
Thanks to this data, we are able to build a unique id for the charge point (location_id + ‘-‘ + evse_uids). 
This identifier is mapped to the chargeBoxSerialNumber, not to the charging station id. 

 

So, you have to update the code to retrieve a charging station by chargeBoxSerialNumber, not by id.

 

https://github.com/Charge-Angels/ev-server/blob/6bb38b43d9192b057a7ecf541fb05d42c096eaff/src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/EMSPTokensEndpoint.ts#L122

Be careful, the buildChargingStationId method is used several times. You may have to update this code also.

 

Hope this helps,
Sébastien


src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/EMSPLocationsEndpoint.ts
```diff
@@ -76,7 +76,7 @@ export default class EMSPLocationsEndpoint extends AbstractEndpoint {
     }
     if (evseUid) {
       const chargingStationId = OCPIUtils.buildChargingStationId(locationId, evseUid);
-      const chargingStation = await ChargingStationStorage.getChargingStation(tenant.id, chargingStationId);
+      const chargingStation = await ChargingStationStorage.getChargingStation(tenant.id, Constants.UNKNOWN_STRING_ID, {}, null, chargingStationId);
```
LucasBrazi06, line 79: Nope, create a new method getChargingStationBySerialNumber(...)

src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/EMSPLocationsEndpoint.ts
```diff
@@ -218,7 +218,7 @@ export default class EMSPLocationsEndpoint extends AbstractEndpoint {
         module: MODULE_NAME, method: 'updateLocation',
         detailedMessages: location
       });
-      await ChargingStationStorage.deleteChargingStation(tenant.id, chargingStationId);
+      await ChargingStationStorage.deleteChargingStation(tenant.id, Constants.UNKNOWN_STRING_ID, chargingStationId);
```
LucasBrazi06, line 221: Also add a new method here.

LucasBrazi06 (changes requested):
Check my comments

src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/EMSPLocationsEndpoint.ts
Jakob8: Ok, I changed it

src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/EMSPLocationsEndpoint.ts
Jakob8: Ok, I changed it

LucasBrazi06 (approved):

PR #2130, opened 2021-02-24
Update registration token site admin part

cabch:


src/server/rest/v1/service/RegistrationTokenService.ts
```diff
@@ -58,6 +58,16 @@ export default class RegistrationTokenService {
         user: req.user
       });
     }
+    // Check site is provided for site admins
+    if (Authorizations.isSiteAdmin(req.user) && !filteredRequest.siteAreaID) {
+      throw new AppError({
+        source: Constants.CENTRAL_SERVER,
+        errorCode: HTTPError.BAD_REQUEST,
```
LucasBrazi06, line 65: Use FORBIDDEN

src/server/rest/v1/service/RegistrationTokenService.ts
```diff
@@ -58,6 +58,16 @@ export default class RegistrationTokenService {
         user: req.user
       });
     }
+    // Check site is provided for site admins
+    if (Authorizations.isSiteAdmin(req.user) && !filteredRequest.siteAreaID) {
+      throw new AppError({
```
LucasBrazi06, line 63: Use AppAuthError exception for authorization

src/server/rest/v1/service/RegistrationTokenService.ts
```diff
@@ -145,7 +155,7 @@ export default class RegistrationTokenService {
     UtilsService.assertObjectExists(action, registrationToken, `Registration Token '${tokenID}' does not exist`,
       MODULE_NAME, 'handleDeleteRegistrationToken', req.user);
     // Check auth
-    if (!Authorizations.canDeleteRegistrationToken(req.user, registrationToken.siteArea?.siteID)) {
+    if (!Authorizations.canDeleteRegistrationToken(req.user, registrationToken?.siteArea?.siteID)) {
```
LucasBrazi06, line 158: No need the first ?, registrationToken is already checked with method assertObjectExists above

src/types/HTTPError.ts
```diff
@@ -60,6 +60,8 @@ export enum HTTPError {
 
   CANNOT_ACQUIRE_LOCK = 510,
   INVALID_FILE_FORMAT = 583,
+
+  BAD_REQUEST = 400,
```
LucasBrazi06, line 64: No need this one

LucasBrazi06 (changes requested):

src/server/rest/v1/service/RegistrationTokenService.ts
cabch: I will replace with this :

`if (Authorizations.isSiteAdmin(req.user) && !filteredRequest.siteAreaID) {`
`      throw new AppAuthError({`
`        errorCode: HTTPAuthError.FORBIDDEN,`
`        user: req.user,`
`        action: Action.CREATE, entity: Entity.TOKEN,`
`        module: MODULE_NAME, method: 'handleCreateRegistrationToken',`
`      });`
`    }`

But i thought it was more of a bad request as it misses the siteID for this request with the siteID user, it's not handled like user hasn't the rights as a "!Authorizations.canCreateRegistrationToken()"

LucasBrazi06: In that case put 2 'if' statements, one for the Auth and one for the missing param.
For the missing param, today we return a general exception (500), eg: assertIdIsProvided()...
This has to be changed everywhere if we go for that error code (BAD_REQUEST) for missing params :-)

cabch: Ok i understood !! I made the changes for Forbidden hope it's ok now

src/server/rest/v1/service/RegistrationTokenService.ts
cabch: done

src/server/rest/v1/service/RegistrationTokenService.ts
cabch: done I forgot to delete it after re ordering the code

src/types/HTTPError.ts
cabch: deleted and handle error as forbidden

src/server/rest/v1/service/RegistrationTokenService.ts
```diff
@@ -58,6 +58,15 @@ export default class RegistrationTokenService {
         user: req.user
       });
     }
+    // Check site is provided for site admins
+    if (Authorizations.isSiteAdmin(req.user) && !filteredRequest.siteAreaID) {
```
LucasBrazi06: Only if the ORG component is active ;-)

src/server/rest/v1/service/RegistrationTokenService.ts
LucasBrazi06: Do not mark the conversation solved yourself (golden rule 44)

LucasBrazi06 (approved):

src/server/rest/v1/service/RegistrationTokenService.ts
cabch: it's for security reason right ? because the tab is not active if ORG not active this is why I didn't think about it

LucasBrazi06: No, If the Org is not active there will be no checks with the Site Admin role but the Registration Token is still visible for the Admin

PR #2136, opened 2021-02-25
Notifications activate user account manually back

cabch:
Sends user url to activate their account
Sends admin url to verify user account

**& Handled tenants logos** in email headers from this issue (cannot link multiple issues to this PR as far as I can see) [https://github.com/Charge-Angels/ev-server/issues/2119](url)


LucasBrazi06 (commented):
When user auto activation is not set, the Admin does not receive the email to activate him (nothing also in the logs)
Could you check?
I will merge the PR in //

PR #2137, opened 2021-02-28
Fix simple pricing computation

ellavas:
Remove the truncate on transaction update.


src/integration/pricing/simple-pricing/SimplePricingIntegration.ts
```diff
@@ -23,6 +23,7 @@ export default class SimplePricingIntegration extends PricingIntegration<SimpleP
     return this.computePrice(transaction, consumptionData);
   }
 
+  // eslint-disable-next-line @typescript-eslint/require-await
```
jerome-benoit, line 26: Please, do not blindly disable eslint rules: you just have to remove the async and the Promise<...> to properly fix the code.

src/utils/Utils.ts
```diff
@@ -475,11 +475,11 @@ export default class Utils {
   }
 
   public static computeSimplePrice(pricePerkWh: number, consumptionWh: number): number {
-    return Utils.truncTo(pricePerkWh * (consumptionWh / 1000), 6);
+    return Utils.roundTo(pricePerkWh * (consumptionWh / 1000), 6);
```
jerome-benoit, line 478: You can't round here: rounding means you charge the user for not existent consumption and that's illegal.

src/utils/Utils.ts
```diff
@@ -475,11 +475,11 @@ export default class Utils {
...
   }
 
   public static computeSimpleRoundedPrice(pricePerkWh: number, consumptionWh: number): number {
-    return Utils.truncTo(pricePerkWh * (consumptionWh / 1000), 2);
+    return Utils.roundTo(pricePerkWh * (consumptionWh / 1000), 2);
```
jerome-benoit, line 482: Same here. The helper method shall be renamed by the way.

jerome-benoit (changes requested):

src/integration/pricing/simple-pricing/SimplePricingIntegration.ts
ellavas: This can't be modified.
The simple pricing extends the abstract pricing which requires async methods.

src/utils/Utils.ts
ellavas: Please check the issue.

jerome-benoit: Rounded pricing computation are still not legal in most countries around the world. 
The issue is with the truncTo() helper: it should properly return the truncated values after the comma.

jerome-benoit:
The issue seems to also impact all computed price in the DB, so a task similar to RecomputeAllTransactionsConsumptionsTask shall be added and run to recompute all prices in DB.

src/utils/Utils.ts
LucasBrazi06: The fix in OCPPUtils and in SimplePricingIntegration classes above should be sufficient to solve this issue.
In this method I would keep all the decimals so we could remove the Utils call.

src/utils/Utils.ts
LucasBrazi06: The fix in OCPPUtils and in SimplePricingIntegration classes above should be sufficient to solve this issue.
Use of the truncTo() should not impact the total price of the transaction which is based on the computeSimplePrice() one.
You can revert to truncTo() after checking.

LucasBrazi06: @ellavas Did you get the last comments?

src/utils/Utils.ts
ellavas: I agree to keep all the decimals here. Can we have problems with the database?

src/utils/Utils.ts
ellavas: I would prefer to remove this method and call the truncTo method on the computed price.

src/utils/Utils.ts
LucasBrazi06: > I agree to keep all the decimals here. Can we have problems with the database?

I don't think so  :-)

> MongoDb added support for Decimal data type in 3.4 version. It is also available from the shell. 3.4 adds support for the decimal128 format with the new decimal data type. The decimal128 format supports numbers with up to 34 decimal digits (i.e. significant digits) and an exponent range of −6143 to +6144

src/utils/Utils.ts
LucasBrazi06: The method was there because it is used somewhere else I think.

LucasBrazi06 (approved):

PR #2139, opened 2021-03-01
Fix and update cipher algorithms support

jerome-benoit:
And use secure default for sensitive data ciphering.

Update: 
+ fix operation mode with authentication;
+ fix random key length generation.


src/types/HTTPError.ts
```diff
@@ -41,7 +41,7 @@ export enum HTTPError {
   SITE_AREA_WITH_NO_SITE_ERROR = 525,
   USER_WITH_NO_SITE_ERROR = 525,
 
-  BILLING_DELETE_ERROR = 510, // Override an already defined status code: Not extended (RFC 2774)
+  BILLING_DELETE_ERROR = 510, // FIXME: override an already defined HTTP status code: Not extended (RFC 2774)
```
LucasBrazi06, line 44: No need to be unique, each code will be used in a dedicated area (Billing...) as we have only 99 errors possible (500-599.)
Only general HTTP Code should be.

src/utils/Utils.ts
```diff
@@ -1352,15 +1352,15 @@ export default class Utils {
...
+    // TODO: change 16 to 32 and test on Mac
     return crypto.randomBytes(16).toString('hex');
   }
 
   public static getDefaultKeyProperties(): CryptoKeyProperties {
     return {
       blockCypher: 'aes',
       blockSize: 256,
-      operationMode: 'ctr'
+      operationMode: 'gcm'
```
LucasBrazi06, line 1363: This option is not working:
- Create a tenant
- Set up a connection in the Asset (where there is sensitive data) 
- Check the connection -> it fails

I replaced with 'ctr', set the password again in the Asset setting and it works.

Could you double check?

LucasBrazi06 (changes requested):

src/types/HTTPError.ts
jerome-benoit: The chosen code shall not override an already existing HTTP status code with a complete different meaning. It shall reuse an already existing with the same meaning or define its own not yet reserved then.

LucasBrazi06: > The chosen code shall not override an already existing HTTP status code with a complete different meaning. It shall reuse an already existing with the same meaning or define its own not yet reserved then.

And where there will be all code taken what will you do!?

jerome-benoit: HTTP code shall only be used to return HTTP protocol related behaviour ;) 
If you need to return application specific errors, an error message shall be returned with application own error code and error message, like in OCPP for example.

src/utils/Cypher.ts
```diff
@@ -7,36 +8,48 @@ import LockingManager from '../locking/LockingManager';
...
-
   public static async encrypt(tenantID: string, data: string, useFormerKey = false, cryptoSetting?: CryptoSetting): Promise<string> {
     const iv = crypto.randomBytes(IV_LENGTH);
     if (!cryptoSetting) {
       cryptoSetting = (await Cypher.getCryptoSettings(tenantID)).crypto;
     }
     const algo = useFormerKey ? Utils.buildAlgorithm(cryptoSetting.formerKeyProperties) : Utils.buildAlgorithm(cryptoSetting.keyProperties);
     const key = useFormerKey ? Buffer.from(cryptoSetting.formerKey) : Buffer.from(cryptoSetting.key);
-    const cipher = crypto.createCipheriv(algo, key, iv);
+    const cipher: CipherGCM = crypto.createCipheriv(algo, key, iv) as CipherGCM;
```
jerome-benoit, line 23: The type definition and the cast is a trick to make TS recognise the authentication related methods.

src/types/HTTPError.ts
LucasBrazi06: > HTTP code shall only be used to return HTTP protocol related behaviour ;)
> If you need to return application specific errors, an error message shall be returned with application own error code and error message in the body, like in OCPP for example.

Okay, then we have another place for a small refactoring then :-).

src/utils/Cypher.ts
LucasBrazi06: > The type definition and the cast is a trick to make TS recognise the authentication related methods.

Should it be removed as we don't rely to a specific algo?

jerome-benoit: > > The type definition and the cast is a trick to make TS recognise the authentication related methods.
> 
> Should it be removed as we don't rely to a specific algo?

It can't be removed, otherwise the transpilation fails: CipherGCM type have the `(get|set)AuthTag()` method, Cipher don't. 
Enforcing that mode type (or CipherCCM, CipherOFB, ...) with authentication just make the transpilation works. The code already handle all operation modes with embedded authentication.

LucasBrazi06: Okay, thanks!

LucasBrazi06 (changes requested):
Still not working with gcm:
image
How to reproduce:
- Create a new Tenant
- Enable Asset management
- Activate the Admin user in the new tenant
- Create an Asset connection to GreenCom in the Integration Settings + Save
- Test the created connection: it should work

LucasBrazi06 (approved):

PR #2146, opened 2021-03-02
Sync imported users

mahdibenromdhane:


src/scheduler/tasks/SynchronizeUsersImportTask.ts
```diff
@@ -0,0 +1,106 @@
...
+          for (const importedUser of importedUsers.result) {
+            const foundUser = await UserStorage.getUserByEmail(tenant.id, importedUser.email);
+            if (foundUser) {
+              importedUser.status = UserImportStatus.ERROR;
+              importedUser.error = 'Email already exists';
+              await UserStorage.saveImportedUser(tenant.id, importedUser);
+            } else {
+              const user: Partial<User> = {
+                firstName: importedUser.firstName,
+                name: importedUser.name.toUpperCase(),
```
ClaudeROSSI, line 33: Do we really need the toUpperCase() here ? ... if we need it we also need to pass a locale() ... otherwise the result is a bit unpredictable!

src/scheduler/tasks/SynchronizeUsersImportTask.ts
```diff
@@ -0,0 +1,106 @@
...
+                email: importedUser.email.toLowerCase(),
+                locale: null, // Defaults to the browser locale
+                issuer: true,
+                deleted: false,
+                role: UserRole.BASIC,
+                status: UserStatus.ACTIVE,
```
ClaudeROSSI, line 39: I'm not sure user should be activated by default. I guess this should take into account some tenant settings. Caro has been working on that aspect!

src/scheduler/tasks/SynchronizeUsersImportTask.ts
```diff
@@ -0,0 +1,106 @@
...
+                createdBy: importedUser.importedBy,
+                createdOn: new Date(),
+                passwordWrongNbrTrials: 0,
+                notificationsActive: true,
+                notifications: {
```
ClaudeROSSI, line 44: As already discussed, I do not feel confortable with these default values.
To be on the safe side, I think we should have an helper method providing the exhaustive list of properties and their default values. Of course, the import may override some of them using its own logic.

src/scheduler/tasks/SynchronizeUsersImportTask.ts
```diff
@@ -0,0 +1,106 @@
...
+                  sendSmtpError: false,
+                  sendBillingNewInvoice: false,
+                  sendComputeAndApplyChargingProfilesFailed: false,
+                  sendAccountVerificationNotification: false,
+                  sendAdminAccountVerificationNotification: false
+                },
+              };
+              try {
+                UserValidator.getInstance().validateUserCreation(user);
+                await global.database.getCollection<any>(tenant.id, 'users')
```
ClaudeROSSI, line 71: ??? I do not understand why we do not call here the regular method for saving a user?
i.e.: UserStorage.saveUser(...). 
BTW: That method is supposed to take care of settings default values (for the properties that you do not provide). The code would then be shorter and easier. right?

src/storage/mongodb/UserStorage.ts
```diff
@@ -275,15 +275,21 @@ export default class UserStorage {
   public static async saveImportedUser(tenantID: string, userToSave: any): Promise<void> {
```
ClaudeROSSI, line 275: userToSave: any

Can't we have a type here? importedUser?

Same comment on UserStorage.getImportedUser() method

src/storage/mongodb/UserStorage.ts
```diff
@@ -666,6 +672,103 @@ export default class UserStorage {
     };
   }
 
+  public static async getImportedUsers(tenantID: string,
+    params: {
+      statuses?: UserImportStatus[]; search?: string
+    },
+    dbParams: DbParams, projectFields?: string[]): Promise<DataResult<any>> {
```
ClaudeROSSI, line 679: please add an ImportedUser type. to avoid the <any> here!

ClaudeROSSI (changes requested):
I think we need at least a change for the call to UserStorage.saveUser().
Let me know what you think!

src/scheduler/tasks/SynchronizeUsersImportTask.ts
mahdibenromdhane: It fixed since we use the userStorage to save the imported user

src/scheduler/tasks/SynchronizeUsersImportTask.ts
mahdibenromdhane: you'r right, I'v replaced it with storage functions.

src/storage/mongodb/UserStorage.ts
mahdibenromdhane: Done

src/scheduler/tasks/SynchronizeUsersImportTask.ts
mahdibenromdhane: I think the user need to be active if we are going to mass insert from a file or the admin will have to activate them one by one.

src/storage/mongodb/UserStorage.ts
mahdibenromdhane: Done

src/scheduler/tasks/SynchronizeUsersImportTask.ts
mahdibenromdhane: I'v removed it from here. because it's already converted to uppercase in the upload.
As discussed the name of the user is always uppercase in the project.

src/server/rest/v1/service/UserService.ts
```diff
@@ -1281,10 +1281,9 @@ export default class UserService {
         name: user.Name,
         firstName: user.First_Name,
         email: user.Email,
-        role: user.Role,
         importedBy: req.user.id,
         status: UserImportStatus.UNKNOWN
-      };
+      } as ImportedUser;
```
ClaudeROSSI, line 1286: can't we avoid the "as" here?

does it work if you type the new object like below?

const newUploadedUser: ImportedUser = {
        name: user.Name,
        firstName: user.First_Name,
        email: user.Email,
        importedBy: req.user.id,
        status: UserImportStatus.UNKNOWN
      };

ClaudeROSSI (changes requested):
please have a look to my latest comment regarding "as" being used ... where it does not seem to be useful!

src/server/rest/v1/service/UserService.ts
mahdibenromdhane: Yes it work, done.

ClaudeROSSI (approved):
I think we can merge this one ... even though we got new feedback from Serge. We'll take them into account on the next round!

LucasBrazi06:
> I think we can merge this one ... even though we got new feedback from Serge. We'll take them into account on the next round!

Done

PR #2148, opened 2021-03-03
Security concept company

roxiciobotea210:
Adds dynamic filters on Company requests (GET by id & GET all/list, UPDATE, DELETE).

Still need to change the token data.

Open points:
* should there be any dynamic filtering on handleGetCompanyLogo?
* is there anything needed for handleCreateCompany?
* is anything missing in general?
* what should tests take into consideration?
* besides companyIDs, what should be removed from the token?


LucasBrazi06 (approved):
Removed IDs in static checks (canXXX())

PR #2156, opened 2021-03-08
Crypto update tests

roxiciobotea210:
Added unit tests to check that:
* key + block size change works;
* key + operation change works;
* key + algo change works;
* key + algo + block size change works;
* key + block size + operation change works;
* key + algo + operation mode change works;

Failure test cases check the following statuses are returned correctly:
* CRYPTO_KEY_LENGTH_INVALID (513)
* CRYPTO_ALGORITHM_NOT_SUPPORTED (512)

Used the following values in tests (in random combinations):
* operation mode: ctr, gcm, ccm, ofb
* block size: 128, 192, 256
* algorithm: aes, camellia

**Open points/not sure about:**
* how to test that default crypto properties are working?
* in the current tests, after the encryption information is updated, should I check anything else?
* how to check failures for CRYPTO_MIGRATION_IN_PROGRESS (511) & CRYPTO_CHECK_FAILED (514)?


test/api/SettingTest.ts
```diff
@@ -21,6 +22,98 @@ chai.use(responseHelper);
...
+            "blockSize" : "${originalCryptoObject.content.crypto.keyProperties.blockSize}",
+            "operationMode" : "${originalCryptoObject.content.crypto.keyProperties.operationMode}"
+        },
+        "migrationToBeDone" : true
+      }
+    }
+}`);
+}
+
+async function getCurrentCryptoData() {
```
jerome-benoit, line 83: All following methods do more than just get, reset or update, they are also doing some validation, checks of the outputs. Their names shall reflect it.

test/api/SettingTest.ts
```diff
@@ -21,6 +22,98 @@ chai.use(responseHelper);
 const testData: TestData = new TestData();
 let initialTenant: Tenant;
 
+// Update pricing setting to have sensitive data to test on it
+async function updatePricingWithSensitiveData():Promise<any> {
```
jerome-benoit, line 26: It's more than just an update operation, it also includes outputs validation for mocha. Its name shall reflect it.

test/api/SettingTest.ts
```diff
@@ -211,5 +248,252 @@ describe('Setting tests', function() {
...
+      // Update operation mode to ctr
+      await updateCryptoSettings(read.data.result[0], read.data.result[0].content.crypto.keyProperties.blockCypher,
+        read.data.result[0].content.crypto.keyProperties.blockSize, 'ctr');
+    });
+    it('Check crypto settings update - change key + algorithm (aes->camellia)', async () => {
+      // Retrieve the crypto setting id
+      const read = await getCurrentCryptoData();
+      // Ensure current algorithm is aes + ensure operation mode works with both camellia & aes
+      await updateCryptoSettings(read.data.result[0], 'aes',
+        read.data.result[0].content.crypto.keyProperties.blockSize, 'ctr');
```
jerome-benoit, line 306: I would prefer that you use our default mode 'gcm' for key + algo change unit tests, we get then some more coverage on our defaults: read.data.result[0].content.crypto.keyProperties.operationMode should give it.

test/api/SettingTest.ts
```diff
@@ -211,5 +248,252 @@ describe('Setting tests', function() {
 
       const clientSecretAfter = _.get(readSettingAfter.data.result[0], readSettingAfter.data.result[0].sensitiveData[0]);
     });
+    it('Check crypto settings update - change key + block size (256->128)', async () => {
+      // Retrieve the crypto setting id
+      const read = await getCurrentCryptoData();
```
jerome-benoit, line 253: I think you can rename 'read' to something more meaningful like cryptoData, currentCryptoData or similar after the rename of `getCurrentCryptoData()` ;)

test/api/SettingTest.ts
```diff
@@ -21,6 +22,98 @@ chai.use(responseHelper);
...
+            "policyId" : "1119",
+            "reportName" : "E-Car Charging"
+        }
+    }
+  }`);
+  const update = await testData.centralService.updateEntity(testData.centralService.settingApi, testData.data);
+  expect(update.status).to.equal(200);
+}
+
+function getCryptoTestData(originalCryptoObject, newKeyProperties:CryptoKeyProperties) {
```
jerome-benoit, line 57: getCryptoTestSettings

jerome-benoit (changes requested):
Some nitpicks ;)

Thanks.

test/api/SettingTest.ts
roxiciobotea210: That is how I initially tried, but we do not support gcm operation mode for camellia algorithm, which is why I changed it for this test.

jerome-benoit: Use idea algo as a target then in these tests.

roxiciobotea210: 'idea' algo still doesn't work with 'gcm'.
image
'aria' would work best. 
There's also this one: 
image
But here i am running into a different kind of error... looks like if i send as block size 'aes128' it fails because it would expect the block size to be a number.

LucasBrazi06 (approved):
Excellent!

PR #2159, opened 2021-03-08
Lock acquisition with timeout implementation.

jerome-benoit:
@OliveGerste have checked that's what he's expecting

Unit tests will be added in a later PR. 


src/locking/LockingManager.ts
```diff
@@ -55,11 +55,62 @@ export default class LockingManager {
...
+          while (!timeoutReached) {
+            try {
+              await LockingStorage.insertLock(lock);
+              break;
+            } catch {
+              await Utils.sleep(1000);
+            }
+          }
+          if (timeoutReached) {
+            throw Error(`Lock acquisition timeout ${timeout}ms reached`);
```
OliveGerste, line 76: Should we also log something here?

OliveGerste (commented):
I tested it in my local environment and it worked. 
To see if it is working in real life, we could use it in the smart charging integration and start two transactions at the exact same time. 
If smart charging runs twice, it works.

src/locking/LockingManager.ts
jerome-benoit: The catch should log error thrown.

jerome-benoit:
> I tested it in my local environment and it worked.
> To see if it is working in real life, we could use it in the smart charging integration and start two transactions at the exact same time.
> If smart charging runs twice, it works.

I will try to find time to add some unit tests, which might be a bit tricky to implement, unless you need it urgently and then I create another issue to add the unit test later on.  

I'm not very happy with the busy loop wait and retry (CPU hog) but I think it should be enough. The other idea I have was to wait and retry by looping over an array of promises trying to get the lock.

LucasBrazi06 (approved):
Gut!

LucasBrazi06 (approved):

PR #2176, opened 2021-03-12
Fixed tags tests

Ninjeneer:


test/api/UserTest.ts
```diff
@@ -207,7 +208,7 @@ describe('User tests', function() {
         it('Should be able to delete a badge that has not been used', async () => {
           testData.newTag = Factory.tag.build({ userID: testData.newUser.id });
           let response = await testData.userService.userApi.createTag(testData.newTag);
-          expect(response.status).to.equal(200);
+          expect(response.status).to.equal(StatusCodes.CREATED);
```
LucasBrazi06, line 211: Why have we now 201 instead of 200?
Will it work with the front-end?
Thanks.

Ninjeneer: Because it's a tag creation, so we need to return a 201 code instead of 200, as HTTP spec describe it
It works with the front-end, nothing to change on that side

LucasBrazi06: Perfect, thanks!

PR #2177, opened 2021-03-12
REST > Fix Ids in chargingstations routes

Ninjeneer:
- Rolled back IDs in charging stations routes
- Fixed parameters for QRCode download


LucasBrazi06 (approved):
Gut!

PR #2179, opened 2021-03-12
Billing - Stripe - 1st version

ClaudeROSSI:
A first version for the immediate billing use-case:

- **immediateBillingAllowed** must be set in billing settings
- generates an invoice per transaction
- finalize the invoice immediately
- tries to charge the invoice using the default payment method of the customer
- can apply taxes (VAT) when taxID is properly set (no UI for doing it yet)
- send mails to inform the user about the status of the new invoice
- Invoice line item: Quantity is always set to 1, line amount corresponds to the **roundedPrice** of the transaction
- The link between Users and Stripe Customer does not rely on the email anymore. 
- The link between Users and Stripe Customer cannot be repaired automatically if broken (when deleting a customer for instance on stripe side)
- Invoice Synchronization is implicit. Triggering a manual sync should be useless.
- ACHTUNG! - The billing settings are very sensitive to changes. Switching from a STRIPE account to another one may break the data consistency.
- The end user can add his/her card via the UI (from Caroline) - The latest created payment method becomes the default payment method of the customer (in STRIPE DB). This payment is the only one used when trying to charge the invoice.
- Invoice PDF document is downloaded and persisted on our side as soon as the invoice is finalized (and cannot be changed anymore)
- WebHooks (listening for stripe events) are not yet used - just logging some information
- Some end-points are in comment because they are not yet used
- Billing Testing:
-  2 categories of tests:
- - Some tests are testing Stripe Capabilities directly - they call the Stripe Implementation
- - Other tests are testing the complete flow via the regular interfaces (REST endpoints) and are very sensitive to the initial test data - In other words, the STRIPE account should not contain inconsistent data when running the tests


src/server/rest/v1/service/BillingService.ts
```diff
@@ -551,4 +591,84 @@ export default class BillingService {
       });
     }
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  public static async handleBillingChargeInvoice(action: ServerAction, req: Request, res: Response): Promise<void> {
```
ClaudeROSSI, line 596: Not yet used ... will be put back when we provide the ability to pay with a different card!

src/server/rest/v1/service/BillingService.ts
```diff
@@ -551,4 +591,84 @@ export default class BillingService {
...
+    // // Invoke the billing implementation (stripe)
+    // let billingInvoice:BillingInvoice = await billingImpl.getInvoice(invoice.invoiceID);
+    // billingInvoice = await billingImpl.chargeInvoice(billingInvoice);
+    // res.json({
+    //   succeeded: true,
+    //   status: billingInvoice.status
+    // });
+  }
+
+  public static async handleBillingWebHook(action: ServerAction, req: Request, res: Response): Promise<void> {
```
ClaudeROSSI, line 639: WebHook are not yet used - just logging the stripe events for now!

LucasBrazi06 (approved):

LucasBrazi06:
@ClaudeROSSI I can merge it if you want (still in Draft)
Let me know.

ClaudeROSSI:
> @ClaudeROSSI I can merge it if you want (still in Draft)
> Let me know.

I think we can merge ... but I'm not sure about the impact on the tests ... I have declared billing.taxID has a boolean property in the test/config.js file. I have not been able to run the tests on my side!

LucasBrazi06:
> Test are only ok when executed whithin VSCode and on my linux machine ?!? I'm lost!
> Green on the left, Almost green on the right side!

LOL !!!
Got them ok on your branch, I'll merge :)
image

PR #2182, opened 2021-03-15
REST > Missing 2 chargingstations endpoints

Ninjeneer:
2 endpoints were missing :
- Smart charging trigger (GET /chargingstations/smartcharging/trigger)
- Charginstation OCPP configuration update (PUT /chargingstations/{id}/configuration)

Also made minor improvements on doc parameters


src/assets/server/rest/v1/docs/e-mobility-oas.json
```diff
@@ -2645,7 +2679,7 @@
             "$ref": "#/components/parameters/SiteID"
           },
           {
-            "$ref": "#/components/parameters/SiteAreaID"
+            "$ref": "#/components/parameters/SiteAreaIDs"
```
LucasBrazi06, line 2682: SiteAreaID should always be singular like any other query params where we can pass several values that are separated by a pipe.

LucasBrazi06 (changes requested):

src/assets/server/rest/v1/docs/e-mobility-oas.json
Ninjeneer: Actually, the Query param's name itself is not modified. I just changed the name of the param in the oas file to differentiate Single ID / Multiple IDs.
As the TriggerSmartCharging endpoint focus on one SiteAreaID, we should not use a param labeled as "Pipe separated IDs"

LucasBrazi06: > Actually, the Query param's name itself is not modified. I just changed the name of the param in the oas file to differentiate Single ID / Multiple IDs.
> As the TriggerSmartCharging endpoint focus on one SiteAreaID, we should not use a param labeled as "Pipe separated IDs"

Okay got it, then we should also align the naming of all the other query params ;-)

Ninjeneer: Indeed! Done

LucasBrazi06 (approved):

PR #2204, opened 2021-03-17
REST > Get Transactions 

Ninjeneer:
In OAS doc, the `ChargeBoxIDs` parameters is only a temporary fix to avoid changing or breaking front-end/mobile, as on server-side we use sometimes "ChargingStationID" or "ChargeBoxID". 
I will standardize those parameters names when deploying API changes to front-end and mobile, and remove the `ChargeBoxIDs` in OAS file.


src/server/rest/v1/service/TransactionService.ts
```diff
@@ -40,6 +40,28 @@ import moment from 'moment-timezone';
...
+    const filteredRequest = TransactionSecurity.filterGetTransactions(req.query);
+    let transactions: DataResult<Transaction>;
+    if (filteredRequest.Status) {
+      if (filteredRequest.Status === 'completed') {
+        transactions = await TransactionService.getTransactionsCompleted(ServerAction.TRANSACTIONS_COMPLETED, req);
+      } else if (filteredRequest.Status === 'active') {
+        transactions = await TransactionService.getTransactionsActive(ServerAction.TRANSACTIONS_ACTIVE, req);
+      }
+    } else {
+      const completed = await TransactionService.getTransactionsCompleted(ServerAction.TRANSACTIONS_COMPLETED, req);
```
LucasBrazi06, line 54: This does not work.
You break the pagination.

src/server/rest/v1/service/security/TransactionSecurity.ts
```diff
@@ -110,4 +110,10 @@ export default class TransactionSecurity {
     }
     return filteredRequest;
   }
+
+  public static filterGetTransactions(request: any): HttpGetTransactionsRequest {
+    return {
+      Status: sanitize(request.Status),
```
LucasBrazi06, line 116: Should be part of the big http transaction type

src/server/rest/v1/service/TransactionService.ts
```diff
@@ -1071,4 +1085,20 @@ export default class TransactionService {
     );
     return transactions;
   }
+
+  private static async getTransactionsCompleted(action: ServerAction, req: Request): Promise<DataResult<Transaction>> {
+    return await TransactionService.getTransactions(req, action, { completedTransactions: true }, [
```
LucasBrazi06, line 1090: Set the query string Status param to 'completed' and call the getTransactions() in which you can provide a IF to add the proper filter (if Status is not provided return all transactions)

src/server/rest/v1/service/TransactionService.ts
```diff
@@ -1071,4 +1085,20 @@ export default class TransactionService {
...
+      'id', 'chargeBoxID', 'timestamp', 'issuer', 'stateOfCharge', 'timezone', 'connectorId', 'meterStart', 'siteAreaID', 'siteID',
+      'stop.roundedPrice', 'stop.price', 'stop.priceUnit', 'stop.inactivityStatus', 'stop.stateOfCharge', 'stop.timestamp', 'stop.totalConsumptionWh',
+      'stop.totalDurationSecs', 'stop.totalInactivitySecs', 'stop.meterStop', 'billingData.invoiceID', 'ocpi', 'ocpiWithCdr', 'tagID', 'stop.tagID',
+    ]);
+  }
+
+  private static async getTransactionsActive(action: ServerAction, req: Request): Promise<DataResult<Transaction>> {
+    return await TransactionService.getTransactions(req, action, { completedTransactions: false }, [
```
LucasBrazi06, line 1098: Set the query string Status param to 'active' and call the getTransactions() in which you can provide a IF to add the proper filter (if Status is not provided return all transactions)

src/server/rest/v1/service/TransactionService.ts
```diff
@@ -685,22 +707,14 @@ export default class TransactionService {
   }
 
   public static async handleGetTransactionsActive(action: ServerAction, req: Request, res: Response, next: NextFunction): Promise<void> {
-    const transactions = await TransactionService.getTransactions(req, action, { completedTransactions: false }, [
-      'id', 'chargeBoxID', 'timestamp', 'issuer', 'stateOfCharge', 'timezone', 'connectorId', 'status', 'meterStart', 'siteAreaID', 'siteID',
-      'currentTotalDurationSecs', 'currentTotalInactivitySecs', 'currentInstantWatts', 'currentTotalConsumptionWh', 'currentStateOfCharge',
-      'currentCumulatedPrice', 'currentInactivityStatus', 'roundedPrice', 'price', 'priceUnit', 'tagID',
-    ]);
+    const transactions = await TransactionService.getTransactionsActive(action, req);
```
LucasBrazi06, line 710: No need to call a private method here if the change is done above.

src/server/rest/v1/service/TransactionService.ts
```diff
@@ -685,22 +707,14 @@ export default class TransactionService {
...
   }
 
   public static async handleGetTransactionsCompleted(action: ServerAction, req: Request, res: Response, next: NextFunction): Promise<void> {
     // Get transaction
-    const transactions = await TransactionService.getTransactions(req, action, { completedTransactions: true }, [
-      'id', 'chargeBoxID', 'timestamp', 'issuer', 'stateOfCharge', 'timezone', 'connectorId', 'meterStart', 'siteAreaID', 'siteID',
-      'stop.roundedPrice', 'stop.price', 'stop.priceUnit', 'stop.inactivityStatus', 'stop.stateOfCharge', 'stop.timestamp', 'stop.totalConsumptionWh',
-      'stop.totalDurationSecs', 'stop.totalInactivitySecs', 'stop.meterStop', 'billingData.invoiceID', 'ocpi', 'ocpiWithCdr', 'tagID', 'stop.tagID',
-    ]);
+    const transactions = await TransactionService.getTransactionsCompleted(action, req);
```
LucasBrazi06, line 717: No need to call a private method here if the change is done above.

src/server/rest/v1/service/TransactionService.ts
```diff
@@ -40,6 +40,28 @@ import moment from 'moment-timezone';
 const MODULE_NAME = 'TransactionService';
 
 export default class TransactionService {
+
+  public static async handleGetTransactions(action: ServerAction, req: Request, res: Response, next: NextFunction): Promise<void> {
+    const filteredRequest = TransactionSecurity.filterGetTransactions(req.query);
```
LucasBrazi06, line 45: Call right await the getTransactions() with the proper fields projection (related to my other comments below)

LucasBrazi06 (changes requested):

src/server/rest/v1/service/TransactionService.ts
Ninjeneer: I understand how to retrieve completed or active transactions from the same getTransactions() method. However I can't figure out how to return both from the same call, as they need different projections, without breaking pagination.

src/server/rest/v1/service/security/TransactionSecurity.ts
Ninjeneer: Done

src/server/rest/v1/service/TransactionService.ts
Ninjeneer: Done

src/server/rest/v1/service/TransactionService.ts
Ninjeneer: Done

src/server/rest/v1/service/TransactionService.ts
Ninjeneer: Removed private method

src/server/rest/v1/service/TransactionService.ts
Ninjeneer: Removed private method

src/server/rest/v1/service/TransactionService.ts
Ninjeneer: Done

src/server/rest/v1/service/TransactionService.ts
Ninjeneer: Removed

LucasBrazi06 (approved):
GUT !!!

PR #2207, opened 2021-03-19
Site auth

roxiciobotea210:
* get rid of sitesAdmin and sitesOwner usage from user token
* retrieve and user sitesAdmin and sites Owner from db
* add unit tests

Pleas check if permissions & checks are correct for siteAdmin and siteOwner scenarios!!!


src/server/rest/v1/service/AuthorizationService.ts
```diff
@@ -69,16 +69,24 @@ export default class AuthorizationService {
...
     for (const site of sites) {
-      site.canRead = Authorizations.canReadSite(userToken);
-      site.canUpdate = Authorizations.canUpdateSite(userToken, site.id);
-      site.canDelete = Authorizations.canDeleteSite(userToken, site.id);
+      if (userToken.role === UserRole.ADMIN) {
+        site.canRead = true;
+        site.canUpdate = true;
+        site.canDelete = true;
+      } else {
+        site.canRead = Authorizations.canReadSite(userToken);
```
LucasBrazi06, line 86: I would apply the same rule you put in Update for Read and Delete:
`site.canUpdate = Authorizations.canUpdateSite(userToken) && (sitesAdmin.includes(site.id) || sitesOwner.includes(site.id));`
Okay?

src/server/rest/v1/service/AuthorizationService.ts
```diff
@@ -499,6 +507,28 @@ export default class AuthorizationService {
...
+      {
+        userID: userToken.id,
+        siteOwner: true
+      }, Constants.DB_PARAMS_MAX_LIMIT,
+      ['siteID']
+    );
+    return userSites.result.map((userSite) => userSite.siteID);
+  }
+
+  private static async getAssignedUserSites(tenantID: string, userID: string) {
```
LucasBrazi06, line 522: Missing return type

LucasBrazi06 (changes requested):
GUT !!!
Just check my few comments.

src/server/rest/v1/service/AuthorizationService.ts
roxiciobotea210: I just realised this was not actually needed. Commit with refactored code coming soon.

src/server/rest/v1/service/AuthorizationService.ts
roxiciobotea210: I added the same rule for Update&Delete, but I don't understand why it should apply for read? Shouldn't basic users be able to read the site info?

roxiciobotea210:
I added a bit of new logic & I am not sure if it is correct. 
From my point of view, a basic user which is site owner/site admin should not be able to remove himself from the site. Is that correct?

src/server/rest/v1/service/AuthorizationService.ts
LucasBrazi06: > I added the same rule for Update&Delete, but I don't understand why it should apply for read? Shouldn't basic users be able to read the site info?

For Basic user yes, but Basic is a role and it may not be the same in the future roles that are going to be built.
We should have the same approach based on auth regardless the roles.

LucasBrazi06:
> I added a bit of new logic & I am not sure if it is correct.
> From my point of view, a basic user which is site owner/site admin should not be able to remove himself from the site. Is that correct?

He can but he will not have access to it anymore :-)
Then he should request an Admin to put him back.

src/server/rest/v1/service/AuthorizationService.ts
roxiciobotea210: The check for update/delete verifies (in addition to the static check) if the user is siteAdmin/siteOwner because only they are allowed to edit or delete a fite. I didn't add a similar check for reading a site because any user is allowed to read... If i add it, then users which are not owners or admins of the sites will not be able to read.

LucasBrazi06: > The check for update/delete verifies (in addition to the static check) if the user is siteAdmin/siteOwner because only they are allowed to edit or delete a fite. I didn't add a similar check for reading a site because any user is allowed to read... If i add it, then users which are not owners or admins of the sites will not be able to read.

Yes you're right!
We are adding the auth actions for the list here in which the Read is always true.
So you can even set it always to true I think, no need to recheck.

LucasBrazi06 (approved):

PR #2210, opened 2021-03-19
REST > Chargingstation schemas : getById / delete / action

Ninjeneer:


src/assets/server/rest/v1/schemas/chargingstation/chargingstation-action.json
```diff
@@ -0,0 +1,96 @@
...
+  "title": "Charging Station Action Request",
+  "type": "object",
+  "properties": {
+    "chargeBoxID": {
+      "$ref": "chargingstation.json#/definitions/id"
+    },
+    "carID": {
+      "$ref": "common.json#/definitions/id"
+    },
+    "args": {
```
LucasBrazi06, line 11: For the charging station's action it's more tricky.
We have one endpoint and we mixed everything unfortunately!
I would like that we match the OCPP standard according the Action and that we identify the mandatory params in the 'args' object.
We can have a meeting for this and let it as it is for the time being.

src/server/rest/v1/service/ChargingStationService.ts
```diff
@@ -695,7 +695,7 @@ export default class ChargingStationService {
 
   public static async handleDeleteChargingStation(action: ServerAction, req: Request, res: Response, next: NextFunction): Promise<void> {
     // Filter
-    const chargingStationID = ChargingStationSecurity.filterChargingStationRequestByID(req.query);
+    const chargingStationID = ChargingStationValidator.getInstance().validateChargingStationDeleteReq(req.params).id;
```
LucasBrazi06, line 698: In the router we have in the delete:
`req.query.ID = req.params.id;`
Why do we have a schema here with the 'id' in lowercase?

src/server/rest/v1/service/ChargingStationService.ts
```diff
@@ -818,16 +818,16 @@ export default class ChargingStationService {
...
+    UtilsService.assertIdIsProvided(action, filteredRequest.id, MODULE_NAME, 'handleGetChargingStation', req.user);
     // Check auth
     if (!Authorizations.canReadChargingStation(req.user)) {
       throw new AppAuthError({
         errorCode: HTTPAuthError.FORBIDDEN,
         user: req.user,
         action: Action.READ, entity: Entity.CHARGING_STATION,
         module: MODULE_NAME, method: 'handleGetChargingStation',
-        value: filteredRequest.ID
+        value: filteredRequest.id
```
LucasBrazi06, line 830: ID should be in uppercase in GET Http verb

src/types/requests/HttpByIDRequest.ts
```diff
@@ -3,3 +3,7 @@ import { HttpDatabaseProjectRequest } from './HttpDatabaseRequest';
 export default interface HttpByIDRequest extends HttpDatabaseProjectRequest {
   ID: string;
 }
+
+export interface HttpRestByIDRequest extends HttpDatabaseProjectRequest {
+  id: string;
```
LucasBrazi06, line 8: Should use HttpByIDRequest only I think.

LucasBrazi06 (changes requested):

src/server/rest/v1/service/ChargingStationService.ts
Ninjeneer: As single ids won't be passed as query strings anymore, I was planning to use lowercase everywhere. What's your point of view ?

src/types/requests/HttpByIDRequest.ts
Ninjeneer: According to my comment above, I'm using this interface as a temporary interface that will replace the old uppercased HttpByIDRequest when every endpoints will be refactored

src/server/rest/v1/service/ChargingStationService.ts
LucasBrazi06: Before you plan anything, you have to discuss it with me first ;-).
As I wrote, in the router, you are assigning the param to the query in uppercase (req.query.ID = req.params.id) so why it ends up with an 'id' after applying the Json schema?
This is not what has been done in the Tenant schema in similar situation.
Will we have a regression in Angular or in the mobile app if you change this?

Ninjeneer: The `id` parameter passed in path is lowercased. I used `req.query.ID = req.params.id` because security classes require an uppercased `ID` parameter coming from query object (as it was passed as a query parameter before). I made this translation to avoid big refactoring on classes destined to be deleted soon.
Now that we are slowly removing security classes, we don't need to do this "translation" anymore and can use params as they are.
This won't have any effect on Angular/Mobile app, as the HTTP call stay the same.

Ninjeneer: > Before you plan anything, you have to discuss it with me first ;-).

Ok, my bad

Yes you're right, we cannot change parameters as old endpoints with old parameters are still used.
I will set back the req.query.ID = req.params.id;

To explain why I did this (too early) :
The id parameter passed in path is lowercased. I used req.query.ID = req.params.id because security classes require an uppercased ID parameter coming from query object (as it was passed as a query parameter before). I made this translation to avoid big refactoring on classes destined to be deleted soon.
Now that we are slowly removing security classes, we don't need to do this "translation" anymore and can use params as they are.

Ninjeneer: Done, rolled back IDs in query.

src/server/rest/v1/service/ChargingStationService.ts
Ninjeneer: Done

src/types/requests/HttpByIDRequest.ts
Ninjeneer: Done, rolled back IDs in query.

src/assets/server/rest/v1/schemas/chargingstation/chargingstation-action.json
Ninjeneer: Sure, as you have a much busier timetable than me, can I let you plan it ?

LucasBrazi06 (approved):
GUT!

PR #2219, opened 2021-03-23
Fix tenant creation error

Ninjeneer:


src/server/rest/v1/service/TenantService.ts
```diff
@@ -365,6 +380,28 @@ export default class TenantService {
     tenant.lastChangedOn = new Date();
     // Update Tenant
     await TenantStorage.saveTenant(tenant, Utils.objectHasProperty(filteredRequest, 'logo') ? true : false);
+    // OICP
+    if (tenant.components && tenant.components.oicp) {
```
LucasBrazi06, line 384: Should also put this code in the handleCreateTenant() method as you can activate the OICP also at creation time?
Then to put this code in one method called from Udapte/Create?

Ninjeneer: > Should also put this code in the handleCreateTenant() method as you can activate the OICP also at creation time?

It already is :)

>  Then to put this code in one method called from Update/Create?

Done

LucasBrazi06 (approved):
GUT!
Thanks a lot!

PR #2225, opened 2021-03-25
Crypto migration task

roxiciobotea210:
Add task for changing encryption key in case it matches the historical one.
* get both keys & compare them
* if they match, generate a new key with the default algorithm
* update key settings
* migrate encrypted data to new key


jerome-benoit (changes requested):
It should be a migration task, not a scheduler task. That's probably because I've not tell you where exactly to add it ;)
That's not difficult to make it a migration task. 
Otherwise looks good to me.

roxiciobotea210:
> It should be a migration task, not a scheduler task. That's probably because I've not tell you where exactly add it ;)
> That's not difficult to make it a migration task.
> Otherwise looks good to me.

Ah, sorry, I created it following the example of the scheduler task Stefania created and that's why I added it here. Will change it to migration, thanks! :)

src/migration/tasks/ChangeCryptoKeyTask.ts
```diff
@@ -0,0 +1,63 @@
...
+        crypto: {
+          formerKey: currentCryptoSettings.crypto.key,
+          formerKeyProperties: currentCryptoSettings.crypto.keyProperties,
+          key: Utils.generateRandomKey(Utils.getDefaultKeyProperties()),
+          keyProperties: Utils.getDefaultKeyProperties(),
+          migrationToBeDone: true
+        }
+      } as CryptoSettings;
+      await SettingStorage.saveCryptoSettings(tenant.id, keySettingToSave);
+      // migrate sensitive data to the new key
```
jerome-benoit, line 43: Nitpick: comment should start with an upper case letter.

jerome-benoit (approved):

LucasBrazi06 (approved):
GUT!

PR #2232, opened 2021-03-26
REST > Charging Profile create endpoint

Ninjeneer:


LucasBrazi06 (changes requested):
Only add required fields in chargingProfile

PR #2234, opened 2021-03-26
billing - repairUser  - to recover from inconsistencies

ClaudeROSSI:
forceSynchronizeUser was not repairing inconsitencies anymore when customerID was set with a wrong value.

This is fixed


ClaudeROSSI:
@LucasBrazi06 : I introduced a regression on forceSynchronizeUser().
I have changed a bit the code to isolate the logic which is responsible for repairing inconsitencies in a dedicated call:

repairUser(user)

+ added a test to make sure this works as expected 
+ some minor test changes

PR #2239, opened 2021-03-28
Async task framework

LucasBrazi06:


src/async-task/AsyncTask.ts
```diff
@@ -0,0 +1,37 @@
+import AsyncTask from '../types/AsyncTask';
+import Constants from '../utils/Constants';
+import Logging from '../utils/Logging';
+import { ServerAction } from '../types/Server';
+import moment from 'moment';
+
+const MODULE_NAME = 'AbstractAsyncTask';
+
+export default abstract class AbstractAsyncTask {
```
jerome-benoit, line 9: Short resume of the email exchange: 

If that class extends `AsyncResource`, the API to run task can then handle sync or async tasks:

+ Method code sketch to execute a task: 

```ts

public execute() {
    if (this.async) {
        this.runInAsyncScope(this.internalExecute.bind(this), this)
    } else {
        this.runInAsyncScope(this.internalExecute.bind(this), this) 
    }
}

protected abstract internalExecute(): (data?) => void | (data?) => Promise<void>;
```

PR #2250, opened 2021-03-30
Billing - stripe - API cleanup

ClaudeROSSI:
- BillingPaymentMethodResult has been removed
- HttpBillingRequest has been cleanup - Trying to apply the same logic for all endpoints (still not clear for the deletePaymentMethod)
- removed the attempt to sanitize the request.user ==> userToken 
- getPaymentMethods - returns Promise<BillingPaymentMethod[]>
- removed console.log of operation results
- removed some TODO comments
- added call to Logging.logSecurityInfo() when deleting payment method


cabch (approved):
ok for me, much clearer and everywhere the same pattern + tested and nothing's broken

PR #2253, opened 2021-03-30
Use lock with timeout for smart charging.

jerome-benoit:
Signed-off-by: Jérôme Benoit <jerome.benoit@piment-noir.org>


OliveGerste:
@LucasBrazi06 
Okay from my side. 
Unfortunately I did not find a real way to test in the local environment with smart charging.

PR #2278, opened 2021-04-07
Add projected fields of companies in the auth definition file

LucasBrazi06:


src/types/UserToken.ts
```diff
@@ -3,6 +3,7 @@ import User from './User';
 export default interface UserToken {
   id?: string;
   role?: string;
+  rolesACL?: string[];
```
jerome-benoit, line 6: It does not seem to be used.

src/server/rest/v1/service/SiteService.ts
```diff
@@ -600,7 +601,7 @@ export default class SiteService {
     // Check data is valid
     UtilsService.checkIfSiteValid(filteredRequest, req);
     // Check static auth for reading company
-    if (!Authorizations.canReadCompany(req.user)) {
+    if (!(await Authorizations.canReadCompany(req.user)).authorized) {
```
jerome-benoit, line 604: Is `.authorized` really needed here?

src/types/UserToken.ts
LucasBrazi06: Yes it is used in check of auths right away without rebuilding it on the fly whenever there is an HTTP request:

image

src/server/rest/v1/service/SiteService.ts
LucasBrazi06: This code is no longer here after merge with master-qa :-)

LucasBrazi06:
@roxiciobotea210 I merged the pull request to keep the next changes readable and I think the current implementation is stable enough.

PR #2295, opened 2021-04-08
Billing - router RESTful endpoints

ClaudeROSSI:
BillingRouter + ServerRoute have been updated 

- ONLY for endpoints dealing with the payment methods for now!


src/server/rest/v1/router/GlobalRouter.ts
```diff
@@ -24,7 +24,6 @@ export default class GlobalRouter {
     this.buildRouteAPI();
     this.buildRouteUtils();
     this.buildRouteDocs();
-    this.buildRouteBilling();
```
jerome-benoit, line 27: You should be whipped for having defined you own routes for billing :)

PR #2297, opened 2021-04-09
Specified tenant creation errors

Ninjeneer:


test/api/TenantTest.ts
```diff
@@ -268,7 +268,7 @@ describe('Tenant tests', function() {
       response = await CentralServerService.defaultInstance.createEntity(
         CentralServerService.defaultInstance.tenantApi, tenant, false);
       // Check
-      expect(response.status).to.equal(597);
+      expect(response.status).to.equal(598);
```
LucasBrazi06, line 271: Use the constant.

src/server/rest/v1/service/TenantService.ts
```diff
@@ -236,7 +236,7 @@ export default class TenantService {
     if (foundTenant) {
```
LucasBrazi06, line 236: I just saw that and to be honest we can remove this check.
There is no unicity on the tenant's name.

LucasBrazi06 (changes requested):

test/api/TenantTest.ts
Ninjeneer: Done

src/server/rest/v1/service/TenantService.ts
Ninjeneer: Ok, removed

LucasBrazi06 (approved):

PR #2307, opened 2021-04-12
Type validation code.

jerome-benoit:
And make it more consistent: 
+ Fix wrong async usage;
+ Move types definition file to the right place;
+ Coding style consistency.

@Ninjeneer could you please be more careful to any type usage, linter error on promises and coding rules in e-mobility?  

The async misusage can lead to performance issue.

Signed-off-by: Jérôme Benoit <jerome.benoit@piment-noir.org>


LucasBrazi06 (approved):

src/server/rest/v1/validator/AuthValidation.ts
```diff
@@ -63,7 +63,7 @@ export default class AuthValidator extends SchemaValidator {
     return data;
   }
 
-  public validateAuthResendVerificationEmail(data: any): Partial<HttpResendVerificationMailRequest> {
+  public validateAuthResendVerificationEmail(data: HttpResendVerificationMailRequest): Partial<HttpResendVerificationMailRequest> {
```
Ninjeneer, line 66: Please correct me if I'm wrong, but I think we can't type `data` here. As it is a user input, we can't determine which type it is before processing the validation. Same for typing below. What do you think ?

LucasBrazi06: > Please correct me if I'm wrong, but I think we can't type `data` here. As it is a user input, we can't determine which type it is before processing the validation. Same for typing below. What do you think ?

Normally you should expect to receive the type HttpResendVerificationMailRequest, nothing else so you can type it.
It's better than 'any'.
What's the problem of typing it here?

Ninjeneer: Ok, I thought that as anything could be sent through the endpoint, we were not supposed to anticipate the structure of the payload (as the any "request" param in each security class) until the validation is processed.

jerome-benoit: We basically validate an already structured request in that code but when the input is a `QueryString.ParsedQs` express type, it's difficult to make the input type match the output type. I think some `Http...Request` type definition need some refinements to allow them to be reused transparently with express `Request` attributes.

PR #2309, opened 2021-04-12
fix charge point validation method

OliveGerste:


LucasBrazi06:
@OliveGerste Is it related to the auto compute of Charging Station's Max Power when manually configured?

LucasBrazi06:
> @OliveGerste Is it related to the auto compute of Charging Station's Max Power when manually configured?

Got my answer in another PR :-)

PR #2320, opened 2021-04-13
Final Auth concept without roles for Company

LucasBrazi06:


roxiciobotea210 (approved):
Code is definitely more complex now, but it looks much cleaner! 
All seems to be working well & I understand the general concept. I will definitely understand it better when implementing Sites & the rest of entities. :)

LucasBrazi06:
> Code is definitely more complex now, but it looks much cleaner!
> All seems to be working well & I understand the general concept. I will definitely understand it better when implementing Sites & the rest of entities. :)

Okay, thanks for your feedback!

roxiciobotea210:
@LucasBrazi06 So should I start on Sites? :)

LucasBrazi06:
> @LucasBrazi06 So should I start on Sites? :)

Yes :-)
Let me know if you need any help!

PR #2322, opened 2021-04-14
REST > getChargingStationTransactions / firmwareDownload / SmartChargingTrigger / InError validation

Ninjeneer:
4 schemas instead of 3, but one of them is extra small.


LucasBrazi06:
@OliveGerste Could you please check this PR :-)
Thanks in advance!

src/assets/server/rest/v1/docs/e-mobility-oas.json
```diff
@@ -5386,6 +5375,14 @@
           "type": "boolean"
         }
       },
+      "RefundStatus": {
+        "in": "query",
+        "name": "RefundStatus",
+        "description": "Pipe separated refund statuses (submitted, notSubmitted, cancelled, approved)",
+        "schema": {
+          "type": "string"
```
jerome-benoit, line 5383: I think you can enforce the string list.

Ninjeneer: I could add pattern check here, but is it relevant as it is only documentation ?

jerome-benoit: The code strictly enforce the list to that enum: 

```ts
export enum RefundStatus {
  SUBMITTED = 'submitted',
  NOT_SUBMITTED = 'notSubmitted',
  CANCELLED = 'cancelled',
  APPROVED = 'approved',
}
```

So must do the inputs validator.

Ninjeneer: You're right, haven't noticed validators patterns were missing for multiple values. 
Done.

src/assets/server/rest/v1/schemas/transaction/transactions-get.json
```diff
@@ -0,0 +1,87 @@
+{
+  "title": "Get Transactions Request",
+  "type": "object",
+  "properties": {
+    "ChargeBoxID": {
+      "$ref": "chargingstation.json#/definitions/id"
```
OliveGerste, line 6: Do you think we need a differentiation between id and ids here? Like it is done in common.json.
My thought is, that we have some requests where we only want to reach one station and some where we have multiple stations. 
So far I did not experience any error because of this, so not urgent I guess.

src/assets/server/rest/v1/schemas/transaction/transactions-get.json
```diff
@@ -0,0 +1,87 @@
...
+    },
+    "Issuer": {
+      "type": "boolean",
+      "sanitize": "mongo"
+    },
+    "ConnectorID": {
+      "$ref": "chargingstation.json#/definitions/connector/properties/connectorId"
+    },
```
OliveGerste, line 14: Like this it is only possible to filter for one connector. If you filter for multiple connectors you will get an error in the front end.

src/assets/server/rest/v1/schemas/transaction/transactions-get.json
```diff
@@ -0,0 +1,87 @@
...
+      "$ref": "common.json#/definitions/ids"
+    },
+    "SiteID": {
+      "$ref": "common.json#/definitions/ids"
+    },
+    "UserID": {
+      "$ref": "common.json#/definitions/ids"
+    },
+    "TagID": {
+      "$ref": "tag.json#/definitions/id"
```
OliveGerste, line 25: Same thought like for the chargebox ids. Do we need a differentiation between singular and plural?
It works currently because the string can be split before the database request even if the definition is singular.

src/storage/mongodb/ChargingStationStorage.ts
```diff
@@ -256,7 +256,7 @@ export default class ChargingStationStorage {
     // Connector Status
     if (params.connectorStatuses) {
       filters['connectors.status'] = { $in: params.connectorStatuses };
-      filters.inactive = false;
+      filters.inactive = true;
```
OliveGerste, line 259: Why is the inactive flag set to true, when we filter for connectorStatuses?
We will not get any results, if we search for e.g. status 'Available' right?

OliveGerste (changes requested):
Looks good so far.
I added some small comments with my thoughts. I also noticed that sometimes the responses in the swagger do not completely match the real responses e.g. you so not get capabilities from GET '​/api​/charging-stations'.

src/assets/server/rest/v1/schemas/chargingstation/chargingstations-get.json
```diff
@@ -37,10 +40,6 @@
       "type": "boolean",
       "sanitize": "mongo"
     },
-    "ErrorType": {
-      "type": "string",
-      "sanitize": "mongo"
-    },
     "LocLongitude": {
       "type": "number",
       "sanitize": "mongo"
```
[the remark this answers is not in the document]
jerome-benoit: For latitude and longitude, the backend is also doing the check but I guess we can but it can be done inside the validator.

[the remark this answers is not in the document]
OliveGerste: It was checked in the `ChargingStationSecurity `, Method:` Utils.containsGPSCoordinates` right?
But this is skipped with the new schemes I think.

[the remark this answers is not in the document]
Ninjeneer: Ok, as I can't add it directly to the schema, I added a check in the validator method.
Also improved the regex for GPS checks as the previous allowed non-sense values

src/assets/server/rest/v1/schemas/transaction/transactions-get.json
Ninjeneer: Yes you're right. 
Added a specific parameters "ids" and enforced the "id" parameter

src/assets/server/rest/v1/schemas/transaction/transactions-get.json
Ninjeneer: Fixed

src/assets/server/rest/v1/schemas/transaction/transactions-get.json
Ninjeneer: Added a specific parameters "ids" and enforced the "id" parameter

src/storage/mongodb/ChargingStationStorage.ts
Ninjeneer: I just pushed a temporary test, my bad

src/assets/server/rest/v1/schemas/chargingstation/chargingstations-get.json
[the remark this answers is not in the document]
jerome-benoit: > Ok, as I can't add it directly to the schema, I added a check in the validator method.
> Also improved the regex for GPS checks as the previous allowed non-sense values

You can reuse the regexp in `Contants.ts` to validate the latitude and longitude inputs. 
Maybe you can also directly the validator package in ajv that offer a bunch of input validation helpers.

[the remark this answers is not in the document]
Ninjeneer: >  Maybe you can also directly the validator package in ajv that offer a bunch of input validation helpers.

Great idea, added a `coordinates` custom ajv format according to regexp in Constants.ts

OliveGerste (approved):

PR #2325, opened 2021-04-14
Ensure amperage limit value is at least the min per phase * number of

jerome-benoit:
phases * number of connectors


src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -2181,6 +2177,7 @@ export default class OCPPUtils {
     } else if (Utils.objectHasProperty(connector, 'amperageLimit') && connector.amperageLimit < connectorAmperageLimitMin) {
       return connectorAmperageLimitMin;
     }
+    return connectorAmperageLimitMin;
```
LucasBrazi06, line 2180: The method's name is not aligned with the code:
checkAndGetConnectorAmperageLimit -> One connector

An here you multiply by the number of connectors:
`const connectorAmperageLimitMin = StaticLimitAmps.MIN_LIMIT_PER_PHASE * numberOfPhases * numberOfConnectors;`

jerome-benoit: Fixed in https://github.com/sap-labs-france/ev-server/pull/2330

PR #2340, opened 2021-04-15
Always delete user physically + usersCars + tags

mahdibenromdhane:


src/server/rest/v1/service/TagService.ts
```diff
@@ -248,6 +249,17 @@ export default class TagService {
         action: action
       });
     }
+    const nbrTransactions = await TransactionStorage.getTransactions(req.user.tenantID,{ tagIDs:[filteredRequest.id.toUpperCase()] }, Constants.DB_PARAMS_MAX_LIMIT);
```
LucasBrazi06: To get the number of Transaction, always use the Constants.DB_PARAMS_COUNT_ONLY otherwise you may have -1 as a result + you are retrieving here all the Transactions just to check the count (rule 22).

src/storage/mongodb/CarStorage.ts
```diff
@@ -15,8 +15,8 @@ const MODULE_NAME = 'CarStorage';
 
 export default class CarStorage {
   public static async getCarCatalog(id: number = Constants.UNKNOWN_NUMBER_ID,
-      params: { withImage?: boolean; } = {},
-      projectFields?: string[]): Promise<CarCatalog> {
+    params: { withImage?: boolean; } = {},
```
LucasBrazi06: Wrong indentation, check your VSCode plugin

src/storage/mongodb/UserStorage.ts
```diff
@@ -925,15 +920,18 @@ export default class UserStorage {
...
+    await global.database.getCollection<any>(tenantID, 'userimages')
+      .findOneAndDelete({ '_id': Utils.convertToObjectID(id) });
     // Delete Site Users
     await global.database.getCollection<any>(tenantID, 'siteusers')
       .deleteMany({ 'userID': Utils.convertToObjectID(id) });
-    // Delete Image
-    await global.database.getCollection<any>(tenantID, 'userimages')
-      .findOneAndDelete({ '_id': Utils.convertToObjectID(id) });
     // Delete Tags
     await global.database.getCollection<any>(tenantID, 'tags')
```
LucasBrazi06, line 930: I think that Tags should be disabled, not deleted.

src/types/ocpi/OCPIAuthorizationInfo.ts
```diff
@@ -11,7 +11,6 @@ export interface OCPIAuthorizationInfo {
 export enum OCPIAllowed {
   ALLOWED = 'ALLOWED',
   BLOCKED = 'BLOCKED',
-  EXPIRED = 'EXPIRED',
```
LucasBrazi06, line 14: This is OCPI enum, still can be used if necessary.

LucasBrazi06 (changes requested):
Some code formatting are work, check your VSCode plugin with Jerome.
Get the MongoDB collection and call the find() in one line.
I proceeded with some changes and I'll merge this PR but one last thing needs to be done:
When user is deleted, the Tag should be either:
- Set inactive and userID kept -> Cannot assign the badge anymore
- Change the active flag to status flag and provide an enum: ACTIVE, INACTIVE, ARCHIVED
Let me know.

PR #2347, opened 2021-04-16
Billing - settings - explicit activation

ClaudeROSSI:
The billing settings have been changed to allow entering the stripe connection settings without activating the billing feature.

**isTransactionBillingActivated**
- This new property can be used to switch ON/OFF the billing of transactions.
- This property should only be set to true when the settings are consistent (this will be done later)

This information is used on start transaction and stored in the transaction billing data as the **withBillingActive** flag.
This flag is propagated until the stop transaction.

**Transaction Billing Data Example:**

image

### DIFF
--- a/src/integration/billing/stripe/StripeBillingIntegration.ts
+++ b/src/integration/billing/stripe/StripeBillingIntegration.ts
@@ -637,7 +637,8 @@ export default class StripeBillingIntegration extends BillingIntegration {
 
     if (!this.settings.billing.isTransactionBillingActivated) {
       return {
-        isTransactionBillingActivated: false
+        // Keeps track whether the billing was activated or not on start transaction
+        withBillingActive: false
       };
     }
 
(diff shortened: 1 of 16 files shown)


src/integration/billing/stripe/StripeBillingIntegration.ts
```diff
@@ -634,6 +634,14 @@ export default class StripeBillingIntegration extends BillingIntegration {
   }
 
   public async startTransaction(transaction: Transaction): Promise<BillingDataTransactionStart> {
+
```
LucasBrazi06, line 637: Blank line

src/integration/billing/stripe/StripeBillingIntegration.ts
```diff
@@ -634,6 +634,14 @@ export default class StripeBillingIntegration extends BillingIntegration {
...
+    if (!this.settings.billing.isTransactionBillingActivated) {
+      return {
+        // Keeps track whether the billing was activated or not on start transaction
+        withBillingActive: false
+      };
+    }
+
```
LucasBrazi06, line 644: Blank line

LucasBrazi06 (approved):

PR #2349, opened 2021-04-16
Statistic ut fixes

cabch:
+ changed .replace to delete only double quotes that are at the beginning and the end of the value we want to put in csv and we keep the one inside the text if any


src/server/rest/v1/service/StatisticService.ts
```diff
@@ -619,27 +628,20 @@ export default class StatisticService {
         });
       }
       // Now build the export file
-      let number: number;
+      let numberOfTransaction: number;
```
jerome-benoit, line 631: With an s ;)

PR #2359, opened 2021-04-19
Site Authorization Concept Enhancement (use dynamic filters)

roxiciobotea210:
I'm not sure if anything is missing, the assign users scenario was very confusing to implement.


LucasBrazi06 (commented):
I haven't gone through every line of code but it looks very good!
I'll test it along the front-end that needs little changes.
Let me know if you need a small task or may we can improve the unit tests?

roxiciobotea210:
> I haven't gone through every line of code but it looks very good!
> I'll test it along the front-end that needs little changes.
> Let me know if you need a small task or may we can improve the unit tests?

Thanks a lot for checking @LucasBrazi06 ! Let me know if you find anything missing.
I was thinking to do the same changes for SiteAreas since this is just for the Sites. Is that ok?

LucasBrazi06:
> I was thinking to do the same changes for SiteAreas since this is just for the Sites. Is that ok?

Yes go ahead.

roxiciobotea210:
Please do not merge this yet, I found some things I can improve while working on SiteAreas, will work on it tomorrow. :)

LucasBrazi06:
@roxiciobotea210 I tested the app and found out that the admin user cannot delete anymore any users.
Could you check?
Thanks!

roxiciobotea210:
> @roxiciobotea210 I tested the app and found out that the admin user cannot delete anymore any users.
> Could you check?
> Thanks!

Issue should be fixed now.

PR #2363, opened 2021-04-20
Allow delete badge with linked transactions

mahdibenromdhane:


src/server/rest/v1/service/TagService.ts
```diff
@@ -157,17 +157,6 @@ export default class TagService {
...
-        source: Constants.CENTRAL_SERVER,
-        errorCode: HTTPError.TAG_HAS_TRANSACTIONS,
-        message: `Cannot delete Tag ID '${tag.id}' which has '${tag.transactionsCount}' transaction(s)`,
-        module: MODULE_NAME, method: 'handleDeleteTag',
-        user: req.user,
-        action: action
-      });
-    }
     // Delete the Tag
     await TagStorage.deleteTag(req.user.tenantID, tag.id);
```
LucasBrazi06, line 161: By the way, when you delete a Tag in TagStorage, we should also delete usertags right?
And I would expect the delete to be at the end of the method.

mahdibenromdhane: - I didn't understand what do you mean by usertags.
- I think we don't the delete tag at the end of the method because we fetch the firstActiveTag to set it as default Tag, if we do the delete later, we might get tag to be deleted as firstActiveTag.

LucasBrazi06: > * I didn't understand what do you mean by usertags.

You're right lol, I'm losing my mind, there is no intermediate collection between tags and users 😩 !!!
My bad!

> * I think we don't delete the tag at the end of the method because we fetch the firstActiveTag to set it as default Tag, if we do the delete later, we might get tag to be deleted as firstActiveTag.

Yes this makes sense but I think at least OCPI could be moved before the delete.

mahdibenromdhane: > > * I didn't understand what do you mean by usertags.
> 
> You're right lol, I'm losing my mind, there is no intermediate collection between tags and users 😩 !!!
> My bad!
> 
> > * I think we don't delete the tag at the end of the method because we fetch the firstActiveTag to set it as default Tag, if we do the delete later, we might get tag to be deleted as firstActiveTag.
> 
> Yes this makes sense but I think at least OCPI could be moved before the delete.

Lool , It happens ! 

Done for the OCPI

LucasBrazi06 (commented):
I changed the code to use the same delete method named deleteTags() from handleDeleteTags() and handleDeleteTag() and get the same behavior as for the delete many.
Let me know if you did that on purpose or if the usage of the same method is correct.

PR #2369, opened 2021-04-22
Eichrecht : Signed Data is not correctly transmited with some systems

WillyLellouch:
Update to the fix + new Unit Tests
The previous fix didn't take into account the previous data in the transaction.
This is now fixed


LucasBrazi06:
@WillyLellouch You sent an email for QA deployment but the task is under Dev cross review.
Should be in To be reviewed if you want me to deploy it ;-)

jerome-benoit (approved):
Good to go

PR #2370, opened 2021-04-22
Initial work to ensure charging stations with rejected boot notification are seen

jerome-benoit:
offline.


src/server/ocpp/services/OCPPService.ts
```diff
@@ -64,6 +64,8 @@ export default class OCPPService {
         heartbeatIntervalSecs = this.chargingStationConfig.heartbeatIntervalOCPPJSecs;
         break;
     }
+    // Get the charging station
+    let chargingStation = await ChargingStationStorage.getChargingStation(headers.tenantID, headers.chargeBoxIdentity);
```
LucasBrazi06, line 68: OCPP validation (Json Schema) should be checked first

src/server/ocpp/services/OCPPService.ts
```diff
@@ -277,7 +279,12 @@ export default class OCPPService {
         error.params.source = headers.chargeBoxIdentity;
       }
       await Logging.logActionExceptionMessage(headers.tenantID, ServerAction.BOOT_NOTIFICATION, error);
-      // FIXME: charging stations in DB with rejected boot notification are still seen as online and usable. OCPP spec is clear on how to handle such case: close websocket, no messages of any kind sent or received, ...
+      if (chargingStation) {
+        // Set boot notification registration status to 'Rejected'
+        chargingStation.registrationStatus = RegistrationStatus.REJECTED;
+        // Save Charging Station
+        await ChargingStationStorage.saveChargingStation(headers.tenantID, chargingStation);
```
LucasBrazi06, line 286: I don't think you can save a Charging Station with the same ID that was rejected because the s/n was different.
You will make a valid charging station Rejected.
So the ID must differs to save it and you will have no trace of it.
If this is the only goal of this PR then it can close it ;-).

LucasBrazi06 (changes requested):

src/server/ocpp/services/OCPPService.ts
jerome-benoit: Saving just the status instead of the whole data structure will be enough.

LucasBrazi06: This does not change the initial problem: 2 charging stations with the same ID with different S/N -> The wrong one must be rejected and you should not override the registration status of the right one.

jerome-benoit: Read the code in the PR.

src/server/ocpp/services/OCPPService.ts
```diff
@@ -150,7 +151,23 @@ export default class OCPPService {
           (chargingStation.chargePointSerialNumber && bootNotification.chargePointSerialNumber &&
             chargingStation.chargePointSerialNumber !== bootNotification.chargePointSerialNumber)) {
           // Not the same Charging Station!
-          // FIXME: valid charging stations in DB with modified parameters cannot be registered again without direct access to the DB
+          // FIXME: valid charging stations in DB with modified parameters cannot be registered at boot notification again without direct
+          //        access to the DB, standard charging station replacement is then not possible. The registration status returned should
+          //        be 'Pending' and a way to manually accept or refuse such stations should be offered.
+          // Do not flag charging stations online with registered boot notification as 'Rejected'
+          if (moment().subtract(Configuration.getChargingStationConfig().maxLastSeenIntervalSecs, 'seconds').isSameOrBefore(chargingStation.lastSeen) && chargingStation.registrationStatus === RegistrationStatus.ACCEPTED) {
```
LucasBrazi06, line 158: What was the line max length by the way? ;-)

src/server/ocpp/services/OCPPService.ts
```diff
@@ -150,7 +151,23 @@ export default class OCPPService {
...
+              tenantID: headers.tenantID,
+              source: chargingStation.id,
+              action: ServerAction.BOOT_NOTIFICATION,
+              module: MODULE_NAME, method: 'handleBootNotification',
+              message: 'Trying to connect a charging station matching an online charging station with identical chargeBoxID, registered boot notification and different attributes',
+              detailedMessages: { headers, bootNotification }
+            });
+          // Set charging station boot notification registration status to 'Rejected'
+          } else {
+            await ChargingStationStorage.saveChargingRegistrationStatus(headers.tenantID, chargingStation.id, { registrationStatus: RegistrationStatus.REJECTED });
```
LucasBrazi06, line 169: I would not also take care of this scenario.
If there is a network outage or some works like in Caen and I connect a charging station with the same ID you will save the status and we should not.
The lastSeen should not be used to assume that this is a charger's replacement.
And the waring message is quite the same as the exception below.
I will not save anything again in this method!

LucasBrazi06 (changes requested):

src/server/ocpp/services/OCPPService.ts
jerome-benoit: The goal is to save the boot notification status in that case to ensure charging station with a rejected status are not seen online. 
The way it's now done after your fruitful code review in that PR handle all use case when charging stations have the same ID properly: 
+ recorded charging station boot notification status is never overridden if the charging station is online and tenant administrator have a warning log on the same ID usage;
+ recorded charging station boot notification status only is overridden if the charging station is offline to ensure station with the same ID but different attributes will not be seen as online ans usable. If the recorded charging station comes back online, the boot registration status is then saved as 'Accepted' and the code behaviour goes back to the previous point. 

If necessary, I can explain next week in the real world the user experience 'bug' that PR is meant to fix. 
	 
And the one to one remplacement scenario handling asked by all of your charging station customers is not handled in that PR.

jerome-benoit:
I put it back in the review pipeline to ensure that my last comment have been read.

LucasBrazi06:
> I put it back in the review pipeline to ensure that my last comment have been read.

Sorry, I must have missed the notif.
I'll check once I got my head out of under the water!

src/server/ocpp/services/OCPPService.ts
LucasBrazi06: > * recorded charging station boot notification status is only overridden if the charging station is offline to ensure station with the same ID but different attributes will not be seen as online and usable. If the recorded charging station comes back online, the boot registration status in the DB record is then saved as 'Accepted' and the code behaviour goes back to the previous point.

Wrong here: as a general assumption, you cannot change a property which does not belong to the entity.
S/N or model is different -> there is nothing to update here.
Moreover, if there is a network cut, the charging station will reconnect without sending a BootNotification.

> And the one to one remplacement scenario handling asked by all of your charging station integrator customers is not handled in that PR.

This can be done by renaming the old charging station to free the ID with all its history and we didn't do it yet because we simply lacked time. That doesn't mean we will not do it. We have our prios.

jerome-benoit: > > * recorded charging station boot notification status is only overridden if the charging station is offline to ensure station with the same ID but different attributes will not be seen as online and usable. If the recorded charging station comes back online, the boot registration status in the DB record is then saved as 'Accepted' and the code behaviour goes back to the previous point.
> 
> Wrong here: as a general assumption, you cannot change a property which does not belong to the entity.
> S/N or model is different -> there is nothing to update here.
> Moreover, if there is a network cut, the charging station will reconnect without sending a BootNotification.

The issue rely on the data model design for the charging stations collection, the record unicity is only enforced on the chargeboxid. The unicity must be on the tuple (id, vendor, model, serial) to match the intended behaviour and allow record creation in DB and boot notification status proper tracking of charging stations with the same id. 

I've removed for now the saving of the boot registration status in case of failure in that PR, let the boot notification status recording even if not yet useful and will create an issue and implement the clean solution in subsequents PRs that will make use of the boot notification status properly.

LucasBrazi06 (approved):

PR #2378, opened 2021-04-23
Iothink integration response filtering according new model

OliveGerste:
Refactoring of the response filtering. 
The building is not usable for smart charging currently, because it represents the whole consumption of the site area (Fast chargers, Building, solar panels, Battery)
We have to clarify with Iothink how to handle this. 
Maybe a property renaming of the consuming asset response will be necessary again after this.


src/integration/asset/iothink/IothinkAssetIntegration.ts
```diff
@@ -92,11 +135,9 @@ export default class IothinkAssetIntegration extends AssetIntegration<AssetSetti
     return consumptions;
   }
 
-  private getPropertyValue(data: any[], propertyName: string, index: number): number {
-    for (const measure of data) {
-      if (measure.tagReference === propertyName) {
-        return Utils.convertToFloat(Utils.createDecimal(measure.logs[index]?.value ? measure.logs[index].value : 0)) * 1000;
-      }
+  private getPropertyValue(data: any, propertyName: string): number {
+    if (typeof data[propertyName] !== 'undefined') {
```
jerome-benoit, line 139: `!Utils.isUndefined()`?

LucasBrazi06:
@OliveGerste Does Gerald is aware about that? Any action from ioThink's side?
Thanks.

src/integration/asset/iothink/IothinkAssetIntegration.ts
```diff
@@ -65,24 +65,67 @@ export default class IothinkAssetIntegration extends AssetIntegration<AssetSetti
 
   private filterConsumptionRequest(asset: Asset, data: any, manualCall: boolean): AbstractCurrentConsumption[] {
     const consumptions: AbstractCurrentConsumption[] = [];
+    // Rename the value property in logs with tag reference
+    for (const dataSet of data.historics) {
+      dataSet.logs.forEach((log) => {
```
LucasBrazi06, line 70: Why don't you use for/of on this one?

src/integration/asset/iothink/IothinkAssetIntegration.ts
```diff
@@ -65,24 +65,67 @@ export default class IothinkAssetIntegration extends AssetIntegration<AssetSetti
...
+        log[dataSet.tagReference] = log['value'];
+        delete log['value'];
+      });
+    }
+    // Create helper map to merge the arrays to combined objects
+    const map = new Map();
+    for (const dataSet of data.historics) {
+      dataSet.logs.forEach((item) => map.set(item.timestamp, { ...map.get(item.timestamp), ...item }));
```
LucasBrazi06, line 78: Use for/of (more readable).

src/integration/asset/iothink/IothinkAssetIntegration.ts
```diff
@@ -65,24 +65,67 @@ export default class IothinkAssetIntegration extends AssetIntegration<AssetSetti
...
     const consumptions: AbstractCurrentConsumption[] = [];
+    // Rename the value property in logs with tag reference
+    for (const dataSet of data.historics) {
+      dataSet.logs.forEach((log) => {
+        log[dataSet.tagReference] = log['value'];
+        delete log['value'];
+      });
+    }
+    // Create helper map to merge the arrays to combined objects
+    const map = new Map();
```
LucasBrazi06, line 76: Better name than 'map'?

src/integration/asset/iothink/IothinkAssetIntegration.ts
```diff
@@ -65,24 +65,67 @@ export default class IothinkAssetIntegration extends AssetIntegration<AssetSetti
...
+    }
+    // Create Array with merged consumptions from map
+    const mergedResponse = Array.from(map.values());
+    // Sort the Array according the timestamp
+    mergedResponse.sort((a, b) => a.timestamp - b.timestamp);
     const energyDirection = asset.assetType === AssetType.PRODUCTION ? -1 : 1;
-    if (!Utils.isEmptyArray(data.historics)) {
-      for (let i = 0; i < data.historics[0].logs.length; i++) {
+    if (!Utils.isEmptyArray(mergedResponse)) {
+      for (const mergedConsumption of mergedResponse) {
```
LucasBrazi06, line 86: Add a type of mergedResponse?

src/integration/asset/iothink/IothinkAssetIntegration.ts
```diff
@@ -65,24 +65,67 @@ export default class IothinkAssetIntegration extends AssetIntegration<AssetSetti
...
+            if (asset.siteArea?.voltage) {
+              consumption.currentInstantAmps = consumption.currentInstantWatts / asset.siteArea.voltage;
+            }
+            consumption.lastConsumption = {
+              timestamp: moment(this.timestampReference).add(mergedConsumption.timestamp, 'seconds').toDate(),
+              value: consumption.currentTotalConsumptionWh
+            };
+            break;
+          case AssetType.CONSUMPTION_AND_PRODUCTION:
+            consumption.currentInstantWatts = this.getPropertyValue(mergedConsumption, IothinkProperty.IO_POW_ACTIVE) * 1000;
```
LucasBrazi06, line 116: Use decimal.js for your computation with float?
Double check everywhere

src/integration/asset/iothink/IothinkAssetIntegration.ts
```diff
@@ -65,24 +65,67 @@ export default class IothinkAssetIntegration extends AssetIntegration<AssetSetti
...
+            consumption.currentStateOfCharge = this.getPropertyValue(mergedConsumption, IothinkProperty.IO_SOC);
+            consumption.currentConsumptionWh = this.getPropertyValue(mergedConsumption, IothinkProperty.IO_ENERGY_DISCHARGE)
+            - this.getPropertyValue(mergedConsumption, IothinkProperty.IO_ENERGY_CHARGE);
+            if (asset.siteArea?.voltage) {
+              consumption.currentInstantAmps = consumption.currentInstantWatts / asset.siteArea.voltage;
```
LucasBrazi06, line 121: decimal.js

src/types/Asset.ts
```diff
@@ -44,8 +44,22 @@ export enum SchneiderProperty {
 }
 
 export enum IothinkProperty {
-  POWER_ACTIVE = 'Puissance_active_totale',
   POWER_L1 = 'Puissance_active_de_phase_1',
   POWER_L2 = 'Puissance_active_de_phase_2',
   POWER_L3 = 'Puissance_active_de_phase_3',
+
+  // Generic
+  IO_POW_ACTIVE = 'io-pow-active',
```
LucasBrazi06, line 52: Why not creating an IoThinkAPIProperty enum?

src/integration/asset/iothink/IothinkAssetIntegration.ts
```diff
@@ -65,24 +65,67 @@ export default class IothinkAssetIntegration extends AssetIntegration<AssetSetti
...
+    // Create Array with merged consumptions from map
+    const mergedResponse = Array.from(map.values());
+    // Sort the Array according the timestamp
+    mergedResponse.sort((a, b) => a.timestamp - b.timestamp);
     const energyDirection = asset.assetType === AssetType.PRODUCTION ? -1 : 1;
-    if (!Utils.isEmptyArray(data.historics)) {
-      for (let i = 0; i < data.historics[0].logs.length; i++) {
+    if (!Utils.isEmptyArray(mergedResponse)) {
+      for (const mergedConsumption of mergedResponse) {
+        if (typeof mergedConsumption[IothinkProperty.IO_POW_ACTIVE] === 'undefined') {
```
LucasBrazi06, line 87: IothinkProperty -> IoThinkProperty

LucasBrazi06 (changes requested):
Few small changes :-)

OliveGerste:
> @OliveGerste Does Gerald is aware about that? Any action from ioThink's side?
> Thanks.

There was an e-mail conversation about this so they should be aware. I will ask for the current status.

src/integration/asset/iothink/IothinkAssetIntegration.ts
OliveGerste: @LucasBrazi06  
Iothink is consistently written like this in the application. 
I did this because it is the company name and in the end one word.
Should this be changed in the whole Application (Front end and back end?)

LucasBrazi06: That's okay, I can live with this ;-).
I close the conversation.

src/types/Asset.ts
OliveGerste: @LucasBrazi06 
Could you explain this a bit more? 
This is already an enum called IothinkProperty similar to the SchneiderProperty one.

OliveGerste: I will probably remove the values above the added ones, because they were used for the parking sud device and I think we will not need them anymore.

LucasBrazi06: An enum type that will contain constants (POWER_L1...)
And we may need to get the power per phase in the Optimizer from the Asset as discussed.

OliveGerste: POWER_L1 etc. are already in an enum type.
The problem is with the new devices for solar, battery and building we do not get the power per phase anymore.

LucasBrazi06: > POWER_L1 etc. are already in an enum type.
> The problem is with the new devices for solar, battery and building we do not get the power per phase anymore.

Weird, I haven't seen the enum in the first place.
That's ok then!
For the rest, if the power is not provided per phase and we have a 3 phases site area, divide it by 3.

src/integration/asset/iothink/IothinkAssetIntegration.ts
OliveGerste: A type of the mergedResponse would look like the property enum with an added timestamp: 
```
export interface IothinkMergedResponse {
  timestamp: number;
  Puissance_active_de_phase_1?: string;
  Puissance_active_de_phase_2:? string;
  Puissance_active_de_phase_3:? string;
  'io-pow-active'?: string;
  'io-energy-input'?: string;
  'io-energy-output'?: string;
  'io-energy'?: string;
  'io-energy-charge'?: string;
  'io-energy-discharge'?: string;
  'io-soc'?: string;
}
```
If we do it like this, the enum is obsolete and can be removed.

LucasBrazi06: Okay.
But we have now just 3 weird namings here, could we separate them?

OliveGerste: This is how the property names of the response are called. 
That is why I used the enum to get the property values from the merged Array response. Do you want to separate the different namings in three different types or do you want blank lines between the different namings?

LucasBrazi06: If it matches the ioThink endpoint, let's keep it as it is then.

OliveGerste:
@LucasBrazi06 
I implemented the requested changes and I think it should be fine now. Let me know if I missed something.
As discussed I kept the enum procedure for filtering  the response.
Regarding the power per phase: I think there was a comment in a resolved conversation. 
We handle this separation in the smart charging. So currently there should be no need to divide the consumption by the number of phases of the site area. We have no added value from this.

LucasBrazi06:
> @LucasBrazi06
> I implemented the requested changes and I think it should be fine now. Let me know if I missed something.
> As discussed I kept the enum procedure for filtering the response.
> Regarding the power per phase: I think there was a comment in a resolved conversation.
> We handle this separation in the smart charging. So currently there should be no need to divide the consumption by the number of phases of the site area. We have no added value from this.

Ok, perfect.
Thanks!

LucasBrazi06 (approved):

PR #2385, opened 2021-04-28
Phase 1: Server initialisation code factor out. 

jerome-benoit:
+ remove unused attributes; 
+ use inheritance for server configuration tunables; 
+ introduce server types;
+ introduce uws helpers. 

I would like to have that PR merged first, I think I've found a way to avoid to the race condition on WS connection caching on top of it. 

Thanks. 


src/server/uWsUtils.ts
```diff
@@ -0,0 +1,53 @@
+import { App, AppOptions, HttpRequest, HttpResponse, SSLApp, TemplatedApp } from 'uWebSockets.js';
```
LucasBrazi06, line 1: Naming of the class uWsUtils.ts -> UWsUtils.ts? :-)

src/server/uWsUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+      serverName: string, serverModuleName: string, listenCb?: () => void, listen = true): void {
+    let cb: () => void;
+    if (listenCb && typeof listenCb === 'function') {
+      cb = listenCb;
+    } else {
+      cb = async () => {
+        await ServerUtils.defaultListenCb(serverModuleName, 'startServer', serverName, serverConfig.protocol, serverConfig.host ?? '::', serverConfig.port);
+      };
+    }
+
```
LucasBrazi06, line 37: BL

LucasBrazi06 (approved):
GUT!
Tu t'es bien amusé ;-)

src/server/uWsUtils.ts
jerome-benoit: The official short name of the library is uWs, I've just reused it.

LucasBrazi06: > The official short name of the library is uWs, I've just reused it.

I know but here this is a class name and the lib doesn't have any classes that start with a minus letter with what I see in the import ;-)

PR #2390, opened 2021-04-29
RFC: Renable retry on axios and disable it on ut.

jerome-benoit:
We can disable it for on other call site if needed.

### DIFF
--- a/src/utils/Configuration.ts
+++ b/src/utils/Configuration.ts
@@ -420,7 +420,7 @@ export default class Configuration {
       Configuration.getConfig().Axios.timeout = Constants.AXIOS_DEFAULT_TIMEOUT;
     }
     if (Configuration.isUndefined(Configuration.getConfig().Axios.retries)) {
-      Configuration.getConfig().Axios.retries = 0;
+      Configuration.getConfig().Axios.retries = 3;
     }
     return Configuration.getConfig().Axios;
   }
--- a/test/api/client/utils/BaseApi.ts
+++ b/test/api/client/utils/BaseApi.ts
@@ -1,6 +1,7 @@
 import AxiosFactory from '../../../../src/utils/AxiosFactory';
 import { AxiosRequestConfig } from 'axios';
 import Constants from '../../../../src/utils/Constants';
+import { IAxiosRetryConfig } from 'axios-retry';
 import config from '../../../config';
 import { performance } from 'perf_hooks';
 import querystring from 'querystring';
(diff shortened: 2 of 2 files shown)


LucasBrazi06:
As I wrote, I don't want to enable the retry.

jerome-benoit:
That was meant to be merged after the test, and the ut part is valid and ensure that retry is left disabled in them whatever is put in the config.json file as axios configuration.

In short, you've closed a PR with code in it that actually fix the ut behaviour in all case.

LucasBrazi06:
I wanted this behavior in the whole project

jerome-benoit:
The behaviour can be overridden by the configuration, the ut use axios to mimic the dashboard or mobile clients that never retry. In order to ensure the ut always behave like the clients, that code is necessary. Otherwise the ut behaviour can be not consistent with the clients one. like @ClaudeROSSI remarked, which is not is expected in ut.

LucasBrazi06:
I want this behavior in the usage everywhere in the projet and feature that needs retry will provide explicitly the conf.

jerome-benoit:
The ut never need to retry, letting the code allows them to eventually retry is a bug.

LucasBrazi06:
Today with the default Axios config, by default, there is not retry included unit tests if I'm not mistaken.

jerome-benoit:
> Today with the default Axios config, by default, there is not retry included unit tests if I'm not mistaken.

Sure but the code still permits them to be enabled, which will lead to the previous misbehaviour reported by @ClaudeROSSI. With the other PR, it just can't happen anymore, whatever is changed in the configuration file and the ut behave then consistently like the clients do. That was the only purpose of the other PR.

LucasBrazi06:
> Sure but the code still permits it to be enabled, which will lead to the previous misbehaviour reported by @ClaudeROSSI. With the other PR, it just can't happen anymore, whatever is changed in the configuration file and the ut behave then consistently like the clients do. That was the only purpose of the other PR.

I decided to extend this behavior to the whole backend, so it should be okay by now I think.

jerome-benoit:
I think I will split put the axios ut configuration in the ut configuration file and do an other PR ;) That will be easier than trying to explain that reusing the global configuration for ut is not the correct.

LucasBrazi06:
lol
The global conf should not retry either so it should be consistent I think with the UT.
Indeed, we can have a talk when I'll be on site, no problem :-)

jerome-benoit:
I'll do the PR, you will understand. For the WSClient, we already do the same: ut have it's own configuration to set it up, so ut can behave the way the want independently of the ev-server configuration, default behaviour.

PR #2396, opened 2021-04-29
Linter silencing and cleanups.

jerome-benoit:

### DIFF
--- a/src/server/ocpp/json/JsonRestWSConnection.ts
+++ b/src/server/ocpp/json/JsonRestWSConnection.ts
@@ -40,7 +40,7 @@ export default class JsonRestWSConnection extends WSConnection {
 
   public onError(errorEvent: ErrorEvent): void {
     // Log
-    Logging.logError({
+    void Logging.logError({
       tenantID: this.getTenantID(),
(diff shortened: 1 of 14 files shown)


src/server/ocpp/json/JsonRestWSConnection.ts
```diff
@@ -40,7 +40,7 @@ export default class JsonRestWSConnection extends WSConnection {
 
   public onError(errorEvent: ErrorEvent): void {
     // Log
-    Logging.logError({
+    void Logging.logError({
```
LucasBrazi06, line 43: Be careful, I sometimes lost logs in prod because of not putting the 'await' keyword.

src/server/ocpp/json/JsonWSConnection.ts
```diff
@@ -86,7 +88,7 @@ export default class JsonWSConnection extends WSConnection {
 
   public onError(errorEvent: ErrorEvent): void {
     // Log
-    Logging.logError({
+    void Logging.logError({
```
LucasBrazi06, line 91: Same comment, avoid 'void'

LucasBrazi06 (approved):

PR #2408, opened 2021-05-04
Billing - minor fixes for beta

ClaudeROSSI:
Some minor changes and fixes

### DIFF
--- a/test/api/BillingStripeTestData.ts
+++ b/test/api/BillingStripeTestData.ts
@@ -18,7 +18,6 @@ import User from '../../src/types/User';
 import UserStorage from '../../src/storage/mongodb/UserStorage';
 import chaiSubset from 'chai-subset';
 import config from '../config';
-import moment from 'moment';
 import responseHelper from '../helpers/responseHelper';
 
 chai.use(chaiSubset);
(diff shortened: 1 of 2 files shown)


test/api/BillingStripeTestData.ts
```diff
@@ -231,7 +230,7 @@ export default class StripeIntegrationTestData {
     expect(lastPaidInvoice.amount).to.be.eq(expectedTotal); // 480 cents - TODO - Billing Invoice exposing cents???
     const lastPaidInvoiceDateTime = new Date(lastPaidInvoice.createdOn).getTime();
     // Stripe is using Unix Epoch for its date - and looses some precision
-    expect(lastPaidInvoiceDateTime).to.be.gte(moment.unix(beforeInvoiceDateTime / 1000).toDate().getTime());
+    expect(Math.trunc(lastPaidInvoiceDateTime / 1000)).to.be.gte(Math.trunc(beforeInvoiceDateTime / 1000));
```
LucasBrazi06, line 233: Use decimal lib ? Utils.createDecimal() ?

ClaudeROSSI: this is just a workaround in the test because the stripe date is in the past when two invoices are generated within the same second.
Why should we use createDecimals here?

LucasBrazi06: > Why should we use createDecimals here?

Yes, every float manipulation requires it.

ClaudeROSSI: done

LucasBrazi06 (approved):

PR #2410, opened 2021-05-05
Dedupe MeterValues in enum + cleanups

jerome-benoit:
Update: 

+ reverted to previous behaviour on the phase assignment. 

### DIFF
--- a/src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/OCPIUtilsService.ts
+++ b/src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/OCPIUtilsService.ts
@@ -170,7 +170,7 @@ export default class OCPIUtilsService {
   }
 
   public static async convertSite2Location(tenant: Tenant, site: Site,
-      options: OCPILocationOptions, withChargingStations): Promise<OCPILocation> {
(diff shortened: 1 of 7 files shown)


OliveGerste (commented):
We have a similar check in the charging station service. Should this be updated too?

jerome-benoit:
> We have a similar check in the charging station service. Should this be updated too?

No, customers move a charging station three-phased in the template to a one phase site area but when they configure it manually, the backend have to make sure the number of its phase math the site area.

OliveGerste:
> > We have a similar check in the charging station service. Should this be updated too?
> 
> No, customers move a charging station three-phased in the template to a one phase site area but when they configure it manually, the backend have to make sure the number of its phase match the site area.

Okay, so we can keep it as it is.

jerome-benoit:
> > > We have a similar check in the charging station service. Should this be updated too?
> > 
> > 
> > No, customers move a charging station three-phased in the template to a one phase site area but when they configure it manually, the backend have to make sure the number of its phase match the site area.
> 
> Okay, so we can keep it as it is.

@OliveGerste: I you approve that PR, could please flag it as it? That will ease Serge work at his return.

LucasBrazi06:
Why does the customer can't first set up his charging station to single phase and next assign it to the site area?

PR #2414, opened 2021-05-10
Billing - all feature toggles on

ClaudeROSSI:
Activating all features for our first dead line:

- All features are ON
- Granting back BILLING permissions to BASIC users
- Checking for Payment Method at transaction start (thorw an exception if nop PM)
- UT Billing - some test adaptation

### DIFF
--- a/src/server/ocpp/utils/OCPPUtils.ts
+++ b/src/server/ocpp/utils/OCPPUtils.ts
@@ -340,6 +340,7 @@ export default class OCPPUtils {
               module: MODULE_NAME, method: 'billTransaction',
               message, detailedMessages: { error: error.message, stack: error.stack }
             });
+            // Prevent from starting a transaction when Billing prerequisites are not met
             throw new BackendError({
               user: transaction.user,
               action: ServerAction.BILLING_TRANSACTION,
@@ -367,12 +368,6 @@ export default class OCPPUtils {
               module: MODULE_NAME, method: 'billTransaction',
               message, detailedMessages: { error: error.message, stack: error.stack }
             });
-            throw new BackendError({
-              user: transaction.user,
-              action: ServerAction.BILLING_TRANSACTION,
-              module: MODULE_NAME, method: 'billTransaction',
-              message, detailedMessages: { error: error.message, stack: error.stack }
-            });
           }
           break;
         // Stop Transaction
@@ -391,12 +386,6 @@ export default class OCPPUtils {
               tenantID: tenantID,
               user: transaction.userID,
               source: Constants.CENTRAL_SERVER,
-              action: ServerAction.BILLING_TRANSACTION,
-              module: MODULE_NAME, method: 'billTransaction',
-              message, detailedMessages: { error: error.message, stack: error.stack }
-            });
-            throw new BackendError({
-              user: transaction.user,
               action: ServerAction.BILLING_TRANSACTION,
               module: MODULE_NAME, method: 'billTransaction',
               message, detailedMessages: { error: error.message, stack: error.stack }
--- a/src/server/ocpp/utils/OCPPUtils.ts
+++ b/src/server/ocpp/utils/OCPPUtils.ts
@@ -331,14 +331,20 @@ export default class OCPPUtils {
               lastUpdate: new Date()
             };
           } catch (error) {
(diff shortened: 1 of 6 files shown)


src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -331,14 +331,21 @@ export default class OCPPUtils {
...
-              message: `Failed to bill the Transaction ID '${transaction.id}'`,
-              detailedMessages: { error: error.message, stack: error.stack }
+              message, detailedMessages: { error: error.message, stack: error.stack }
+            });
+            // Prevent from starting a transaction when Billing prerequisites are not met
+            throw new BackendError({
+              user: transaction.user,
+              action: ServerAction.BILLING_TRANSACTION,
+              module: MODULE_NAME, method: 'billTransaction',
+              message, detailedMessages: { error: error.message, stack: error.stack }
```
jerome-benoit, line 348: Could you please explain us the conditions that trigger the error throwing (to ease the PR review without proper `git checkout` on the branch)?

I think error throwing is fine to stop the start transaction, maybe a bit brutal. 
Can you return a boolean in that static method that will tell if the transaction is allowed or not?
The method will keep logging the details and the caller will then be able report a short message to the end users. Maybe you can also return an object like `{ cardSubcriptionError: true, updateTransactionError: false, etc. }` to permit to display more accurate error messages to end users

src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -331,14 +331,21 @@ export default class OCPPUtils {
               lastUpdate: new Date()
             };
           } catch (error) {
+            const message = `Billing - startTransaction failed - transaction ID '${transaction.id}'`;
```
jerome-benoit, line 334: Not sure Serge will like the message factorisation in that PR ;)

jerome-benoit (changes requested):

src/server/ocpp/utils/OCPPUtils.ts
ClaudeROSSI: When billing is ON, there is a lot of situations where a start transaction should be rejected. The main reasons are the ones below:
- User is not in sync (i.e.: customer does not yet exist on the stripe side)
- customerID is inconsistent (i.e.: the customer has been deleted on the stripe side)
- billing settings are incomplete (e.g.: Tax Rate has been archived and cannot be used anymore)
- Customer has no payment method (no Card information)
- Customer has a card method but the card has expired
- Stripe connection fails (i.e.: STRIPE server not responding)
- etc

As you can see we have a lot of different issues that could impact the start transaction phase.
By throwing an exception I simply prevent the transaction from starting. But with the current  OCPPUtils coding there is no way to provide a clear feedback (with the root cause of the problem) to the end user.
Don't we already have such use-case where we need to prevent a start transaction and provide a clear feedback to the end user?
Thanks

LucasBrazi06: No 😔

LucasBrazi06: It's rejected and we check the logs.

ClaudeROSSI: Welcome back Serge ;)
So my understanding is that we can merge it as it is (i.e.: throwing that BackendError)  and we will have to think about a way to notify the user that his/her payment method has been refused.

jerome-benoit: I think so, the error message reporting refinement can be done later one.

ClaudeROSSI:
BILLING module has been switched OFF on SLF

So, this PR can be merged with no side effect on PROD.

ClaudeROSSI:
**utall** tests are failing on "Can read consumption of a started transaction ..." 
is this a new problem?

Thanks

**Test output**
> 1) Transaction tests
       With components Organization and Pricing (tenant utall)
         Where basic user
           Using function "readAllConsumption"
             Can read consumption of a started transaction without meter values:

      AssertionError: expected 403 to equal 200
      + expected - actual

      -403
      +200

ClaudeROSSI:
FYI: Transaction tests are still failing but I have the same behavior when testing master-qa.

jerome-benoit (approved):

PR #2416, opened 2021-05-10
OCPI: opening times need to disable twentyfourseven if hours are

jerome-benoit:
specified


src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/OCPIUtilsService.ts
```diff
@@ -226,7 +226,8 @@ export default class OCPIUtilsService {
                   period_begin: '08:00',
                   period_end: '18:00'
                 },
-              ]
+              ],
+              twentyfourseven: false
```
LucasBrazi06, line 230: Weird, the doc says 'Choice: one of two'

image

PR #2427, opened 2021-05-12
Tags csv import back

cabch:
user is added with status pending if it does not exist
tag is assigned to the user if added or found
tags are imported to importedTags with users inside and processed in async task


src/utils/Utils.ts
```diff
@@ -1444,7 +1444,14 @@ export default class Utils {
     return Configuration.isCloudFoundry() ? cfenv.getAppEnv().name : os.hostname();
   }
 
-  public static replaceDoubleQuotes(value: any): string {
+  public static escapeCsvValues(value: any): string {
+    // replace starting and ending double quotes by nothing
+    // replace double quotes inside value to double double quotes to display double quote in csv editor
     return typeof value === 'string' ? '"' + value.replace(/^"|"$/g, '').replace(/"/g, '""') + '"' : value;
```
ClaudeROSSI, line 1450: I was expecting that this method also encodes the comma, etc ...
to make sure a value like "Yes, we can" can be imported.

ClaudeROSSI (changes requested):
Please call me to discuss my comments.

ClaudeROSSI:
Why don't we rely on something like:
https://github.com/mafintosh/csv-parser

src/utils/Utils.ts
cabch: as we double quote the csv values to handle the comma inside a row value, here we double the double quotes so they are considered as simple double quotes when we display the value in excel for example

src/utils/Utils.ts
```diff
@@ -1444,7 +1444,14 @@ export default class Utils {
     return Configuration.isCloudFoundry() ? cfenv.getAppEnv().name : os.hostname();
   }
 
-  public static replaceDoubleQuotes(value: any): string {
+  public static escapeCsvValues(value: any): string {
```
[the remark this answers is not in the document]
cabch: renamed ✅

src/utils/Utils.ts
```diff
@@ -1444,7 +1444,14 @@ export default class Utils {
...
+    // replace starting and ending double quotes by nothing
+    // replace double quotes inside value to double double quotes to display double quote in csv editor
     return typeof value === 'string' ? '"' + value.replace(/^"|"$/g, '').replace(/"/g, '""') + '"' : value;
   }
+
+  public static unEscapeCsvValues(value: any): string {
```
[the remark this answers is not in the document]
cabch: modified ✅

src/utils/Utils.ts
cabch: And like that, no need to specifically handle the csv separator as it is inside the double quotes

src/async-task/tasks/TagsImportAsyncTask.ts
```diff
@@ -75,6 +82,10 @@ export default class TagsImportAsyncTask extends AbstractAsyncTask {
                 createdBy: { id: importedTag.importedBy },
                 createdOn: importedTag.importedOn,
               };
+              // Save user if any and get the ID to assign tag
+              if (importedTag.email && importedTag.name && importedTag.firstName) {
+                await this.assignTag(tenant, importedTag, tag);
```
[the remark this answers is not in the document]
cabch: done ✅

jerome-benoit:
@cabch: Could you please sync with master-qa branch and ensure the unit tests are run successfully on GitHub ? 

Thanks.

jerome-benoit:
@roxiciobotea210: Could you please sync with master-qa branch and ensure the unit tests are run successfully on GitHub ? 

Thanks.

src/async-task/tasks/TagsImportAsyncTask.ts
```diff
@@ -95,7 +99,7 @@ export default class TagsImportAsyncTask extends AbstractAsyncTask {
                 message: `Error when importing Tag ID '${importedTag.id}': ${error.message}`,
                 detailedMessages: { tag: importedTag, error: error.message, stack: error.stack }
               });
-            }
+            } // Do we add a finally that delete tags if the error comes from "foundTag" ? for now the found tags are left as is in the importedTag collection
```
ClaudeROSSI, line 102: Let's stick to the import of users logic. No finally required!
The error is for now simply stored (on line 93). That way we know that the import of a particular tag failed or not.
I guess we will introduce later a UI to cleanup all tags in errors

ClaudeROSSI: i think you can remove this comment then

src/server/rest/v1/service/TagService.ts
```diff
@@ -760,14 +761,24 @@ export default class TagService {
...
       // Validate Tag data
       TagValidator.getInstance().validateImportedTagCreation(newImportedTag);
       // Set properties
       newImportedTag.importedBy = importedTag.importedBy;
       newImportedTag.importedOn = importedTag.importedOn;
       newImportedTag.status = ImportStatus.READY;
+      try {
+        UserValidator.getInstance().validateImportedUserCreation(newImportedTag as ImportedUser);
+      } catch (error) {
+        newImportedTag.email = '';
```
ClaudeROSSI, line 778: i think it makes sense to log the error here (maybe as a warning ...)

src/storage/mongodb/TagStorage.ts
```diff
@@ -42,7 +42,10 @@ export default class TagStorage {
     const uniqueTimerID = Logging.traceStart(tenantID, MODULE_NAME, 'saveImportedTag');
     const tagMDB = {
       _id: importedTagToSave.id,
-      description: importedTagToSave.description,
+      description: Utils.unescapeCsvValue(importedTagToSave.description),
+      name: Utils.unescapeCsvValue(importedTagToSave.name),
+      firstName: Utils.unescapeCsvValue(importedTagToSave.firstName),
+      email: Utils.unescapeCsvValue(importedTagToSave.email),
```
ClaudeROSSI, line 48: I think calling the unescape here this is not correct.

You do not know if you have to unescape because the property values may come either from the importedTag or from the existing tag (already unescaped).

The unescape must be done by the calling layer:
at the beginning of the for loop:
>> for (const importedTag of importedTags.result) {

src/storage/mongodb/TagStorage.ts
```diff
@@ -62,7 +65,10 @@ export default class TagStorage {
     const uniqueTimerID = Logging.traceStart(tenantID, MODULE_NAME, 'saveImportedTags');
     const importedTagsToSaveMDB: any = importedTagsToSave.map((importedTagToSave) => ({
       _id: importedTagToSave.id,
-      description: importedTagToSave.description,
+      description: Utils.unescapeCsvValue(importedTagToSave.description),
+      name: Utils.unescapeCsvValue(importedTagToSave.name),
+      firstName: Utils.unescapeCsvValue(importedTagToSave.firstName),
+      email: Utils.unescapeCsvValue(importedTagToSave.email),
```
ClaudeROSSI, line 71: idem. Please call me ;)

src/utils/Utils.ts
```diff
@@ -1503,7 +1503,14 @@ export default class Utils {
...
-    return typeof value === 'string' ? '"' + value.replace(/^"|"$/g, '').replace(/"/g, '""') + '"' : value;
+  public static escapeCsvValue(value: any): string {
+    // replace starting and ending double quotes by nothing
+    // replace double quotes inside value to double double quotes to display double quote in csv editor
+    return typeof value === 'string' ? '"' + value.replace(/"/g, '""') + '"' : value;
+  }
+
+  public static unescapeCsvValue(value: any): string {
+    // replace double double quotes inside value into simple double quotes
+    return typeof value === 'string' ? value.replace(/^"|"$/g, '').replace(/""/g, '"') : value;
```
ClaudeROSSI, line 1514: to be on the safe side i would suggest to throw an exception when a string value does not start and end with a double quote.
This should not happen. If this happens, it means that the calling layer is trying to unescape something which was not escaped properly. This will produce unpredictable results.

ClaudeROSSI (changes requested):
Please check my comments

src/utils/Utils.ts
cabch: in fact starting and ending double quotes are handled by the csvToJson during import so at that Utils.unescapeCsvValue time there is no more double quotes at the begining and at the end

src/storage/mongodb/TagStorage.ts
cabch: done ✅

src/storage/mongodb/TagStorage.ts
cabch: you're right ✅

src/server/rest/v1/service/TagService.ts
cabch: done ✅

src/async-task/tasks/TagsImportAsyncTask.ts
cabch: ok ✅

cabch:
> @cabch: Could you please sync with master-qa branch and ensure the unit tests are run successfully on GitHub ?
> 
> Thanks.

Would have been glad to do it by myself ;) sounds like it's ok now on GitHub

src/utils/Utils.ts
```diff
@@ -1503,14 +1503,17 @@ export default class Utils {
...
-    // replace starting and ending double quotes by nothing
-    // replace double quotes inside value to double double quotes to display double quote in csv editor
+    // add double quote start and end
+    // replace double quotes inside value to double double quotes to display double quote correctly in csv editor
     return typeof value === 'string' ? '"' + value.replace(/"/g, '""') + '"' : value;
   }
 
+  // when importing values
   public static unescapeCsvValue(value: any): string {
+    // starting and ending double quotes are handle by csvToJson
```
ClaudeROSSI, line 1515: Do you mean that csvToJson already unescape the values? 
That would be good, but then there is no need to unescape values anymore.

To be on the safe side, we need to make sure the logic used during the export is compliant with the one csvToJon() applies when unescaping.

ClaudeROSSI (commented):
we need to clarify what logic cvsToJson applies when converting imported values

src/utils/Utils.ts
cabch: yes csvtojson already escape by default double quotes at the beginning and the end.

the reason why we need to keep unescaping is for replacing the double double quote by simple double quotes

ClaudeROSSI: ok. unescapeCsvValue() does nothing because the csvToJson lib does it for us.

ClaudeROSSI (approved):

PR #2435, opened 2021-05-18
Fix transaction tests

roxiciobotea210:
Add `userID` field back into projectFields. It is needed for the check in `Authorizations.canReadTransaction`. 
It is removed later in the code in case the user who made the request does not have the appropriate permissions (READ on USER entity).


jerome-benoit (approved):
Since we need to deploy in production, I will merge it as is. But I think a way to ensure the required `userID` field  is here in the authorisation check  code must be added and log an exhaustive message on error.

PR #2457, opened 2021-05-21
Fix consumption instantAmps calculation from metervalues

jerome-benoit:

### DIFF
--- a/src/server/ocpp/utils/OCPPUtils.ts
+++ b/src/server/ocpp/utils/OCPPUtils.ts
@@ -509,8 +509,7 @@ export default class OCPPUtils {
       transaction.stateOfCharge = Utils.convertToInt(transaction.currentStateOfCharge);
     }
     transaction.currentTotalDurationSecs = moment.duration(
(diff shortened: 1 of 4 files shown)


src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -822,7 +820,9 @@ export default class OCPPUtils {
       stopMeterValues.push({
         id: (id++).toString(),
         ...meterValueBasedProps,
-        value: (transaction.currentInstantAmps ? transaction.currentInstantAmps : transaction.currentInstantAmpsDC),
+        value: (transaction.currentInstantAmps
+          ? transaction.currentInstantAmps / Utils.getNumberOfConnectedPhases(await ChargingStationStorage.getChargingStation(tenantID, transaction.chargeBoxID), null, transaction.connectorId)
```
OliveGerste, line 824: We could add the charging station as a parameter to the method. So we can avoid this extra database call.
It should be available in both locations, where the method is called.

jerome-benoit: Done.

PR #2458, opened 2021-05-21
Bug fix authorization

roxiciobotea210:
Fixes: `admin that are also site admin or owner can't see everything in the organisation tree`

The problem is that extra filters added in case the user is siteOwner or siteAdmin overlap with the fact that an admin can do anything. For now I can add a check that prevents dynamic filters from being applied in case the user is an admin. But the issue will need to be revisited when Serge is back to see what is the elegant way to address it. (I think that after all the entities will be updated to the new auth concept, it won't be an issue anymore, but I need to check with Serge.)

### DIFF
--- a/src/server/rest/v1/service/AuthorizationService.ts
+++ b/src/server/rest/v1/service/AuthorizationService.ts
@@ -723,7 +723,7 @@ export default class AuthorizationService {
 
   private static async processDynamicFilters(tenant: Tenant, userToken: UserToken, action: Action, entity: Entity,
       authorizationFilters: AuthorizationFilter, authorizationContext: AuthorizationContext, extraFilters?: Record<string, any>): Promise<void> {
-    if (!Utils.isEmptyArray(authorizationContext.filters)) {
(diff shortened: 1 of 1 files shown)


src/server/rest/v1/service/AuthorizationService.ts
```diff
@@ -723,7 +723,7 @@ export default class AuthorizationService {
 
   private static async processDynamicFilters(tenant: Tenant, userToken: UserToken, action: Action, entity: Entity,
       authorizationFilters: AuthorizationFilter, authorizationContext: AuthorizationContext, extraFilters?: Record<string, any>): Promise<void> {
-    if (!Utils.isEmptyArray(authorizationContext.filters)) {
+    if (userToken.role !== UserRole.ADMIN && !Utils.isEmptyArray(authorizationContext.filters)) {
```
jerome-benoit, line 726: Could you please add a comment (TODO) here explaining a better way should be found to 'merge' the 'merge' the authorization if a user has several role? 

Thanks.

jerome-benoit (changes requested):

jerome-benoit (approved):

PR #2460, opened 2021-05-21
Add TagID filter on GET /Users request

roxiciobotea210:
The filter is used in calls form the frontend but it was not implemented on the backend side.


src/server/rest/v1/service/UserService.ts
```diff
@@ -1316,6 +1316,7 @@ export default class UserService {
         issuer: filteredRequest.Issuer,
         siteIDs: (filteredRequest.SiteID ? filteredRequest.SiteID.split('|') : null),
         userIDs: (filteredRequest.UserID ? filteredRequest.UserID.split('|') : null),
+        tagID: (filteredRequest.TagID ? filteredRequest.TagID : null),
```
jerome-benoit, line 1319: Does the filter allow several tags ID? 
If yes, you have to use an array.

jerome-benoit (commented):
One question.

src/server/rest/v1/service/UserService.ts
roxiciobotea210: No, at the moment the filter only supports on TagID.

roxiciobotea210: Thanks for mentioning, it does need to support several tag IDs! Working on the change now.

roxiciobotea210:
@jerome-benoit Done.

jerome-benoit (approved):

PR #2465, opened 2021-05-21
Asset - connection refresh interval

OliveGerste:
The interval has been added to the asset connection setting, to handle cases where the provider is only able to refresh in a certain interval for all assets.


jerome-benoit (commented):
I have a question: 
The retrieval is done via a scheduler task? 
That's why you do not rely on `setInterval()`?

OliveGerste:
> I have a question:
> The retrieval is done via a scheduler task?
> That's why you do not rely on `setInterval()`?

Yes the retrieval is done by a scheduler task with an interval of 1 minute. 
But specific asset integrations like wit have a refresh interval of 10 minutes.

PR #2467, opened 2021-05-24
sap cp xsuaa integration (draft)

mocotrah:
Notes:
The following solution will use the AppRouter as the single entry point for all emobility modules (dashboard, rest server, etc.).
All calls from the web app (dashboard) to the rest server will pass via AppRouter.
This reason of this approach is because we haven't succeeded to use 2 passports in the AuthService.ts class. If we find a way to do that, we can discuss about using direct calls to the rest server (but what about cross origin then?).

(see Visio diagram in archi folder)

Other ideas:
- maintain the ev-server using always the emobility token but in a different header than "Authorization". Have an intermediate processing in AppRouter side, during the login phase, to login using a technical user but transfer in the payload the user id which logged in via xsuaa in the web app. This call will return the emobility token of the end user back. Subsequent calls to the rest server will be made directly from the webapp using the emobility token in custom header. This solution implies using technical user/pass in 2 env vars in the approuter deployed app.
- investigate using 2 passports in ev-server
- Question: is the cross origin protection activated ?
- Question: is the csrf protection activated ?

Other details:
- added new AdvancedConfiguration.ts to provide a parameter to clearly distinguish between built-in login and xsuaa login (this will enable us to hide the classical login route "/login")
- because of AppRouter, the rest server will receive the header "Authorization: <xsuaatoken>" which will be validated in AuthService. In addition the web app (which will store the emobility token in its local storage) will send the emobility token in a second header named "emobilitytoken".
- the principle of xsuaa is to hide the jwt token from the browser application. The web application will only receive a JSESSIONID which the browser will send to the approuter and the approuter will inject the Authorization: <xsuaatoken>
- when the web app is launched, it will check if it has a emobility token in its local storage. If not, it will go to the "login" web page and make an automatic call to "/loginxsuaa" server route. The server will send back the emobility token. 
- for the subsequent calls, the web app will send the emobility token in the special header, and the server will read it and use it.
- must discuss about the end-user onboarding process. For the moment, during login xsuaa, if the user is not found in the database, he will get a error (user not found).


LucasBrazi06:
@mocotrah Maybe I'm wrong but every dev for the SAP standard should be done in the SAP internal repository, right?
Let me know.

mocotrah:
> @mocotrah Maybe I'm wrong but every dev for the SAP standard should be done in the SAP internal repository, right?
> Let me know.

yes, you are right, Jerome asked me to create a draft PR with the xsuaa implementation if you are interested to integrate it in the open source repo.

jerome-benoit:
> @mocotrah Maybe I'm wrong but every dev for the SAP standard should be done in the SAP internal repository, right?
> Let me know.

Since we have driven the SSO integration with SAP IdP, I've asked @mocotrah to push the code here for review to ensure it's done in a pluggable fashion for usage in different landscape and different IdPs.
The final goal is to use the existing standard development task force to introduce SSO with IdPs in e-Mobility. 

@mocotrah: I will do a full and formal review when time permits but the initial review shows that the authentication mechanism is not dynamically 'pluggable': you have to detect the type of authentication and switch case to the right helper authentification method and not rely on hardcoded tunable in a configuration file. 

Thanks.

LucasBrazi06:
> Since we have driven the SSO integration with SAP IdP, I've asked @mocotrah to push the code here for review to ensure it's done in a pluggable fashion for usage in different landscape and different IdPs.

If it's to do IDP integration in a world wide standard way, that's ok otherwise NO.
And this is not prio for the time being (none of our customer requested it yet I think.)

mocotrah:
ok I have an idea, I can detect the presence of xsuaa service. I'll take a look and get back to you.

jerome-benoit:
> > Since we have driven the SSO integration with SAP IdP, I've asked @mocotrah to push the code here for review to ensure it's done in a pluggable fashion for usage in different landscape and different IdPs.
> 
> If it's to do IDP integration in a world wide standard way, that's ok otherwise NO.
> And this is not prio for the time being (none of our customer requested it yet I think.)

The authentification mechanism can differ (SAML, Oauth, ...), but in the end we get a token (JWT most of the time) and the code to deal with it shall be generic given a small part that might be needed to cope with IdP specificity.

PR #2474, opened 2021-05-26
Replace tag visualID cypher with a string ObjectID 

mahdibenromdhane:


src/migration/tasks/AddVisualIDPropertyToTagsTask.ts
```diff
@@ -28,7 +28,7 @@ export default class AddVisualIDPropertyToTagsTask extends MigrationTask {
       for (const tagMDB of tagsMDB) {
         await global.database.getCollection<any>(tenant.id, 'tags').findOneAndUpdate(
           { _id: tagMDB._id },
-          { $set: { visualID: Cypher.hash(tagMDB._id) } });
+          { $set: { visualID: new ObjectID().toString() } });
```
mahdibenromdhane, line 31: It's not possible to use the updateMany since the same instance of ObjectID is set to all the documents (duplication error)

jerome-benoit (approved):

PR #2476, opened 2021-05-26
Wit asset integration

OliveGerste:


jerome-benoit:
@OliveGerste: is it Ok to postpone that PR for the next release?

OliveGerste:
> @OliveGerste: is it Ok to postpone that PR for the next release?

Good question. I only know, that the topic is urgent for Monaco. 
@LucasBrazi06 What do you think?

LucasBrazi06:
> > @OliveGerste: is it Ok to postpone that PR for the next release?
> 
> Good question. I only know, that the topic is urgent for Monaco.
> @LucasBrazi06 What do you think?

This PR is not big and should not destabilize the prod, so I'll check and merge it.

LucasBrazi06 (approved):

PR #2480, opened 2021-05-27
billing - fix tests to consume new endpoints

ClaudeROSSI:
Billing Test now using new rest end point for getting Invoices


cabch (approved):
ok pour moi

PR #2484, opened 2021-05-27
Billing - async bill transaction

ClaudeROSSI:
**BILLING_ASYNC_BILL_TRANSACTION** FeatureToggle has been added and activated to postpone all billing operations when called on the **stopTransaction** method.

The impact is the one below:
- On stop transaction, the billing simply creates the asyncTask to postpone the billing flow (billing + payment)
- The t**ransaction.billingData.status** set to "PENDING" and the transaction gets stopped

The **AsyncTaskManager** will trigger the billing flow after a few seconds. The invoice is then generated in a DRAFT state and gets finalized (and eventually paid) accordingly to the billing settings.

Billing Tests have been adapted to wait 3 seconds in order to make sure the async task gets executed. This 3 seconds delay is maybe not sufficient. We need to find a better way to make sure the test behavior remains predictable.


ClaudeROSSI:
@LucasBrazi06 , @jerome-benoit : It sounds like the AsyncTasks are not being executed when running unittests.
The asyncTaskManager config seems correct to me. Only the monitorDBChange flag is missing. Can we activate it when running tests?
Thanks
Calude

image

src/async-task/tasks/BillTransactionAsyncTask.ts
```diff
@@ -0,0 +1,53 @@
...
+        if (billingImpl) {
+
+          // Get the Transaction to bill
+          const transactionID: string = this.asyncTask.parameters.transactionID;
+          const transactionLock = await LockingHelper.createBillTransactionLock(tenant.id, Number(transactionID));
+          if (transactionLock) {
+            try {
+              const transaction = await TransactionStorage.getTransaction(tenant.id, Number(transactionID));
+              if (!transaction) {
+                throw new Error(`Unknown Transaction ID '${this.asyncTask.parameters.transactionID}'`);
```
jerome-benoit, line 29: You have to use the `Error` overloads we have.

src/async-task/tasks/BillTransactionAsyncTask.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export default class BillTransactionAsyncTask extends AbstractAsyncTask {
+  protected async executeAsyncTask(): Promise<void> {
+    const tenant = await TenantStorage.getTenant(this.asyncTask.tenantID);
+    // Check if OCPI component is active
+    if (Utils.isTenantComponentActive(tenant, TenantComponents.BILLING)) {
+      try {
+        const billingImpl = await BillingFactory.getBillingImpl(tenant.id);
+        if (billingImpl) {
+
```
jerome-benoit, line 21: Unneeded blank line.

ClaudeROSSI: ok

src/async-task/tasks/BillTransactionAsyncTask.ts
ClaudeROSSI: I have tried to stick to what we currently have in the other asyncTasks ... they all throw 'Error' directly.
So I'll keep it for now. ok?

src/async-task/tasks/BillTransactionAsyncTask.ts
```diff
@@ -0,0 +1,53 @@
...
+            } finally {
+            // Release the lock
+              await LockingManager.release(transactionLock);
+            }
+          }
+        }
+      } catch (error) {
+        await Logging.logActionExceptionMessage(tenant.id, ServerAction.BILLING_TRANSACTION, error);
+        // TODO - to be clarified - Propagate the error to have the chance to run the task again!
+        throw error;
```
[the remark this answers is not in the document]
ClaudeROSSI: well ... after a short discussion with Serge I think the error propagation is useless. So, I'll remove line 49.
Recovering from billing errors will for sure require a specific procedure which is not yet in place.

jerome-benoit (approved):

PR #2487, opened 2021-05-28
Log when status notification available is received while a transaction is ongoing

jerome-benoit:
And update dependencies to silence TS version support warnings.


ClaudeROSSI (approved):
ok

PR #2497, opened 2021-05-31
billing - transaction projection - add invoice number

ClaudeROSSI:
TransactionStorage.**getTransaction** - Projection has been changed to allow testing whether the invoiceNumber is properly set or not.


jerome-benoit (approved):
Looks good to me.

PR #2498, opened 2021-05-31
[2/2] Adjust visual id migration tags

mahdibenromdhane:


jerome-benoit:
@mahdibenromdhane That version is bug free?

mahdibenromdhane:
> @mahdibenromdhane That version is bug free?

Yes.

src/storage/mongodb/MongoDBStorage.ts
```diff
@@ -105,7 +105,6 @@ export default class MongoDBStorage {
       { fields: { deleted: 1, createdOn: 1 } },
       { fields: { issuer: 1, createdOn: 1 } },
       { fields: { userID: 1, issuer: 1 } },
-      { fields: { visualID: 1 }, options: { unique: true } }
```
jerome-benoit, line 108: The goal is to have an index on that new field and to ensure its uniqueness, right? So you probably have removed that line by mistake.

jerome-benoit (changes requested):

src/migration/tasks/AddVisualIDPropertyToTagsTask.ts
```diff
@@ -19,19 +20,23 @@ export default class AddVisualIDPropertyToTagsTask extends MigrationTask {
...
+            { $set: { visualID: visualID } });
+          updated++;
+        }
       }
-    }
+      skip += Constants.BATCH_PAGE_SIZE;
+    } while (skip < count);
     // Log in the default tenant
     if (updated > 0) {
       await Logging.logDebug({
```
jerome-benoit, line 42: The migration version is not bumped.

jerome-benoit (changes requested):

src/storage/mongodb/MongoDBStorage.ts
mahdibenromdhane: The unique index field need to be added after filling the visualID so we need to merge in 3 steps:

1. Remove unique index
2. fill the visualID
3. Re Add the index

jerome-benoit (approved):

src/migration/tasks/AddVisualIDPropertyToTagsTask.ts
mahdibenromdhane: Fixed

PR #2499, opened 2021-06-01
periodic billing - update Transaction Billing Data

ClaudeROSSI:
The periodic billing operation now updates the billing data of each transaction in order to be able to show the final invoice number when looking at the transaction history.

### DIFF
--- a/src/integration/billing/stripe/StripeBillingIntegration.ts
+++ b/src/integration/billing/stripe/StripeBillingIntegration.ts
@@ -429,7 +429,7 @@ export default class StripeBillingIntegration extends BillingIntegration {
         // Save
         await TransactionStorage.saveTransaction(this.tenantID, transaction);
       } catch (error) {
-        // catch stripe errors and send the information back to the client
+        // Catch stripe errors and send the information back to the client
         await Logging.logError({
           tenantID: this.tenantID,
           action: ServerAction.BILLING_PERFORM_OPERATIONS,
--- a/src/integration/billing/stripe/StripeBillingIntegration.ts
+++ b/src/integration/billing/stripe/StripeBillingIntegration.ts
@@ -417,6 +417,7 @@ export default class StripeBillingIntegration extends BillingIntegration {
     return billingInvoice;
   }
 
+  // TODO - move this method to the billing abstraction to make it common to all billing implementation
   private async _updateTransactionsBillingData(billingInvoice: BillingInvoice): Promise<void> {
     await Promise.all(billingInvoice.sessions.map(async (session) => {
       const transactionID = session.transactionID;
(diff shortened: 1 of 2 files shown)


src/integration/billing/stripe/StripeBillingIntegration.ts
```diff
@@ -20,6 +20,7 @@ import { ServerAction } from '../../../types/Server';
 import Stripe from 'stripe';
 import StripeHelpers from './StripeHelpers';
 import Transaction from '../../../types/Transaction';
+import TransactionStorage from '../../../storage/mongodb/TransactionStorage';
```
ClaudeROSSI, line 23: @LucasBrazi06 : This introcuce a direct dependency between the strip concrete implementation and the TransactionStorage. Is this ok with you?

src/integration/billing/stripe/StripeBillingIntegration.ts
```diff
@@ -410,9 +411,36 @@ export default class StripeBillingIntegration extends BillingIntegration {
     await StripeHelpers.updateInvoiceAdditionalData(this.tenantID, billingInvoice, operationResult);
     // Send a notification to the user
     void this.sendInvoiceNotification(billingInvoice);
+    if (FeatureToggles.isFeatureActive(Feature.BILLING_ASYNC_UPDATE_TRANSACTION)) {
+      await this._updateTransactionsBillingData(billingInvoice);
+    }
     return billingInvoice;
   }
 
+  private async _updateTransactionsBillingData(billingInvoice: BillingInvoice): Promise<void> {
```
ClaudeROSSI, line 420: @LucasBrazi06 : I'm starting some private methods with an "_" to make it clear that this is a somehow a core feature and that the method is supposed to be called  in a specific context (e.g.: a scheduled job or an async task). 
Is this ok?

src/integration/billing/stripe/StripeBillingIntegration.ts
```diff
@@ -410,9 +411,36 @@ export default class StripeBillingIntegration extends BillingIntegration {
...
+      try {
+        const transaction = await TransactionStorage.getTransaction(this.tenantID, Number(transactionID));
+        // Update Billing Data
+        transaction.billingData.stop.invoiceStatus = billingInvoice.status;
+        transaction.billingData.stop.invoiceNumber = billingInvoice.number;
+        transaction.billingData.lastUpdate = new Date();
+        // Save
+        await TransactionStorage.saveTransaction(this.tenantID, transaction);
+      } catch (error) {
+        // catch stripe errors and send the information back to the client
```
jerome-benoit, line 432:
```suggestion
        // Catch stripe errors and send the information back to the client
```

jerome-benoit (approved):

src/integration/billing/stripe/StripeBillingIntegration.ts
LucasBrazi06: I think we already discussed it, right?
Ohterwise, Stripe -> Storage dependency does not hurt me.

src/integration/billing/stripe/StripeBillingIntegration.ts
LucasBrazi06: Usually we avoid the '_' in front of the private method.
Maybe we can talk about it?

PR #2506, opened 2021-06-01
Make AddVisualIDPropertyToTagsTask re-entrant

LucasBrazi06:


jerome-benoit (approved):

src/migration/tasks/AddVisualIDPropertyToTagsTask.ts
```diff
@@ -20,23 +20,24 @@ export default class AddVisualIDPropertyToTagsTask extends MigrationTask {
...
-            { _id: tag.id },
-            { $set: { visualID: visualID } });
+            { _id: tag['_id'] },
+            { $set: { visualID } });
           updated++;
         }
       }
-      skip += Constants.BATCH_PAGE_SIZE;
-    } while (skip < count);
+    } while (tags.length === Constants.BATCH_PAGE_SIZE); // Avoid infinite loop due to issues in the update process
```
jerome-benoit, line 40:
```suggestion
    } while (tags.length <= Constants.BATCH_PAGE_SIZE); // Avoid infinite loop due to issues in the update process
```
At the and of the collection, the tags.length could be less than the requested batch size.

src/migration/tasks/AddVisualIDPropertyToTagsTask.ts
```diff
@@ -20,23 +20,24 @@ export default class AddVisualIDPropertyToTagsTask extends MigrationTask {
...
+      // Get the tags
+      tags = await global.database.getCollection<any>(tenant.id, 'tags')
+        .find({ visualID: { $exists: false } })
+        .limit(Constants.BATCH_PAGE_SIZE)
+        .toArray();
+      if (!Utils.isEmptyArray(tags)) {
         const visualID = new ObjectID().toHexString();
-        for (const tag of tags.result) {
+        for (const tag of tags) {
           await global.database.getCollection<any>(tenant.id, 'tags').update(
```
jerome-benoit, line 34:
```suggestion
          await global.database.getCollection<any>(tenant.id, 'tags').updateOne(
```
Silence a deprecation warning.

jerome-benoit (approved):

src/migration/tasks/AddVisualIDPropertyToTagsTask.ts
LucasBrazi06: > At the end of the collection, tags.length could be less than the requested batch size.

This is a 'while', I tested it and it was working fine with the '==='.
You should revert it or it'll never run on big collection ;-).

PR #2511, opened 2021-06-02
Locking expiration mechanism

OliveGerste:
First approach for an expiration time for locks. 
Unfortunately I have some problems with my unit tests and cannot check if everything works properly. That's why only draft PR.


src/locking/LockingHelper.ts
```diff
@@ -24,9 +16,9 @@ export default class LockingHelper {
     return lock;
   }
 
-  public static async createSiteAreaSmartChargingLock(tenantID: string, siteArea: SiteArea): Promise<Lock | null> {
-    const lock = LockingManager.createExclusiveLock(tenantID, LockEntity.SITE_AREA, `${siteArea.id}-smart-charging`);
-    if (!(await LockingManager.acquire(lock))) {
+  public static async createSiteAreaSmartChargingLock(tenantID: string, siteArea: SiteArea, timeout: number): Promise<Lock | null> {
```
jerome-benoit, line 19:
```suggestion
  public static async createSiteAreaSmartChargingLock(tenantID: string, siteArea: SiteArea, timeoutMs: number = 0): Promise<Lock | null> {
```

src/locking/LockingHelper.ts
```diff
@@ -24,9 +16,9 @@ export default class LockingHelper {
...
+    const lock = LockingManager.createExclusiveLock(tenantID, LockEntity.SITE_AREA, `${siteArea.id}-smart-charging`, 180);
+    if (!(await LockingManager.acquire(lock, timeout))) {
```
jerome-benoit, line 21:
```suggestion
    if (!(await LockingManager.acquire(lock, timeoutMs))) {
```

src/types/Locking.ts
```diff
@@ -6,6 +6,7 @@ export default interface Lock {
   key: string;
   type: LockType;
   timestamp: Date;
+  expirationDate?: Date;
```
jerome-benoit, line 9: Like I said in Serge's PR stacked over yours: relying on unix timestamp here will make things simple and speed.

src/locking/LockingHelper.ts
LucasBrazi06: My PR includes this change already and has to be merged if everybody agrees

src/locking/LockingHelper.ts
LucasBrazi06: My PR has to be merged first

src/types/Locking.ts
LucasBrazi06: Already answered in my PR

src/locking/LockingManager.ts
```diff
@@ -41,63 +41,16 @@ export default class LockingManager {
...
-          break;
-        default:
-          throw new BackendError({
-            action: ServerAction.LOCKING,
-            module: MODULE_NAME, method: 'tryAcquire',
-            message: `Cannot acquire a lock entity '${lock.entity}' ('${lock.key}') with an unknown type '${lock.type}'`,
-            detailedMessages: { lock }
-          });
+      // Check if specific lock for the asset has an expiration date
+      if (retry && await LockingManager.checkAndReleaseExpiredLock(lock)) {
```
OliveGerste, line 48: In my original PR I also checked if the new lock has an expiration date, because normally if the new one does not have an expiration date, the one in the db has none either. 
(So we could reduce data base traffic of the lock collection)
But it should be fine like this also.

LucasBrazi06: > In my original PR I also checked if the new lock has an expiration date, because normally if the new one does not have an expiration date, the one in the db has none either.
> (So we could reduce data base traffic of the lock collection)
> But it should be fine like this also.

I wouldn't do this assumption :-).
What matters is if the lock in the DB has an expiration date to remove it.
No correlation between the two.
Btw, do we have to put in place the Job?

OliveGerste: Okay. 
Normally the job is not needed if we put an expiration date on every lock type. 
If we want to have locks without an expiration date, we would have to discuss when the locks should be released by the job.

LucasBrazi06: > Okay.
> Normally the job is not needed if we put an expiration date on every lock type.
> If we want to have locks without an expiration date, we would have to discuss when the locks should be released by the job.

No, it's just to remove expired locks if you do not request them anymore and then have a clean 'locks' collection, but that's okay for the time being :-).

jerome-benoit: Not allowing lock without cleanup once the expiration date is reached is risky for some OCPI related tasks.

LucasBrazi06: > Not allowing lock without cleanup once the expiration date is reached is risky for some OCPI related tasks.

It is released when requested again, I don't see the problem here!

LucasBrazi06 (approved):

src/locking/LockingManager.ts
jerome-benoit: It will allow to run the same heavy task twice, which defeats all the locking primary purpose. 
For the pull tokens task, it really might happen if the scheduled task interval is higher than the default lock expiration date.

LucasBrazi06: > It will allow to run the same heavy task twice, which defeats all the locking primary purpose.
> For the pull tokens task, it really might happen if the scheduled task interval is higher than the default lock expiration date.

I still don't get it why it's a problem!
Releasing an expired lock is an exceptional use case that should never happen.

jerome-benoit: Locking per design are meant to avoid concurrent execution. The introduced workaround to remove stalled locks must be used with caution when use to protect against concurrent execution of long running and heavy process.  
Since it's a workaround but a one that defeats the primary locking design, it probably should not be enabled by default at a low level and only inside the locking helper to avoid human error and ensure the developer build the helper correctly for its target.

LucasBrazi06: It's ok as it is.

PR #2518, opened 2021-06-04
Remove setTimeout in lock manager

LucasBrazi06:


jerome-benoit:
`setTimout() `is a lot more optimised than doing a busy loop calling gettimeofday syscall in its condition: the event loop is stuck by its code execution. 

The existing construct is the documented way node.js event loop friendly to avoid to stuck the event loop implementing properly a busy wait loop. 

So I would really prefer that you do not merge that change, other wise each lock invocation with a acquire timeout will block the event loop.

LucasBrazi06:
> `setTimout() `is a lot more optimised than doing a busy loop calling gettimeofday syscall in its condition: the event loop is stuck by its code execution.
> 
> The existing construct is the documented way node.js event loop friendly avoiding to stuck the event loop and implementing properly a busy wait loop.
> 
> So I would really prefer that you do not merge that change, otherwise each lock invocation with an acquire timeout will block the event loop.

The rest of the code works has before, same constraint, no more impact on the event loop compared to the old code where you still wait before retrying.
The setTimeout was added on top just to set a flag, just removed it -> simplification.
Moreover setTimeout is inaccurate under heavy load.
So I don't see your point here.

jerome-benoit:
> The rest of the code works has before, same constraint, no more impact on the event loop compared to the old code where you still wait before retrying.
> The setTimeout was added on top just to set a flag, just removed it -> simplification.
> Moreover setTimeout is inaccurate under heavy load.
> So I don't see your point here.

You introduce a call to a syscall on a while loop condition on a code path that must be as light as possible, the event loop must handle them all + the check return boolean instead of a light boolean check. Under heavy load, the impact of that kind of heavy busy loop on the runtime is huge. 
The previous implementation does not have that issue under heavy load. 

Locking primitive must be as light as possible. Their inherent and internal complexity to reach that goal is not issue since it is hidden. 

If you really want to remove the setTimeout() for whatever reason, I think you have to find a mechanism as light.

LucasBrazi06:
> If you really want to remove the setTimeout() for whatever reason, I think you have to find a mechanism as light.

The current implementation in this PR is satisfactory for me.
I'm not looking for optimal perfs here.

jerome-benoit:
> > If you really want to remove the setTimeout() for whatever reason, I think you have to find a mechanism as light.
> 
> The current implementation in this PR is satisfactory for me.
> I'm not looking for optimal perfs here.

I do not see the point of replacing an optimized and scalable implementation in a hot code path by a not optimized one (and an utterly not optimized one that will do millions of useless calls and branching) that will not scale by design. 

Finding such a change satisfactory is beyond my comprehension and the simplification tradeoff does come into the game for 4 lines of code given the huge degradation introduced.

OliveGerste (commented):
The code seems to work, but I do not know if it leads to performance issues, when the while loop is blocking the event loop. I would leave this up to you and Jerome.

LucasBrazi06:
> The code seems to work, but I do not know if it leads to performance issues, when the while loop is blocking the event loop. I would leave this up to you and Jerome.

Thanks Tobias :-)

PR #2520, opened 2021-06-04
Billing - test data cleanup

ClaudeROSSI:
The purpose of these changes is to make it possible for the customer to experiment the stripe integration with a stripe test account.
As a consequence, we need to be able to clear all test data once the customer goes live.

new methods have been added to:
-  delete all invoices with **liveMode** set to false and clear the corresponding billingData aattched to the corresponding transactions.
- clear all users having a billingData with liveMode set to false. The purpose here is to get rid of all STRIPE customer references.

So far, no endpoints are using the new methods.
However, utBilling tests can be used to check the behavior!

### DIFF
--- a/src/integration/billing/BillingIntegration.ts
+++ b/src/integration/billing/BillingIntegration.ts
@@ -1,4 +1,4 @@
-import { BillingChargeInvoiceAction, BillingDataTransactionStart, BillingDataTransactionStop, BillingDataTransactionUpdate, BillingInvoice, BillingInvoiceItem, BillingInvoiceStatus, BillingOperationResult, BillingPaymentMethod, BillingTax, BillingUser, BillingUserSynchronizeAction } from '../../types/Billing';
+import { BillingChargeInvoiceAction, BillingDataTransactionStart, BillingDataTransactionStop, BillingDataTransactionUpdate, BillingInvoice, BillingInvoiceItem, BillingInvoiceStatus, BillingOperationResult, BillingPaymentMethod, BillingStatus, BillingTax, BillingUser, BillingUserSynchronizeAction } from '../../types/Billing';
 import FeatureToggles, { Feature } from '../../utils/FeatureToggles';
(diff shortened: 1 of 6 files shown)


src/storage/mongodb/BillingStorage.ts
```diff
@@ -26,6 +26,12 @@ export default class BillingStorage {
     return invoicesMDB.count === 1 ? invoicesMDB.result[0] : null;
   }
 
+  public static async getInvoicesInTestMode(tenantID: string): Promise<DataResult<BillingInvoice>> {
+    // Returns all invoices that are to be finalized and paid
+    const invoicesMDB = await BillingStorage.getInvoices(tenantID, { liveMode: false }, Constants.DB_PARAMS_MAX_LIMIT);
```
LucasBrazi06, line 31: No need an intermediate var:
`return BillingStorage.getInvoices(tenantID, { liveMode: false }, Constants.DB_PARAMS_MAX_LIMIT);`

src/integration/billing/BillingIntegration.ts
```diff
@@ -359,6 +362,163 @@ export default abstract class BillingIntegration {
...
+        statuses: [UserStatus.ACTIVE],
+        withTestBillingData: true
+      }, Constants.DB_PARAMS_MAX_LIMIT);
+    if (users.count > 0) {
+      return users.result;
+    }
+    return [];
+  }
+
+  private async _clearUserTestBillingData(user: User): Promise<void> {
```
LucasBrazi06, line 508: '_' to remove to keep project consistency?

src/storage/mongodb/BillingStorage.ts
LucasBrazi06: You can also call right away the BillingStorage.getInvoices(...), you will not create a new method whenever you have a filter to call in the main one? :-)

LucasBrazi06 (approved):
Just check my non blocking feedback

PR #2522, opened 2021-06-07
Handle Transaction.End meter value outside the Stop Transaction + Do not stop the Transaction in Status Notification

LucasBrazi06:
Handle Transaction.End meter value outside the Stop Transaction
Do not automatically stop the ongoing Transaction when Status Notification changes from Charging to Available
Code cleanup


package.json
```diff
@@ -59,8 +59,10 @@
     "eslint": "cross-env TIMING=1 eslint --ext .js,.ts build/types src test",
     "eslint:fix": "cross-env TIMING=1 eslint --ext .js,.ts --fix build/types src test",
     "autocannon": "autocannon",
-    "mochatest": "cross-env TS_NODE_FILES=true nyc mocha test/**/*Test.ts",
-    "mochatest:no-parallel": "cross-env TS_NODE_FILES=true nyc mocha --no-parallel test/**/*Test.ts",
+    "mochatest": "cross-env TS_NODE_FILES=true mocha --no-parallel test/**/*Test.ts",
+    "mochatest:parallel": "cross-env TS_NODE_FILES=true mocha test/**/*Test.ts",
+    "mochatest:nyc": "cross-env TS_NODE_FILES=true nyc mocha --no-parallel test/**/*Test.ts",
+    "mochatest:nyc-parallel": "cross-env TS_NODE_FILES=true nyc mocha test/**/*Test.ts",
```
jerome-benoit, line 65: The default setting must be parallel to avoid time wasting (50% supplemental time without parallel) for developers and GitHub actions. 
And the de facto standard for scripts naming is: `command:cli-options`.
If debug is needed, the explicit no-parallel scripts are here for that reason.

jerome-benoit (changes requested):
How will be handled the stalled running transactions in case of not receiving a stop transaction? (because failures or bugs on charging stations)

LucasBrazi06:
> How will be handled the stalled running transactions in case of not receiving a stop transaction? (because failures or bugs on charging stations)

Indeed!
We have 2 solutions here: The Soft Stop (manual in the UI) and a Start Transaction that will stop the ongoing one.
Here we are still dealing with an unusual use case that should not happen.

src/server/ocpp/services/OCPPService.ts
```diff
@@ -90,214 +80,42 @@ export default class OCPPService {
...
-            chargingStation.siteAreaID = token.siteAreaID;
-            chargingStation.siteID = siteArea.siteID;
-            // Set the same coordinates
-            if (siteArea?.address?.coordinates?.length === 2) {
-              chargingStation.coordinates = siteArea.address.coordinates;
-            }
-          }
-        }
+        // Create Charging Station
+        chargingStation = await this.checkAndRegiterNewChargingStation(tenant, bootNotification, headers);
```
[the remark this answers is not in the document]
LucasBrazi06: Fixed

package.json
LucasBrazi06: The parallel execution triggers a lot of failures in the unit tests on my side.
Putting this option by default is longer but the unit tests are all passing successfully which I prefer then.
And you know what I think about the standard stuff already ;-)

jerome-benoit: > The parallel execution triggers a lot of failures in the unit tests on my side.
> Putting this option by default is longer but the unit tests are all passing successfully which I prefer then.

I think you have an issue with your setup, the parallel mode since its enablement is working well for all developers and in the GitHub action. 
If want to revert it properly, you have to: 
+ set parallel to false in the mocha configuration file, 
+ use the --parallel option on moche in scripts requiring it,
+ run the test script with parallel mode in the GitHub action. 

> And you know what I think about the standard stuff already ;-)

De facto standard are made by developers for developers and here it's just common sense: calling a command with one option but the naming scheme state the contrary ...

LucasBrazi06: I think Madhi has also issues on running cars unit tests (random.)
I'm not the only one here, I'm pretty sure.
And I'm having these random issues since a long time, just found the root cause today.

jerome-benoit:
> > How will be handled the stalled running transactions in case of not receiving a stop transaction? (because failures or bugs on charging stations)
> 
> Indeed!
> We have 2 solutions here: The Soft Stop (manual in the UI) and a Start Transaction that will stop the ongoing one.

A scheduler task can also check for running transactions and if one transaction: 
+ is attached to a charging station online with the last status notification set "Available';
and
+ have not received a meter values since ... one hour?

a soft stop is then triggered. 

We can't anything smart on offline charging stations because they should buffer in case of not reaching the backend, taking a decision in that case can end a transaction that will be ended properly when the charging station comes back online.

LucasBrazi06:
> A scheduler task can also check for running transactions and if one transaction:
> 
> * is attached to a charging station online with the last status notification set to "Available';
>   and
> * have not received a meter values since ... one hour?
> 
> a soft stop is then triggered.
> 
> We can't do anything smart on offline charging stations because they should buffer in case of not reaching the backend, taking a decision in that case can end a transaction that will be ended properly when the charging station comes back online.

Yes indeed, the clean up job is a good idea!
For the time being we can live with the counter measures in place, time is money ;-).

PR #2524, opened 2021-06-07
REST > Added user getDefaultCarTag / getSites / AssignSites endpoint + doc

Ninjeneer:


test/api/SiteOrgTest.ts
```diff
@@ -65,6 +65,16 @@ async function assignUserToSite(userRole, site): Promise<any> {
   return testData.userService.siteApi.addUsersToSite(site.id, [userContext.id]);
 }
 
+/**
+ * @param userRole
+ * @param sites
+ */
```
LucasBrazi06, line 71: No need comments for method

LucasBrazi06 (approved):

test/api/SiteOrgTest.ts
Ninjeneer: This is an ESLint auto generated JSDoc. Should we remove this rule ?

LucasBrazi06: > This is an ESLint auto generated JSDoc. Should we remove this rule ?

Yes, I think so.

PR #2473, opened 2021-05-26
Car authorization concept refactoring

roxiciobotea210:
Refactor authorization concept for Car entity.

PR #2531, opened 2021-06-08
Check authorized flag everywhere

roxiciobotea210:
Check authorized flag for all refactored method calls (as Mehdi wonderfully did for canListUsers scanario :) ).

(Looks like the reason I was not able to push commits was because I was not in the github team anymore. If possible, could I stay a team member until [this](https://github.com/sap-labs-france/ev-server/pull/2473) PR gets merged? Just in case it will need adjustments. Thanks!)


LucasBrazi06 (approved):
Perfect!
Big Thanks!
Yes, I'll keep you in the team ;-)

PR #2534, opened 2021-06-09
fix chargex tags import

cabch:
modified visual id header
test for name, firstname and email or put ''

### DIFF
--- a/src/server/rest/v1/service/TagService.ts
+++ b/src/server/rest/v1/service/TagService.ts
@@ -702,7 +702,7 @@ export default class TagService {
     if (writeHeader) {
       headers = [
         'id',
-        'visual ID',
+        'visualID',
         'description',
         'firstName',
         'name',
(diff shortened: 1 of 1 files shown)


src/server/rest/v1/service/TagService.ts
```diff
@@ -775,9 +775,9 @@ export default class TagService {
         id: importedTag.id.toUpperCase(),
         visualID: importedTag.visualID,
         description: importedTag.description ? importedTag.description : `Badge ID '${importedTag.id}'`,
-        name: importedTag.name.toUpperCase(),
-        firstName: importedTag.firstName,
-        email: importedTag.email,
+        name: importedTag.name?.toUpperCase() || '',
```
LucasBrazi06, line 778: Maybe user props should be set only if there are a name, firstName and email present and all mandatory, so we don't have case where name is present and not firstName and we avoid to have also name: null.
What do you think?

cabch: I agree, I made the modifications

LucasBrazi06 (approved):

PR #2537, opened 2021-06-09
Billing optimize periodic billing

ClaudeROSSI:
Periodic Billing:

- index has been added to make sure we get a stable bevaviour when the number of invoices increases.
> {  fields: { createdOn: 1 } }

- pagination has been added to make sure we do not load all invoices in memory
- same filtering logic applies when testing but on the current day (instead of the previous month)
- unused convenience methods have been removed from BillingStorage
- tests have been adapted a bit

### DIFF
--- a/src/integration/billing/BillingIntegration.ts
+++ b/src/integration/billing/BillingIntegration.ts
@@ -537,8 +537,8 @@ export default abstract class BillingIntegration {
 
   private preparePeriodicBillingQueryParameters(forceOperation: boolean): { limit: number, sort: Record<string, unknown>, filter: Record<string, unknown> } {
     // Prepare filtering to process Invoices of the previous month
-    const startDateTime = moment().date(0).date(1).toDate();
-    const endDateTime = moment().date(1).toDate();
+    const startDateTime = moment().date(0).date(1).startOf('day').toDate(); // 1st day of the previous month 00:00:00 (AM)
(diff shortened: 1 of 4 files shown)


src/storage/mongodb/BillingStorage.ts
```diff
@@ -135,7 +113,7 @@ export default class BillingStorage {
     aggregation.pop();
     // Sort
     if (!dbParams.sort) {
-      dbParams.sort = { name: 1 };
+      dbParams.sort = { _id: 1 };
```
ClaudeROSSI, line 116: I do not know why we had 'name' here in the previous version. I assumed this was a bug!

src/integration/billing/BillingIntegration.ts
```diff
@@ -391,7 +397,7 @@ export default abstract class BillingIntegration {
   }
 
   private async _clearAllInvoiceTestData(): Promise<void> {
-    const invoices: DataResult<BillingInvoice> = await BillingStorage.getInvoicesInTestMode(this.tenantID);
+    const invoices: DataResult<BillingInvoice> = await BillingStorage.getInvoices(this.tenantID, { liveMode: false }, Constants.DB_PARAMS_MAX_LIMIT);
```
ClaudeROSSI, line 400: out of scope - but the getInvoicesInTestMode() helper method has been removed

LucasBrazi06 (approved):

PR #2546, opened 2021-06-11
Fixed user Basic cannot read his own data in Car

mahdibenromdhane:


src/server/rest/v1/service/CarService.ts
```diff
@@ -582,7 +582,7 @@ export default class CarService {
         UtilsService.assertObjectExists(action, foundUser, `User ID '${userToCheck.user.id}' does not exist`,
           MODULE_NAME, 'handleAssignCarUsers', loggedUser);
         // Auth
-        if (!(await Authorizations.canReadUser(loggedUser)).authorized) {
+        if (!(await Authorizations.canReadUser(loggedUser, { UserID: loggedUser.id })).authorized) {
```
LucasBrazi06, line 585: Just FI, this will not be the final solution

LucasBrazi06 (approved):

src/server/rest/v1/service/CarService.ts
mahdibenromdhane: Yes, I will create an issue.

PR #2554, opened 2021-06-14
Refactored notify end of charge

LucasBrazi06:


src/server/ocpp/services/OCPPService.ts
```diff
@@ -1135,31 +1135,36 @@ export default class OCPPService {
...
+        const connector = Utils.getConnectorFromID(chargingStation, transaction.connectorId);
+        if (connector.status === ChargePointStatus.CHARGING) {
+          // Check the last 3 consumptions
+          const consumptions = await ConsumptionStorage.getTransactionConsumptions(
+            tenant.id, { transactionId: transaction.id }, { limit: 3, skip: 0, sort: { startedAt: -1 } });
+          if (consumptions.count === 3) {
+            // Check the consumptions
+            const noConsumption = consumptions.result.every((consumption) =>
+              consumption.consumptionWh === 0 &&
+              consumption.limitSource !== ConnectorCurrentLimitSource.CHARGING_PROFILE);
```
OliveGerste, line 1162: I think we are missing a check here. Normally we have a charging profile in place when the car is fully charged. 
So let us say we have a charging profile in place with a limit of 11kW, but the car is not charging anymore. 
The consumptionWh, will be 0, but the limitSource will be the charging profile so no notification is sent, even if the car finished charging.

src/server/ocpp/services/OCPPService.ts
```diff
@@ -1135,31 +1135,36 @@ export default class OCPPService {
...
             // Send Notification
-            this.notifyEndOfCharge(tenant, chargingStation, transaction);
-          } else {
-            // Check last consumptions
-            const consumptions = await ConsumptionStorage.getTransactionConsumptions(
-              tenant.id, { transactionId: transaction.id }, { limit: 3, skip: 0, sort: { startedAt: -1 } });
-            if (consumptions.result.every((consumption) => consumption.consumptionWh === 0 &&
-                (consumption.limitSource !== ConnectorCurrentLimitSource.CHARGING_PROFILE ||
-                consumption.limitAmps >= StaticLimitAmps.MIN_LIMIT_PER_PHASE * Utils.getNumberOfConnectedPhases(chargingStation, null, transaction.connectorId)))) {
```
OliveGerste, line 1153: This or condition checked, if there is a cp in place and if this limit is above the min limit. If it is above, we send the end of charge notif.

src/server/ocpp/services/OCPPService.ts
```diff
@@ -1135,31 +1135,36 @@ export default class OCPPService {
...
+        // Check if optimal charge has been reached (85%)
+        } else if (transaction.currentStateOfCharge >= this.chargingStationConfig.notifBeforeEndOfChargePercent) {
+          // Send Notification
+          this.notifyOptimalChargeReached(tenant, chargingStation, transaction);
+        }
+      // No battery information: check last consumptions
+      } else {
+        // Connector status must be 'Charging'
+        const connector = Utils.getConnectorFromID(chargingStation, transaction.connectorId);
+        if (connector.status === ChargePointStatus.CHARGING) {
```
OliveGerste, line 1154: If the car is not consuming energy the connector status changes to SuspendedEV right? So I think this would need to be added here.

LucasBrazi06: What about checking SuspendedEV instead of Charging?

OliveGerste: In theory I would say this works. 
But sometimes the state of the connector also switches to SuspendedEVSE.
E.G.:
image
image

I am not completely sure why it happens. Maybe the car is not able to charge at the min limit. Or it is a similar issue like with the Tesla Model 3, where it had something to do with mode 3 and the ocpp parameters.

LucasBrazi06: Ok, perfect, I just provided an enhancement in the code.
Let me know.

OliveGerste: I think we need status SuspendedEV here like you mentioned, because normally the charge is suspended by the EV and not by the EVSE, when the EV is fully charged.

I just wanted to mention, that we also have other cases, where the status says the EVSE suspends the charge. Sorry for the misunderstanding.

LucasBrazi06: No, you're right,  my bad! :-)
I just added the SuspendedEV status.
Let me know.

OliveGerste: Looks good to me know 👍

src/server/ocpp/services/OCPPService.ts
```diff
@@ -1135,31 +1135,38 @@ export default class OCPPService {
...
+          // Send Notification
+          this.notifyEndOfCharge(tenant, chargingStation, transaction);
+        // Check if optimal charge has been reached (85%)
+        } else if (transaction.currentStateOfCharge >= this.chargingStationConfig.notifBeforeEndOfChargePercent) {
+          // Send Notification
+          this.notifyOptimalChargeReached(tenant, chargingStation, transaction);
+        }
+      // No battery information: check last consumptions
+      } else {
+        // Connector' status must be 'Supended'
```
OliveGerste, line 1152: Small typo 'Supended'

OliveGerste (approved):

src/server/ocpp/services/OCPPService.ts
LucasBrazi06: Fixed :-)

PR #2571, opened 2021-06-17
HOTFIX - periodic billing - refresh data required to get the latest amount

ClaudeROSSI:
minor fix for periodic billing.

The invoice amount shown (on our side) was not the latest one!


ClaudeROSSI:
@LucasBrazi06, @jerome-benoit : The utbilling tests are failing when run in the pipeline. Do we have them running in parallel now?

Locally everything is fine when forcing the mocha explorer settings like mentioned in our team channel.

PR #2573, opened 2021-06-17
Start transaction - check payment method

ClaudeROSSI:
Preparing the server in order to make it possible to prevent a start transaction when the billing prerequisites are not met.

**New enum** 

> enum **StartTransactionCheck** {
>   OK = 'ok',
>   BILLING_NO_PAYMENT_METHOD = 'no_payment_method',
>   BILLING_NO_TAX = 'billing_no_tax', 
>   BILLING_NO_SETTINGS = 'billing_no_settings', 
>   BILLING_INCONSISTENT_SETTINGS = 'billing_inconsistent_settings',
> }

The UI will enable the start transaction button only when the check status is set to **StartTransactionCheck.OK**

The status is send together with the Tag and the Car when using the endpoint **getUserDefaultTagCar** endpoint.


src/server/rest/v1/service/UserService.ts
```diff
@@ -110,8 +111,27 @@ export default class UserService {
...
+          action: action,
+          message: `Start Transaction checks failed for ${user.id}`,
+          detailedMessages: { error: error.message, stack: error.stack }
+        });
+        status = StartTransactionCheck.BILLING_INCONSISTENT_SETTINGS;
+      }
+    }
     // Return
-    res.json({ tag, car });
+    res.json({ tag, car, status });
```
LucasBrazi06, line 134: As discussed, return an array of enum

LucasBrazi06 (changes requested):

ClaudeROSSI:
@LucasBrazi06 : Changes have been done

 
image

LucasBrazi06 (approved):

PR #2577, opened 2021-06-18
billing - invoice - normalize currency

ClaudeROSSI:
just making sure the currency format on our side matches the one defined in the pricing settings.
i.e. : uppercase on our side (while lowercase in stripe invoices)


cabch (approved):
ok for me

PR #2592, opened 2021-06-23
Billing settings - live mode property

ClaudeROSSI:
billing settings - now provide liveMode boolean property


src/server/rest/v1/service/BillingService.ts
```diff
@@ -846,9 +846,14 @@ export default class BillingService {
         module: MODULE_NAME, method: 'handleGetBillingSetting',
       });
     }
-    const billingSettings: BillingSettings = await SettingStorage.getBillingSetting(req.user.tenantID);
+    const billingSettings = await SettingStorage.getBillingSetting(req.user.tenantID);
     UtilsService.assertObjectExists(action, billingSettings, 'Failed to load billing settings', MODULE_NAME, 'handleGetBillingSetting', req.user);
     UtilsService.hashSensitiveData(req.user.tenantID, billingSettings);
+    try {
```
LucasBrazi06, line 852: You can put the try...catch in the BillingService.isConnectedToLiveAccount(req) and return false in the catch?

ClaudeROSSI: Actually when there is an error I do not know if it's a live account or not.
Putting false (while it might be true) is somehow dangerous ... 
I'll change it to make it clear!

ClaudeROSSI: done ... but I'm not sure you will accept my solution!
We have 3 possible values:
- liveMode: true
- liveMode: false
live mode not set (undefined)

The 3rd one denotes an abnormal situation ... the UI may check it to switch off some features as well.

ClaudeROSSI:
This PR is useless for now.
I'll find a better solution and do it again!

PR #2600, opened 2021-06-25
HOTFIX - async task manager - running same task several times!!!

ClaudeROSSI:
Very tricky issue due to Promise.map!

When **n** async tasks were in a PENDING state. The async task manager was running **n** times the lastest one!
This leading to an exception when the task tries to acquire an exclusive lock.


ClaudeROSSI:
@LucasBrazi06 : Please call me to discuss that one!

LucasBrazi06 (approved):

LucasBrazi06:
> @LucasBrazi06 : Please call me to discuss that one!

I got it, the fucking switch...case + var scope issue in JS :-(

LucasBrazi06:
@jerome-benoit Is there a way to highlight this issue with an eslint rule?
Thanks!

jerome-benoit:
Probably https://www.npmjs.com/package/eslint-plugin-no-closure.

PR #2618, opened 2021-06-30
tenant creation endpoint - schema validation accepting the ID

ClaudeROSSI:
tenant created by the unit tests had a different id than the expected one.

This was due to the validator of the create tenant endpoint.

### DIFF
--- a/src/assets/server/rest/v1/schemas/tenant/tenant-create.json
+++ b/src/assets/server/rest/v1/schemas/tenant/tenant-create.json
@@ -2,6 +2,9 @@
   "title": "Create one Tenant",
   "type": "object",
   "properties": {
+    "id": {
+      "$ref": "common.json#/definitions/id"
+    },
     "name": {
       "$ref": "tenant.json#/definitions/name"
     },


ClaudeROSSI:
@Ninjeneer : I did some tests ... regular tenant creation still works properly when done by a super admin.
Do you see any reason why the "tenant creation endpoint" should not accept the ID as input?
Thanks,
Claude

ClaudeROSSI:
uttest tenants have now the expected ID.

LucasBrazi06 (approved):

Ninjeneer:
> @Ninjeneer : I did some tests ... regular tenant creation still works properly when done by a super admin.
> Do you see any reason why the "tenant creation endpoint" should not accept the ID as input?

You should have an error message returned : either in the body of the response, either in your super-tenant logs. You can also check by adding a breakpoint on the `throw` instruction in SchemaValidator.ts and inspect the 'errors' variable

First of all, is the id passed in the test formatted as a 24 characters long hexadecimal number ?

PR #2619, opened 2021-06-30
rest endpoint - url - encode params

ClaudeROSSI:
Removed dependency to safe-url-assembler

- The URL generation used to test REST endpoints has been adapted
- URL path is properly encoded using **encodeUriComponent** method.


ClaudeROSSI:
sorry - wrong branch

PR #2621, opened 2021-06-30
Change tenant id to object site storage

RohanAntony:


RohanAntony:
reopening another PR instead

PR #2623, opened 2021-06-30
Validation : del-many-tags endpoint + del-one-tag validation + Updated Postman collection

Ninjeneer:
- Added delete many tags endpoint + doc
- Single tag delete validation
- Removed tag security class
- Updated tags postman collection


LucasBrazi06:
@Ninjeneer We will use the Tag visualID as a key in Tag service endpoint instead the Tag ID. You can check with Mahdi to be in sync.
Thanks!

PR #2628, opened 2021-06-30
Fix issuer bug

mahdibenromdhane:


src/server/rest/v1/service/UserService.ts
```diff
@@ -773,7 +773,7 @@ export default class UserService {
     const users = await UserStorage.getUsers(req.user.tenantID,
       {
         search: filteredRequest.Search,
-        issuer: filteredRequest.Issuer,
+        issuer: Utils.isBoolean(filteredRequest.Issuer) || filteredRequest.Issuer ? Utils.convertToBoolean(filteredRequest.Issuer) : null,
```
LucasBrazi06, line 776: Do we have the same issue in other endpoint?

PR #2634, opened 2021-07-02
Billing - periodic operation -  some fixes

ClaudeROSSI:
For a DRAFT invoice, the final invoice date is only known when it gets finalized. The invoice synchronization takes now the finalized_at property into consideration.

The side effect of that change is that the initial date gets updated during the periodic billing process. This as an impact on the pagination logic in place when going thru the list of invoices of the previous month. The pagination logic has been updated to make sure the periodic billing process does not 'forget' any invoices !

### DIFF
--- a/src/integration/billing/BillingIntegration.ts
+++ b/src/integration/billing/BillingIntegration.ts
@@ -195,7 +195,7 @@ export default abstract class BillingIntegration {
       for (const invoice of invoices.result) {
         try {
           // Skip invoices that are already PAID or not relevant for the current billing process
-          if (this.invoiceMustBeSkipped(invoice)) {
+          if (this.isInvoiceOutOfPeriodicOperationScope(invoice)) {
             continue;
           }
           // Make sure to avoid trying to charge it again too soon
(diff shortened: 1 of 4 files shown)


ClaudeROSSI:
The billing tests are failing with HTTP Status Code 595 - User account Changed!
I do not have the problem when running the tests locally!?!

@LucasBrazi06 :
I'm changing the user locale while running the Billing tests. Is this forbidden?
Thanks,
Calude

The check of the user hash is not new!. Rigth
?
    // Check User's Hash
      if (userHashID !== this.buildUserHashID(user)) {
        throw new AppError({
          source: Constants.CENTRAL_SERVER,
          errorCode: HTTPError.USER_ACCOUNT_CHANGED,
          message: 'User has been updated and will be logged off',
          module: MODULE_NAME,
          method: 'isSessionHashUpdated',
          user: req.user
        });
      }

LucasBrazi06:
> @LucasBrazi06 :
> I'm changing the user locale while running the Billing tests. Is this forbidden?

You'll have to relog in that case and get the new token ;-)

ClaudeROSSI:
> > @LucasBrazi06 :
> > I'm changing the user locale while running the Billing tests. Is this forbidden?
> 
> You'll have to relog in that case and get the new token ;-)

Ok ... i'll remove these locale tests for now. I do not know how to 'relog' with the test helpers that we have so far!

ClaudeROSSI:
Tests are still failing but not the ones for the billing!

 1) Car
       Success cases
         With component Car (utcar)
           Where admin user
             Should be able to get car catalogs:
     TypeError: Cannot read property 'id' of undefined
      at Context.<anonymous> (/home/runner/work/ev-server/ev-server/test/api/CarTest.ts:94:43)

I'm lost!

PR #2636, opened 2021-07-02
Smart charging Test excluding stations

OliveGerste:
In some tests generated logs are used to check certain functionalities


LucasBrazi06:
@OliveGerste PR in Draft but in To Be Reviewed pipeline?

OliveGerste:
> @OliveGerste PR in Draft but in To Be Reviewed pipeline?

Hi @LucasBrazi06, I would need a review on the new approach. (For excluding charging stations and testing with generated logs)

Especially if you can think of another case, where a charging station needs to be excluded. Besides of the cases implemented. Should we also take into account the charging station capabilities?

Drafts need to be put in dev cross for a review then right?

LucasBrazi06:
> Drafts need to be put in dev cross for a review then right?

Yes and you assign someone to review your PR :-).
Thanks!

PR #2666, opened 2021-07-15
REST > new routes for ChargingStation and Users QueryString filters

Ninjeneer:
As renaming parameters will lead again to a retro-compatibility issue, I implemented additional route to handle this problem


LucasBrazi06:
@Ninjeneer A big PR has been merged with the Tag Visual ID that impacted the OAS file + some endpoints.
We also now try to pass all the query params named WithXXX (WithUser, WithTag...) to optimize lookups in MongoDB and to not return sensitive data when not requested.
In the end there will be no hardcoded WithXXX param in the Service classes.
Keep that in mind when resolving the conflicts.
I also updated the Mobile App accordingly.

LucasBrazi06 (changes requested):
Why introducing a V2 here?

Ninjeneer:
> 
> 
> Why introducing a V2 here?

Because I changed the request format to match the new qurey string parameter "Status". 
Query for users based on their status (Accepted, Locked...) is now used by a new parameter "UserStatus" and "Status" is now used for users in error. (`/users?UserStatus=A` // `/users?Status=in-error`)

This request format change will cause once again a retro-compatibility issue if I apply it to the V1 route

LucasBrazi06:
> (`/users?UserStatus=A` // `/users?Status=in-error`)
> 
> This request format change will cause once again a retro-compatibility issue if I apply it to the V1 route

Okay thanks!
What if we keep the Status as it is and we add UserStatus=in-error?
The V1 is not yet complete, I would like to avoid a V2 at all cost.
Let me know.

Ninjeneer:
> What if we keep the Status as it is and we add UserStatus=in-error?

This solution will work, but this will break a design logic : Users, Charging Stations and Transactions are sharing a same status : `in-error`, so it would be more logic to keep for all them the same QueryParam name (Status) to match this state, and then use specific QueryParams for entity related states. What do you think ?

LucasBrazi06:
> This solution will work, but this will break a design logic : Users, Charging Stations and Transactions are sharing a same status : `in-error`, so it would be more logic to keep for all them the same QueryParam name (Status) to match this state, and then use specific QueryParams for entity related states. What do you think ?

In that we can also keep the status in the URL as it is today to avoid the V2?

Ninjeneer:
Yes we can. 
Then, for future endpoints, should I keep the status in the URL to match existing Users and ChargingStation routes, or should I use query strings ? The latter won't cause any retro-compatibility issue here as there is no overlap on 'Status' param like Users

PR #2689, opened 2021-07-27
Add consumption production limitation to site area chart

RohanAntony:
Adding changes in mongodb pipeline to fetch cumulated data in a single object instead of as multiple arrays.


RohanAntony:
Opened a second PR with proper and merged changes

PR #2693, opened 2021-07-28
Added CZ language

MichalAnthis:
@ClaudeROSSI 
Hello, 

correctly named and validated json file.


ClaudeROSSI (approved):
some text are missing - we will fix it in another PR.
We also need to declare that the cz locale is now supported

PR #2704, opened 2021-07-29
Data transfer   back

rmangin-fr:
Do not merge this !
Should I copy OCPPDataTransferStatus stuff into OCPPClient ?


src/client/ocpi/OCPIChargingStationClient.ts
```diff
@@ -1,4 +1,5 @@
...
+import { OCPPDataTransferRequest, OCPPDataTransferResponse } from '../../types/ocpp/OCPPServer';
```
rmangin-fr, line 2: @LucasBrazi06: I need your advise, I'm not sure if I should copy definitions into the Client file ?

ClaudeROSSI: as discussed, please mimic what is done for other the commands. Dedicated interfaces have to be defined

ClaudeROSSI (changes requested):
as discussed, please mimic the logic of other commands

rmangin-fr:
Need to test on QA

LucasBrazi06 (approved):
GOOD JOB!!!

PR #2708, opened 2021-07-30
added limit to schema validator

cabch:
limit was excluded from the request so set to 100 by default but the constant for EXPORT_PAGE_SIZE was always (as it has to be) to 1000

image

So at the second loop in exportToCsv method, at the end, skip was set to 2000 even if we retrieved only 200 from db

### DIFF
--- a/src/assets/server/rest/v1/schemas/tag/tags-get.json
+++ b/src/assets/server/rest/v1/schemas/tag/tags-get.json
@@ -22,6 +22,9 @@
     "SortFields": {
       "$ref": "common.json#/definitions/sortFields"
     },
+    "Limit": {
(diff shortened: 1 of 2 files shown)


Ninjeneer (approved):

ClaudeROSSI:
I'm a bit surprised by the fix.
Why does the code rely on the EXPORT_PAGE_SIZE constants while the size of the chunk can be different?
This sounds weird!

LucasBrazi06:
> I'm a bit surprised by the fix.
> Why does the code rely on the EXPORT_PAGE_SIZE constants while the size of the chunk can be different?
> This sounds weird!

Agreed, we should always get the same chunck size in the export CSV.
The fix is not the right one I think.

PR #2724, opened 2021-08-05
Add asset and charging station consumption, asset production and limitation to Site Area Chart

RohanAntony:
Steps involved:
1. Facet the data into 4 pipelines all processing same documents and performing aggregation
2. Perform union of all 4 pipeline data into a single array 
3. Unwind the array into individual objects 
4. Group fields by each timestamp available and set default for missing value of parameter
5. Return the new array of objects as Consumption array

QA 
1. Check if Net consumption in KW is valid for given timestamp
2. Check if Net consumption in A is valid for given timestamp
3. Check if Asset production in KW is valid for given timestamp
4. Check if Asset production in A is valid for given timestamp
5. Check if Asset consumption in KW is valid for given timestamp
6. Check if Asset consumption in A is valid for given timestamp
7. Check if Charging station consumption in KW is valid for given timestamp
8. Check if Charging station consumption in A is valid for given timestamp
9. Check if Limit in KW is valid for given timestamp
10. Check if Limit in A is valid for given timestamp
11. Check if the translated legends are shown in respective languages


LucasBrazi06:
@ClaudeROSSI Could you check this PR?

ClaudeROSSI (approved):

PR #2726, opened 2021-08-05
i18nChecker

rmangin-fr:
To test, type the command:
npm run check:i18n

Current issue with content of fr.json!
Not considering cz.json since it is not part of the supported languages.

Let me know if anything has to be reviewed.


ClaudeROSSI:
Ca fonctionne puisque tu as deja trouver une anomalie:

ClaudeROSSI (approved):
Cool!

I think we should have it running as an automated test to make sure we stop as soon the i18n resources are inconsistent. But this can be added in a second PR!

PR #2734, opened 2021-08-14
Fix computation of evse_uid and evse_uid when a charging station contains several charge points with the flag cannotChargeInParallel set to true

sebastien-savalle:
Fix computation of evse_uid and evse_uid when a charging station contains several charge points with the flag cannotChargeInParallel set to true


ClaudeROSSI (approved):

ClaudeROSSI:
I'll try to deliver it as an hotfix

ClaudeROSSI:
all tests are green

PR #2735, opened 2021-08-15
WIP : Invoices modal back

cabch:
I let comments on what I tried so far

Need to ask stripe the best way to achieve what we want to do :
use the .pay() when doing a one time payment so the payment is attached to the invoice

- today it sounds like it works with normal card number NOT with 3DS card !! // please retry normal card payment
- didn't find a way to attach payment even with payment intent (.pay() wait for a payment_method ID OR a source)
- when doing a stripe payment directly, looking at the logs it looks like they use 'source' stuff but they also say it's deprecated in the doc...

### DIFF
--- a/src/authorization/Authorizations.ts
+++ b/src/authorization/Authorizations.ts
@@ -662,6 +662,18 @@ export default class Authorizations {
       { user: userID, owner: loggedUser.id });
   }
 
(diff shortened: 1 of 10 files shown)


ClaudeROSSI:
code moved to another branch

PR #2742, opened 2021-08-18
When charging stations in site are public, throw error if site is made private

RohanAntony:
Check if an error is thrown when a site with public charging stations is set to private.


RohanAntony:
Do I need to log the Error?

LucasBrazi06:
> Do I need to log the Error?
Normally, when you throw the exception, it should be already logged.

src/server/rest/v1/service/SiteService.ts
```diff
@@ -392,6 +393,19 @@ export default class SiteService {
     site.name = filteredRequest.name;
     site.companyID = filteredRequest.companyID;
     if (Utils.objectHasProperty(filteredRequest, 'public')) {
+      if (!filteredRequest.public) {
+        const publicChargingStations = (await ChargingStationStorage.getChargingStations(req.tenant,
+          { siteIDs: [site.id], public: true, withSiteArea: true }, Constants.DB_PARAMS_MAX_LIMIT)).result;
```
LucasBrazi06, line 398: Limit to one charging stations and project only the charging station's ID

LucasBrazi06 (changes requested):

LucasBrazi06:
Like the other issue, create an issue + link your PR to it + paste the link in Asana.

RohanAntony:
@LucasBrazi06 While throwing the AppError, why do we not see the message in the notification that pops up? It seems like a longer process if the user has to go to the logs and read the message.

LucasBrazi06:
> @LucasBrazi06 While throwing the AppError, why do we not see the message in the notification that pops up? It seems like a longer process if the user has to go to the logs and read the message.

You have to catch the exception in the front-end and display an i18n message ;-) so I would expect a PR in the dashboard also.

LucasBrazi06:
@RohanAntony  Still not attached to an issue, just did it.

src/server/rest/v1/service/SiteService.ts
```diff
@@ -392,6 +393,20 @@ export default class SiteService {
     site.name = filteredRequest.name;
     site.companyID = filteredRequest.companyID;
     if (Utils.objectHasProperty(filteredRequest, 'public')) {
+      if (!filteredRequest.public) {
+        const publicChargingStation = await ChargingStationStorage.getPublicChargingStationInSite(req.tenant, {
```
LucasBrazi06, line 397: No need to create a specific method in Storage for this you can call the method right away.

LucasBrazi06 (changes requested):
You have also to check when you set a charging station public on a non-public site.

LucasBrazi06 (approved):
Ok, still charging station to handle

PR #2744, opened 2021-08-18
Invoices - accept one shot payment

ClaudeROSSI:


ClaudeROSSI:
The PoC for paying an invoice now works properly

BUT !!!! Each payment attempts add a new payment method to the customer.
We need to getr rid of it.

I think the flow must be a bit different.
When "off_session" is used, we get a requires-action error code on the payment intent 
The solution seems to be the one described at:
> c.f.: https://stripe.com/docs/billing/subscriptions/overview#requires-action

Screenshot from 2021-08-18 15-41-48

ClaudeROSSI:
I can make it run with the "off_session" option.
In that case the payment succeeds and the card is not added to the customer.
However, the drawback here is that the invoice is not marked as paid.
The only option is then to use the 'paid_out_of_band' option when calling invoices.pat() ...
But this again as another side effect - there is no clear links in the STRIPE dashboard between the invoice and its payment!

ClaudeROSSI:
When paying using the link that STRIPE provides to pay invoice, everything woks as expected.
However, when looking at the logs it sounds like STRIPE relies on the deprecated SOURCE Api to achieve the invoiced payment. This is very strange and needs to be clarified!

cabch (approved):
approved but don't forget it works with saving the pm - lets merge this before merging the front

PR #2762, opened 2021-09-01
Add additional ids such as siteID, siteAreaID, companyID and chargingStationID to logs for site admin

RohanAntony:
Adding siteID, siteAreaID, companyID and chargingStationID which will be used during getLogs in later PR to fetch data. This will replace the operation of fetching siteID based on chargingStationIDs and instead use the siteID available in the logs directly.  

Submodule PRs
[ev-sap-pricing-convergent-charging](https://github.com/sap-labs-france/ev-sap-pricing-convergent-charging/pull/19)
[ev-sap-smart-charging](https://github.com/sap-labs-france/ev-sap-smart-charging/pull/55)


RohanAntony:
@LucasBrazi06 In the final commit, I have populated the siteID, siteAreaID and companyID in WSConnection.ts file assuming the URL has the values. The discussion on the mail stated that we are fetching the chargingStation object from DB using the chargingStationID but there is no use of ChargingStationStorage in WSConnection. 

Should I keep the approach of fetching data from the URL or change it and add a DB call to fetch the appropriate charging station and populate the fields?

LucasBrazi06:
> The discussion on the mail stated that we are fetching the chargingStation object from DB using the chargingStationID but there is no use of ChargingStationStorage in WSConnection.

It's because it's in its sub-class JsonWSConnection ;-)

LucasBrazi06:
@RohanAntony Create an issue in GitHub and link all of your PR to it.
Put the issue's link into the Asana task.

src/server/ocpp/json/JsonCentralSystemServer.ts
```diff
@@ -45,13 +45,22 @@ export default class JsonCentralSystemServer extends CentralSystemServer {
     this.startWSServer();
   }
 
-  public getChargingStationClient(tenantID: string, chargingStationID: string): ChargingStationClient {
+  public getChargingStationClient(tenantID: string, chargingStationDetails: {
```
LucasBrazi06, line 48: I would have prefered the location data in an dedicated optional param:
```
public getChargingStationClient(tenantID: string, chargingStationID: string,
  chargingStationLocation?: { siteID: string, siteAreaID: string, companyID: string }) {
```
or
```
public getChargingStationClient(tenantID: string, chargingStationID: string,
  siteID?: string, siteAreaID?: string, companyID?: string) {
```

LucasBrazi06 (changes requested):

RohanAntony:
@LucasBrazi06 In the last commit that was pushed a while back, I have made changes to logging.ts and other files related to it. I ran the test cases on my local machine which covers all of them except OCPP 1.6 service test case. This has been consistently failing on my machine for every branch including master-qa. I don't know on how else to test these latest changes except for the test cases.

RohanAntony:
I believe, I need to make changes to AppError and BackendError which also get logged.

LucasBrazi06:
@RohanAntony I can check the unit test on my side if you want on your branch?

LucasBrazi06:
> I believe, I need to make changes to AppError and BackendError which also get logged.

AppError is for HTTP Service endpoint (REST, OCPI...), the rest is BackendError.

LucasBrazi06:
@RohanAntony I just added the most important part you missed + conversion to ObjectIDs :-)

LucasBrazi06:
@RohanAntony I think you missed the Exception that are also logged:


I'll put back the issue to In Process ;-).

RohanAntony:
> @RohanAntony I just added the most important part you missed + conversion to ObjectIDs :-)
> 

Do we not need to add it for chargingStationID which is also an ObjectID?

LucasBrazi06:
> Do we not need to add it for chargingStationID which is also an ObjectID?

No, Charging Station ID is a string in the DB.
You can check this with the Storage classes or with Robot3T.

PR #2788, opened 2021-09-07
REST > Assets JSON schemas

Ninjeneer:
Merging https://github.com/sap-labs-france/ev-dashboard/pull/2551 is required before merging this one. Otherwise, it won't work


src/server/rest/v1/validator/AssetValidator.ts
```diff
@@ -20,10 +33,39 @@ export default class AssetValidator extends SchemaValidator {
...
+    return data;
+  }
+
+  public validateAssetsGetReq(data: any): HttpAssetsRequest {
+    // Validate schema
+    this.validate(this.assetsGet, data);
+    return data;
+  }
+
+  public validateAssetCreateReq(data: Partial<Asset>): Partial<Asset> {
```
LucasBrazi06, line 54: All 'data' should be of type any and return type not Partial -> Asset

LucasBrazi06 (approved):
Small comment

PR #2791, opened 2021-09-07
exportOCPPParameters

rmangin-fr:


src/server/rest/v1/service/ChargingStationService.ts
```diff
@@ -920,6 +920,7 @@ export default class ChargingStationService {
     // Check and Get Charging Station
     const chargingStation = await UtilsService.checkAndGetChargingStationAuthorization(
       req.tenant, req.user, filteredRequest.ID, action, null, {
+        withSite: true,
```
LucasBrazi06, line 923: Good catch.
The migration has not yet been done on this endpoint but I would like that you pass all the params withXXX from the front-end, filter them with the Json Schema and enhance the OAS file (check also with Alix to perform this change also).

src/server/rest/v1/service/ChargingStationService.ts
```diff
@@ -935,6 +936,7 @@ export default class ChargingStationService {
   public static async handleExportChargingStationsOCPPParams(action: ServerAction, req: Request, res: Response, next: NextFunction): Promise<void> {
     // Always with site
     req.query.WithSite = 'true';
+    req.query.WithSiteArea = 'true';
```
LucasBrazi06, line 939: This one is okay as we force it to get the site area in the CSV file.

LucasBrazi06 (changes requested):

src/server/rest/v1/service/ChargingStationService.ts
rmangin-fr: I changed this part but the other one you coded for the export OCPP params from Sites UI is still hardcoded, want to change?

PR #2792, opened 2021-09-07
Refactoring of IDs in REST Json schemas

LucasBrazi06:


src/assets/server/rest/v1/schemas/tag/tag.json
```diff
@@ -2,14 +2,7 @@
...
     "id": {
-      "type": "string",
-      "pattern": "^[\\w\\d]*$",
-      "sanitize": "mongo"
-    },
-    "IDs": {
-      "type": "string",
-      "pattern": "^(([\\w\\d]*)(\\||[\\w\\d])*)?$",
-      "sanitize": "mongo"
+      "$ref": "common.json#/definitions/string-id"
```
Ninjeneer, line 5: We are loosing a bit of accuracy here, as the pattern of "string-id" allows everything while tag IDs should be only letters an digit. I would let this one as it is, what do you think ?

src/assets/server/rest/v1/schemas/transaction/transactions-get.json
```diff
@@ -38,16 +38,16 @@
       "sanitize": "mongo"
     },
     "ConnectorID": {
-      "$ref": "chargingstation.json#/definitions/connectorIDs"
+      "$ref": "common.json#/definitions/string-ids"
```
Ninjeneer, line 41: number-ids ?

src/assets/server/rest/v1/schemas/transaction/transactions-inerror-get.json
```diff
@@ -16,20 +16,20 @@
...
-      "$ref": "common.json#/definitions/IDs"
+      "$ref": "common.json#/definitions/ids"
     },
     "ErrorType": {
       "type": "string",
       "sanitize": "mongo"
     },
     "ConnectorID": {
-      "$ref": "chargingstation.json#/definitions/connectorIDs"
+      "$ref": "common.json#/definitions/string-ids"
```
Ninjeneer, line 29: number-ids ?

src/assets/server/rest/v1/schemas/common/common.json
```diff
@@ -6,11 +6,31 @@
...
+      "pattern": "^(([^|]*)(\\||[^|])*)?$",
+      "sanitize": "mongo"
+    },
+    "number-id": {
+      "type": "number",
+      "pattern": "^(\\d)*$",
+      "sanitize": "mongo"
+    },
+    "number-ids": {
+      "type": "number",
```
Ninjeneer, line 30: Type string, as it will be separated by pipes

Ninjeneer (changes requested):

src/assets/server/rest/v1/schemas/tag/tag.json
LucasBrazi06: I didn't want to put too many constraints on string ids because you will have to create a new type in common to handle the string-tag-id and string-tag-ids in filters also.
But as you want :).

src/assets/server/rest/v1/schemas/transaction/transactions-get.json
LucasBrazi06: It was created as string for the filter (tried with numbers but it failed):

```
"connectorIDs": {
      "type": "string",
      "pattern": "^(\\d\\|?)*$",
      "sanitize": "mongo"
    },
```

src/assets/server/rest/v1/schemas/transaction/transactions-inerror-get.json
LucasBrazi06: Same comment as above

src/assets/server/rest/v1/schemas/common/common.json
LucasBrazi06: Okay, fixed.

src/assets/server/rest/v1/schemas/transaction/transactions-get.json
Ninjeneer: Now that number-ids is fixed, it should work

src/assets/server/rest/v1/schemas/tag/tag.json
Ninjeneer: I will tend to prioritize accuracy against constraints in the files, as users, one day, always mess up their environment :)
It's up to you, you're the boss :)

src/assets/server/rest/v1/schemas/transaction/transactions-get.json
LucasBrazi06: Yes indeed :-).
Just fixed it.

Ninjeneer (approved):

PR #2827, opened 2021-09-20
Hotfix > Fixed ChargingStation action json schema reference

Ninjeneer:
Causes Error 500 on every charging station request


LucasBrazi06 (approved):
Thanks!

PR #2833, opened 2021-09-20
Car Connector - Tronity enablement

OliveGerste:
The access token is stored in the connection of the car connector settings. 
The handling of the sensitive data of the token is currently done in the Tronity Integration class. 
Every time the end user changes the car connector connection config, the token is deleted. 
Let me know if you think something needs to be changed here.


LucasBrazi06:
> The access token is stored in the connection of the car connector settings.
> The handling of the sensitive data of the token is currently done in the Tronity Integration class.
> Every time the end user changes the car connector connection config, the token is deleted.
> Let me know if you think something needs to be changed here.

What does the access token contains?

OliveGerste:
> > The access token is stored in the connection of the car connector settings.
> > The handling of the sensitive data of the token is currently done in the Tronity Integration class.
> > Every time the end user changes the car connector connection config, the token is deleted.
> > Let me know if you think something needs to be changed here.
> 
> What does the access token contains?

The token in the connection contains the barear token + expiration date to access the api of tronity

LucasBrazi06:
> The token in the connection contains the barear token + expiration date to access the api of tronity.
> The purpose is to reduce the amount of calls to their backend.

Okay but you can decode it and check the data inside.
Nevertheless, if someone accesses the DB and get the token, could he send requests to Tronity?

LucasBrazi06:
PR not linked to the issue ;-)

OliveGerste:
> PR not linked to the issue ;-)
> Just did it

Thanks. 
It is weird sometimes. I am pretty sure I linked both PRs to the issue. 
It was not the first time, I experienced this. I will double check in the future

OliveGerste:
> > The token in the connection contains the barear token + expiration date to access the api of tronity.
> > The purpose is to reduce the amount of calls to their backend.
> 
> Okay but you can decode it and check the data inside.
> Nevertheless, if someone accesses the DB and get the token, could he send requests to Tronity?

The token is encrypted similar to passwords and client secrets.

LucasBrazi06:
> It is weird sometimes. I am pretty sure I linked both PRs to the issue.
> It was not the first time, I experienced this. I will double check in the future

Agree, there are often problems with ZenHub :-(

src/integration/car-connector/tronity-connector/TronityCarConnectorIntegration.ts
```diff
@@ -0,0 +1,141 @@
...
+        module: MODULE_NAME,
+        method: 'getCurrentSoC',
+        action: ServerAction.CAR_CONNECTOR,
+        message: 'Error while retrieving the SOC',
+        detailedMessages: { request, error: error.stack }
+      });
+    }
+  }
+
+
```
LucasBrazi06, line 85: BL

src/assets/server/rest/v1/schemas/car/car.json
```diff
@@ -76,11 +76,11 @@
       ],
       "properties": {
         "carConnectorID": {
-          "type": "string",
+          "type": ["string", "null"],
```
LucasBrazi06, line 79: Weird the "null" value :S

src/integration/car-connector/tronity-connector/TronityCarConnectorIntegration.ts
```diff
@@ -0,0 +1,141 @@
...
+        {
+          'axios-retry': {
+            retries: 0
+          },
+          headers: this.buildFormHeaders()
+        }),
+      `Time out error (5s) when getting the token with the connection URL '${this.connection.tronityConnection.apiUrl}/oauth/authentication'`
+    );
+    const currentTime = new Date();
+    const token : CarConnectorConnectionToken = {
```
LucasBrazi06, line 99: ES (Extra Space) :-)

src/types/Setting.ts
```diff
@@ -333,12 +333,15 @@ export interface CarConnectorConnectionSetting {
   description: string;
   timestamp: Date;
   type: CarConnectorConnectionType;
+  token?: CarConnectorConnectionToken;
   mercedesConnection?: CarConnectorMercedesConnectionType;
```
LucasBrazi06, line 337: I would expect 'mercedes' and 'tronity'.
We have a link between the Type and the name of the prop:


And we use it when updating the Tenant's components in the mater tenant.

LucasBrazi06 (changes requested):

src/types/Setting.ts
OliveGerste: The whole car connector framework is built like the asset framework.

If we choose the connection at master tenant level, we are not able to use different connectors for different vehicles, which is required, when we have vehicles of different manufacturers in our fleet.

LucasBrazi06: My bad, the mercedesConnector prop is one level below so that's okay.

LucasBrazi06 (approved):

PR #2839, opened 2021-09-21
Fix apply template on backed up connectors

LucasBrazi06:


OliveGerste (commented):
I think the changes made will not solve the root cause of the problem.

The main issue is in OCPPService, when the cs boot notification is handled and the template is applied.

There we call a method `enrichChargingStation`, which clears all connectors:

https://github.com/sap-labs-france/ev-server/blob/f08c5710e662fe35918191e921cb252a5ebe475a/src/server/ocpp/services/OCPPService.ts#L2023

After this we apply the template by iterating through the connectors, but this is an empty array now: 
https://github.com/sap-labs-france/ev-server/blob/f08c5710e662fe35918191e921cb252a5ebe475a/src/server/ocpp/utils/OCPPUtils.ts#L2057-L2059

As a result the connectors are not enriched with the template.

LucasBrazi06:
> The main issue is in OCPPService, when the cs boot notification is handled and the template is applied.

At boot notification the template is only applied at charging station level.
The boot notification should always be followed by status notifications in which the connectors are created/retrieved and the template is applied.

> After this we apply the template by iterating through the connectors, but this is an empty array now:

Yes this code is not used anymore, should be removed

OliveGerste:
> > The main issue is in OCPPService, when the cs boot notification is handled and the template is applied.
> 
> At boot notification the template is only applied at charging station level.
> The boot notification should always be followed by status notifications in which the connectors are created/retrieved and the template is applied.

Okay if I am right the following method is used for this:
https://github.com/sap-labs-france/ev-server/blob/f08c5710e662fe35918191e921cb252a5ebe475a/src/server/ocpp/services/OCPPService.ts#L859-L893

If I understand correctly we only apply the template, if the charging station does not have connectors and backup connectors, which can be retrieved from the database right?

LucasBrazi06:
> Okay if I am right the following method is used for this.

Yes, this is the right method where we apply the template against the connectors.

> If I understand correctly we only apply the template, if the charging station does not have connectors and backup connectors, which can be retrieved from the database right?

In Boot Notif, only at charging station level
In Status Notif only at connector level

OliveGerste:
> > If I understand correctly we only apply the template, if the charging station does not have connectors and backup connectors, which can be retrieved from the database right?
> 
> In Boot Notif, only at charging station level
> In Status Notif only at connector level

In the method we only take action, if no connector is already available from the db):
https://github.com/sap-labs-france/ev-server/blob/f08c5710e662fe35918191e921cb252a5ebe475a/src/server/ocpp/services/OCPPService.ts#L859-L862

If no connector is available at the db, the backup connector is used (I think this is the case for Mougins 16):
https://github.com/sap-labs-france/ev-server/blob/f08c5710e662fe35918191e921cb252a5ebe475a/src/server/ocpp/services/OCPPService.ts#L862-L869

And only if no back up is available, then the template is applied:
https://github.com/sap-labs-france/ev-server/blob/f08c5710e662fe35918191e921cb252a5ebe475a/src/server/ocpp/services/OCPPService.ts#L870-L889

We always have ha backup connector available if the station was registered before, so in this case we cannot enrich the connectors anymore. Or do I make a mistake here?

LucasBrazi06:
> We always have ha backup connector available if the station was registered before, so in this case we cannot enrich the connectors anymore. Or do I make a mistake here?

You're absolutely right :-)
And this is why I made a fix in this PR :-) where I always apply the template to the found connector regardless its origin (backup or new.)

OliveGerste:
> You're absolutely right :-)
> And this is why I made a fix in this PR :-) where I always apply the template to the found connector regardless its origin (backup or new.)

Ah okay now I finally understood the whole scenario. 
Okay, then the fix should do it, sorry for the long process of understanding from my side.

LucasBrazi06:
> Ah okay now I finally understood the whole scenario.
> Okay, then the fix should do it, sorry for the long process of understanding from my side.

No problem :-)

PR #2856, opened 2021-09-23
Split get configuration

rmangin-fr:


LucasBrazi06 (commented):
I think we can also get rid of the executeChargingStationCommand() method

LucasBrazi06 (approved):
I validate it but check to remove the ChargingStationService.executeChargingStationCommand()

PR #2887, opened 2021-09-29
Site Admin assign users to his sites bug

mahdibenromdhane:


src/server/rest/v1/service/UserService.ts
```diff
@@ -672,44 +672,46 @@ export default class UserService {
...
-    if (Authorizations.isAdmin(req.user) || Authorizations.isSuperAdmin(req.user)) {
-      // Save User Status
-      if (newUser.status) {
-        await UserStorage.saveUserStatus(req.tenant, newUser.id, newUser.status);
-      }
-      // Save User Role
-      if (newUser.role) {
-        await UserStorage.saveUserRole(req.tenant, newUser.id, newUser.role);
+    // Save User Status
+    if (newUser.status) {
```
LucasBrazi06, line 676: Auth must be checked I think to avoid creating an Admin user via Postman, right?

src/server/rest/v1/service/UserService.ts
```diff
@@ -672,44 +672,46 @@ export default class UserService {
...
+      await UserStorage.saveUserStatus(req.tenant, newUser.id, newUser.status);
+    }
+    // Save User Role
+    if (newUser.role) {
```
LucasBrazi06, line 680: Same here?

src/server/rest/v1/service/UserService.ts
```diff
@@ -672,44 +672,46 @@ export default class UserService {
...
-    // Assign user to all sites with auto-assign flag set
-    const sites = await SiteStorage.getSites(req.tenant,
-      { withAutoUserAssignment: true },
-      Constants.DB_PARAMS_MAX_LIMIT
-    );
-    if (!Utils.isEmptyArray(sites.result)) {
-      const siteIDs = sites.result.map((site) => site.id);
-      await UserStorage.addSitesToUser(req.tenant, newUser.id, siteIDs);
+    let siteIDs = [];
+    if (!Utils.isEmptyArray(req.user.sitesAdmin)) {
```
LucasBrazi06, line 702: We want to remote it from the token, so don't use it ;-)
Check authorizationFilter in the datasource you should have site IDs (check DynamicAuthorizationDataSourceName).

LucasBrazi06 (changes requested):

src/server/rest/v1/service/UserService.ts
mahdibenromdhane: I think the new BasicUser assert will prevent that.

src/server/rest/v1/service/UserService.ts
mahdibenromdhane: True ! Done.

LucasBrazi06:
Put the code directly to master-qa.
Closing.

PR #2910, opened 2021-10-03
Sanitize csv output

TheGodMother25:


src/utils/Utils.ts
```diff
@@ -872,6 +872,10 @@ export default class Utils {
     return !Object.keys(obj).length;
   }
 
+  public static isEmptyString(str: string): boolean {
+    return str ? str.length === 0 : true;
```
LucasBrazi06, line 876: [low] If this returns true, it means 'null' string and not 'empty' string ;-)
So the method name should be called isNullOrEmptyString()
I'll change it.

src/utils/Utils.ts
```diff
@@ -1556,4 +1560,29 @@ export default class Utils {
   public static unescapeCsvValue(value: any): void {
     // double quotes are handle by csvToJson
   }
+
+  public static sanitizeCSVExport(data: any): any {
```
LucasBrazi06, line 1564: [low] You can use 'unknown' instead of 'any' to tell the linter that you know what you are doing :-)

src/utils/Utils.ts
```diff
@@ -1556,4 +1560,29 @@ export default class Utils {
...
+    if (!data || typeof data === 'number' || typeof data === 'bigint' || typeof data === 'symbol' || Utils.isBoolean(data) || typeof data === 'function') {
+      return data;
+    } else if (typeof data === 'string') {
```
LucasBrazi06, line 1567: [low] You can chain with simple 'if' statements, no need the 'else if', just to make it more readable.

src/utils/Utils.ts
```diff
@@ -1556,4 +1560,29 @@ export default class Utils {
...
+      }
+      return sanitizedData;
+    } else if (typeof data === 'object') {
+      // If the data is an object, apply the sanitizeCSVExport function for each attribute
+      for (const key of Object.keys(data)) {
+        data[key] = Utils.sanitizeCSVExport(data[key]);
+      }
+      return data;
+    }
+  }
```
LucasBrazi06, line 1587: Default 'return' statement missing?

TheGodMother25: it's null :p
I'll add it

src/utils/Utils.ts
LucasBrazi06: btw an empty string is considered as null in JS :-)
So 'str.length === 0' will never be evaluated if 'str' equals to ""

src/utils/Utils.ts
TheGodMother25: the data.result is of type any :(

LucasBrazi06: hahaha, we have not yet changed it everywhere :).
If it's too many changes, forget it.
Thanks!

LucasBrazi06 (approved):
Thanks!
Don't forget to check the pipeline where the PR is (Requested change to move to To be reviewed)

PR #2913, opened 2021-10-04
technical: and & or filters

rmangin-fr:
Free text search + NON-technical users is non-effective.

### DIFF
--- a/src/storage/mongodb/UserStorage.ts
+++ b/src/storage/mongodb/UserStorage.ts
@@ -654,10 +654,15 @@ export default class UserStorage {
     // Select users with test billing data
     if (params.withTestBillingData) {
       const expectedLiveMode = !params.withTestBillingData;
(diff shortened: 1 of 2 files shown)


src/storage/mongodb/UserStorage.ts
```diff
@@ -664,16 +664,16 @@ export default class UserStorage {
...
+          $or: [
+            { technical: { $in: [false, null] } },
+            { technical: { $exists: false } }
+          ]
+        };
+        if (filters.$and) {
+          filters.$and.push(technicalFilter);
         } else {
-          filters.$or = technicalFilter;
+          filters.$and = [ technicalFilter ];
```
LucasBrazi06, line 676: Indeed, it's $and with the rest of the query. Good catch.
Could you change also: params.notSynchronizedBillingData (just above)

PR #3015, opened 2021-11-02
Tariff id in site

rmangin-fr:


src/server/rest/v1/service/SiteService.ts
```diff
@@ -405,6 +405,10 @@ export default class SiteService {
       }
       site.public = filteredRequest.public;
     }
+    if (Utils.objectHasProperty(filteredRequest, 'tariffID') &&
+      filteredRequest.tariffID !== '') {
```
ClaudeROSSI, line 409: does not work. ... once a  tariff ID is set we cannot clear it anymore!

ClaudeROSSI (changes requested):

ClaudeROSSI (approved):
ok

PR #3023, opened 2021-11-04
Tariff id in charging station

rmangin-fr:


src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/OCPIUtilsService.ts
```diff
@@ -936,6 +936,27 @@ export default class OCPIUtilsService {
...
+    // Site Area?
+    if (chargingStation.siteArea && chargingStation.siteArea.tariffID) {
+      return chargingStation.siteArea.tariffID;
+    }
+    // Site?
+    if (chargingStation.site && chargingStation.site.tariffID) {
+      chargingStation.site.tariffID;
+    }
+    // Tenant?
+    // TODO: Get TariffID from SettingStorage (pending prerequisite PR).
```
rmangin-fr, line 956: Waiting to merge issue 2960.

src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/OCPIUtilsService.ts
```diff
@@ -936,6 +936,27 @@ export default class OCPIUtilsService {
...
+    // Connector?
+    if (connector.tariffID) {
+      return connector.tariffID;
+    }
+    // Charging Station?
+    if (chargingStation.tariffID) {
+      return chargingStation.tariffID;
+    }
+    // Site Area?
+    if (chargingStation.siteArea && chargingStation.siteArea.tariffID) {
```
ClaudeROSSI, line 948:
```suggestion
    if (chargingStation.siteArea?.tariffID) {
```

src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/OCPIUtilsService.ts
```diff
@@ -936,6 +936,27 @@ export default class OCPIUtilsService {
...
+      return chargingStation.siteArea.tariffID;
+    }
+    // Site?
+    if (chargingStation.site && chargingStation.site.tariffID) {
```
ClaudeROSSI, line 952:
```suggestion
    if (chargingStation.site?.tariffID) {
```

src/server/ocpi/ocpi-services-impl/ocpi-2.1.1/OCPIUtilsService.ts
```diff
@@ -936,6 +936,27 @@ export default class OCPIUtilsService {
...
+      chargingStation.site.tariffID;
+    }
+    // Tenant?
+    // TODO: Get TariffID from SettingStorage (pending prerequisite PR).
+    // Default.
+    return defaultTariff;
+    // OLD rules (for setting them into the proper objects)
```
ClaudeROSSI, line 959: Unreachable code after line #958. Why do we keep it?
Don't we have a "transition phase" where we need to keep it to avoid side-effects?

src/server/rest/v1/service/ChargingStationService.ts
```diff
@@ -169,6 +169,11 @@ export default class ChargingStationService {
         }
       }
       chargingStation.public = filteredRequest.public;
+      if (filteredRequest.public) {
+        chargingStation.tariffID = filteredRequest.tariffID;
+      } else if (chargingStation.tariffID) {
+        delete chargingStation.tariffID;
+      }
```
ClaudeROSSI, line 176: Why don't we just assign the tariffID ? 
It would be convenient to allow switching ON/OFF the "public" property without loosing the actual value of the tariffID. 
Let me know what you think.

src/server/rest/v1/service/ChargingStationService.ts
```diff
@@ -219,6 +224,11 @@ export default class ChargingStationService {
           connector.numberOfConnectedPhase = filteredConnector.numberOfConnectedPhase;
         }
         connector.phaseAssignmentToGrid = filteredConnector.phaseAssignmentToGrid;
+        if (filteredRequest.public) {
+          connector.tariffID = filteredConnector.tariffID;
+        } else if (connector.tariffID) {
+          delete connector.tariffID;
+        }
```
ClaudeROSSI, line 231: same comment ... can't we just assign the tariffID in any cases?

ClaudeROSSI (changes requested):
a few comments

ClaudeROSSI (commented):
Let's wait for the other PR before approving this one!

src/server/rest/v1/service/ChargingStationService.ts
```diff
@@ -165,6 +165,9 @@ export default class ChargingStationService {
       }
       chargingStation.public = filteredRequest.public;
     }
+    if (Utils.objectHasProperty(filteredRequest, 'tariffID')) {
```
ClaudeROSSI, line 168: I'm not sure why we need to check for the property here.
This is something which is not done in **SiteAreaService**.

From my perspective the check is useless. On the other hand, this was done like that for some other properties.
@LucasBrazi06 , @OliveGerste - can't we just check 

if ( filteredRequest.tariffID ) {
...
}

ClaudeROSSI (approved):
approved.

One comment on some Utils.objectHasProperty() usages (which seem useless to me).

src/server/rest/v1/service/ChargingStationService.ts
OliveGerste: This bit was added by @rmangin-fr. 
I am also not sure why we check it like this.

ClaudeROSSI: ok. let's keep it.
This should have no impact.

LucasBrazi06: Copy/Paste without thinking ;-)
This is used only for boolean prop check.
You can use a simple IF statement.
Thanks!

OliveGerste: @LucasBrazi06 
I think the has property check is required if we want to be able to clear the tariffID. 
In this case the value of tariffID is null and if we check with IF statement the null value will set the statement false and it is not cleared in the db.

LucasBrazi06: @OliveGerste
You're absolutely right!
In the new Auth framework this will be filtered out by the fields you have the right to see then this code will no longer be necessary but ChargingStationService has not yet been migrated :-( !
My bad!

PR #3024, opened 2021-11-04
Asset integration test cases for ioThink

RohanAntony:
Test cases for ioThink asset integration
Before executing the test case, set the below parameters in the local.json
```
"assetConnectors": {
  "ioThink": {
    "user": ioThink username,
    "password": ioThink password,
    "meterID": meterID connection string used to connect to actual asset,
    "url": optional parameter, defaults to "https://api.kheiron-sp.io"
  }
}
```

Test:
- Setup the parameters as metioned above
- Verify if the asset connector test cases for ioThink are passing successfully


LucasBrazi06:
@OliveGerste Will you review this PR?

OliveGerste:
> @OliveGerste Will you review this PR?

Yes I can do it.

test/api/AssetTest.ts
```diff
@@ -239,6 +249,107 @@ describe('Asset', function() {
         expect(response.status).to.be.eq(StatusCodes.OK);
         expect(response.data.count).to.be.eq(1);
       });
+
+    });
+
+    describe('Where admin user with asset ', () => {
```
OliveGerste, line 255: You can include the tests in the already existing admin test section.

test/api/AssetTest.ts
```diff
@@ -239,6 +249,107 @@ describe('Asset', function() {
...
+        // Update
+        await testData.userService.updateEntity(
+          testData.userService.assetApi,
+          testData.newAsset
+        );
+      });
+
+      it('should be able to retrieve latest consumption', async () => {
+        const response = await testData.centralUserService.assetApi.retrieveLatestConsumption(testData.newAsset.id);
+        expect(response.data?.status).to.be.eq('Success');
```
OliveGerste, line 350: Please also check the response properties returned

test/api/AssetTest.ts
```diff
@@ -239,6 +249,107 @@ describe('Asset', function() {
...
+          await testData.centralUserService.settingApi.create(setting);
+        }
+        await testData.centralUserService.settingApi.update(setting);
+        const settingsUpdated = await testData.centralUserService.settingApi.readAll({});
+        const settingUpdated: SettingDB = settingsUpdated.data.result.find((s) => s.identifier === TenantComponents.ASSET);
+        expect(settingUpdated.content.asset.connections[0].id).to.be.eq(settingID);
+        testData.ioThinkAssetConnectorID = settingID;
+      });
+
+      it('should be able to connect ioThink asset connector connection', async () => {
```
OliveGerste, line 330: If the credentials are changed, the connection cannot be validated, as long as the token is still in the cache.

OliveGerste (changes requested):
Please handle the case when the asset connection is not set in the local test config. 
You can add pending state like it is done in smart charging.
Additionally please update the config-template.json with the new properties.

It would be also nice to have test cases for wrong parameters (credentials, meterID, etc.)

test/api/AssetTest.ts
```diff
@@ -142,6 +148,10 @@ describe('Asset', function() {
         }
       });
 
+      after(async () => {
```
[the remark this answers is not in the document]
RohanAntony: I kept it thinking I have to log out the user in the after section. Should I logout the user in after section and set it up in before section of describe?

test/api/AssetTest.ts
RohanAntony: The before section of this describe checks if the keys for ioThink are set. If not, this single describe case fails whereas all the non asset connector based admin test cases are still checked for in the previous describe.

test/api/AssetTest.ts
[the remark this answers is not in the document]
RohanAntony: I think I will try an alternative of skipping certain testcases if the variables are not set like [here](https://stackoverflow.com/questions/32723167/how-to-programmatically-skip-a-test-in-mocha). This way, I can integrate all of them into one single admin test cases.

test/api/AssetTest.ts
RohanAntony: I think I will try an alternative of skipping certain testcases if the variables are not set like [here](https://stackoverflow.com/questions/32723167/how-to-programmatically-skip-a-test-in-mocha). This way, I can integrate all of them into one single admin test cases.

test/api/AssetTest.ts
RohanAntony: There are no other properties except status in the data field here. This checks if the latest consumption is retrieved from ioThink and stored. Post storage, a success response is sent.

test/api/AssetTest.ts
RohanAntony: I think its better to leave it in a separate describe instead of adding it to admin cases. This is because, in 'admin only' case, we are not creating an asset in the before section and the test is written to check if asset creation is supported for an admin. Here we are checking if an admin with existing asset can connect to a connector. This way, we need an asset created for every test case in this describe due to which it is added in the before section but not in the previous 'admin only' describe. 
One alternative is to create an asset in the first asset connector test case but this will create a dependency between the test. Then one can't run the test cases independently and need to run all of them so that the asset is created. With the current way, the before section is executed before a single or all test case that is run.

test/api/AssetTest.ts
RohanAntony: I have added another test case that changes the settings of the connection and invalidates the connection. It seems to be working which should mean that the token is getting refreshed.

test/api/AssetTest.ts
OliveGerste: Okay, should be fine then

test/api/AssetTest.ts
```diff
@@ -320,6 +363,82 @@ describe('Asset', function() {
...
+          Factory.asset.build({
+            siteAreaID: testData.newSiteArea.id,
+            assetType: 'PR',
+          })
+        );
+        testData.createdAssets.push(testData.newAsset);
+        expect(testData.newAsset).to.not.be.null;
+      });
+
+      it('Should be able to create an ioThink asset connector', async function() {
```
OliveGerste, line 398: This test is using different functionality, if it is running the first time after context creation.
The creation of the asset connector is only tested in the first run. If your run the tests more than once updating the connector is checked.

OliveGerste (changes requested):
One additional thing to fix. 
I also made some cleanups in the code.

test/api/AssetTest.ts
RohanAntony: The key used to access the token cache in IoThinkAssetIntegration as well as WitAssetIntegration had been changed to include user and encrypted password along with original connection id. This way, if the details such as username or password is changed, a new token is fetched again instead of reusing existing tokens.

OliveGerste: Correct, but the test case says 'Should be able to create an ioThink asset connector'. 
But it is only created in the first test run, in the second it is updated. 
This has nothing to do with the token.

OliveGerste (changes requested):

test/api/AssetTest.ts
RohanAntony: Added the changes to delete the settings once the asset connector tests are executed.

LucasBrazi06:
@RohanAntony PR with review assigned to people other than me should be put in Cross review pipeline.
I just moved it there ;-)

OliveGerste (approved):

PR #3039, opened 2021-11-08
Migration transactions endpoint 1

rmangin-fr:


LucasBrazi06:
@rmangin-fr I think there are files in this PR that should not be changed.
I don't think you'll have time to work on this one and Transaction Service has been changed meantime, so we should start over with this one.

PR #3062, opened 2021-11-16
Sub site area enablement

OliveGerste:


src/utils/Utils.ts
```diff
@@ -1571,4 +1572,65 @@ export default class Utils {
...
+    const siteAreaTrees: SiteArea[] = [];
+    // Build tree
+    siteAreaList.forEach((siteArea) => {
+      if (!Utils.isNullOrUndefined(siteArea.parentSiteAreaID)) {
+        // Check if site area chain is meeting the constraints
+        if (!Utils.isNullOrUndefined(hashTable[siteArea.parentSiteAreaID]) &&
+          hashTable[siteArea.parentSiteAreaID].smartCharging === hashTable[siteArea.id].smartCharging &&
+          hashTable[siteArea.parentSiteAreaID].siteID === hashTable[siteArea.id].siteID &&
+          hashTable[siteArea.parentSiteAreaID].voltage === hashTable[siteArea.id].voltage &&
+          hashTable[siteArea.parentSiteAreaID].numberOfPhases === hashTable[siteArea.id].numberOfPhases) {
```
ClaudeROSSI, line 1606: These checks do not do anything anymore because siteID, voltage and numberOfPhases are always undefined.

Please check if these properties are part of the projection used when loading the site areas.

ClaudeROSSI (changes requested):
The check of the hierarchy consistency does not work anymore.
Please check my comments

src/server/rest/v1/service/SiteAreaService.ts
```diff
@@ -123,6 +123,9 @@ export default class SiteAreaService {
     // Check and Get Site Area
     const siteArea = await UtilsService.checkAndGetSiteAreaAuthorization(
       req.tenant, req.user, siteAreaID, Action.DELETE, action);
+    const siteAreas = await SiteAreaStorage.getSiteAreas(req.tenant, { siteIDs: [siteArea.siteID] }, Constants.DB_PARAMS_MAX_LIMIT, ['id', 'parentSiteAreaID', 'siteID', 'smartCharging', 'name']);
+    // Remove site area from array and check site area chain validity
+    UtilsService.checkIfSiteAreaTreeValid(siteAreas.result.filter((siteAreaToDelete) => siteAreaToDelete.id !== siteArea.id), req, action);
```
ClaudeROSSI, line 128: I'm not sure to understand why we check the whole hierarchy again when deleting a site area.

From my perspective it would be easier to just make sure no "site areas" have the current area as parent.
What do you think?

ClaudeROSSI (changes requested):
just one additional comment

OliveGerste:
> The check of the hierarchy consistency does not work anymore. Please check my comments

Thanks again for the review. I have implemented your suggestions, added the properties to the JSON schemes and also added the tests for the missing properties.

src/assets/schemas/site-area/site-area.json
```diff
@@ -37,6 +37,11 @@
     "siteID": {
       "$ref": "common#/definitions/id"
     },
+    "parentSiteAreaID": {
+      "type": ["string", "null"],
+      "pattern": "^$|^[0-9a-fA-F]{24}$",
+      "sanitize": "mongo"
```
ClaudeROSSI, line 43: This is a regular ID. Why don't you reuse the usual ID definition (like the line 38 for siteID)?

```suggestion
     "$ref": "common#/definitions/id"
```

src/assets/server/rest/v1/schemas/site-area/site-area-create.json
```diff
@@ -19,14 +19,19 @@
       "$ref": "site-area#/definitions/voltage"
     },
     "smartCharging": {
-      "$ref": "site-area#/definitions/smartCharging"
+      "$ref": "site-area#/definitions/smartCharging",
+      "default": false
```
ClaudeROSSI, line 23: ??? You are the only one using this!

The property is not mandatory. So why do we need to define a default value here?

Note that in some schemas (like for the settings) we have :
>>> "nullable": true

But it is maybe a different use-case!

src/assets/server/rest/v1/schemas/site-area/site-area-create.json
```diff
@@ -19,14 +19,19 @@
...
     },
     "accessControl": {
       "$ref": "site-area#/definitions/accessControl"
     },
     "siteID": {
       "$ref": "site-area#/definitions/siteID"
     },
+    "parentSiteAreaID": {
+      "$ref": "site-area#/definitions/parentSiteAreaID",
+      "default": null
```
ClaudeROSSI, line 33: idem! why?

src/assets/server/rest/v1/schemas/site-area/site-area-update.json
```diff
@@ -22,14 +22,19 @@
...
     },
     "accessControl": {
       "$ref": "site-area#/definitions/accessControl"
     },
     "siteID": {
       "$ref": "site-area#/definitions/siteID"
     },
+    "parentSiteAreaID": {
+      "$ref": "site-area#/definitions/parentSiteAreaID",
+      "default": null
```
ClaudeROSSI, line 36: idem

ClaudeROSSI (changes requested):
i have some comments again.

please let me know what you think

ClaudeROSSI:
@OliveGerste : I cannot create a site area anymore even with no parent site area (with the latest changes). 
Please double check.

src/assets/server/rest/v1/schemas/site-area/site-area-create.json
OliveGerste: The reason for the default value is, that the smart charging property always had a default, when we used the security classes. (filterBoolean-Method)
So when the property was not set in the request, smart charging was set to false. So every existing site area has the property set.
If we just leave the property, because not mandatory, this causes errors.

For example: One site area is already saved with the property set to false and a new one without the property. 
In this case the check for the consistency of the tree fails. 
As a result you would not be able to add new site area as parents to existing site areas, which already have the property.

src/assets/schemas/site-area/site-area.json
OliveGerste: Yes in fact it is a regular id, but it needs to be nullable, because otherwise we are not able to clear the site area parent again. 

I will change it to "nullable: true" instead of the array.

src/assets/server/rest/v1/schemas/site-area/site-area-create.json
OliveGerste: If we do not set the default here, the validator returns an empty string for the property. 
This empty string is then used to search for site areas. with an empty id. This results in errors.
This is why I use null as default for the property.

src/assets/server/rest/v1/schemas/site-area/site-area-update.json
OliveGerste: idem

OliveGerste:
> @OliveGerste : I cannot create a site area anymore even with no parent site area (with the latest changes). Please double check.

You are right, there is a problem if no parent is set.
Reason: the front end is sending empty strings for the parent ID if it is not set. These empty strings are used to search for the parent site area (which is not existing of course).
Would you consider this as a front end bug? Or should I handle empty strings in the backend?

src/server/rest/v1/service/UtilsService.ts
```diff
@@ -1380,6 +1380,51 @@ export default class UtilsService {
...
+          if (siteArea.parentSiteAreaID === siteAreaToDelete) {
+            throw new AppError({
+              errorCode: HTTPError.SITE_AREA_HIERARCHY_DEPENDENCY_ERROR,
+              message: 'Site Area has dependencies to other site areas',
+              module: MODULE_NAME, method: 'checkIfSiteAreaTreeValid',
+              user: req.user.id
+            });
+          }
+        });
+        break;
```
ClaudeROSSI, line 1398: This is ok.
But why a switch here?
Why don't we but the DELETE case in a dedicated method - e.g.: checkIfSiteAreaHasDependencies()

src/types/SiteArea.ts
```diff
@@ -37,8 +38,11 @@ export default interface SiteArea extends CreatedUpdatedProps, SiteAreaAuthoriza
   smartCharging: boolean;
   accessControl: boolean;
   chargingStations: ChargingStation[];
+  assets: Asset[];
```
ClaudeROSSI, line 41: ??? is this from another PR ???

ClaudeROSSI (approved):
approve with comments.

ClaudeROSSI:
@OliveGerste : I added a few comments.

This is still DRAFT? Are you ok to make it "Ready for review"?
Thanks
Claude

src/types/SiteArea.ts
OliveGerste: This comes with the refactoring of the smart charging requests creation. 
It is used in the smart charging integration. I do a lookup for assets from site area to save db requests and improve performance.

src/utils/Utils.ts
```diff
@@ -1587,4 +1588,65 @@ export default class Utils {
   public static hash(data: string): string {
     return crypto.createHash('sha256').update(data).digest('hex');
   }
+
+  public static buildSiteAreaTree(siteAreaId: string, siteAreaList: SiteArea[]): SiteArea {
```
LucasBrazi06, line 1592: Call the renamed buildSiteAreaTree() and rename this one findRootElementInSiteAreaTree() or use a simple find JS method and remove this one.

src/utils/Utils.ts
```diff
@@ -1587,4 +1588,65 @@ export default class Utils {
...
+      if (requestedTree) {
+        return siteAreaTree;
+      }
+      // If it is not containing the ID go to next tree
+      continue;
+    }
+    return null;
+  }
+
+  public static buildSiteAreaTrees(siteAreaList: SiteArea[]): SiteArea[] {
```
LucasBrazi06, line 1607: To rename to buildSiteAreaTree()

src/utils/Utils.ts
```diff
@@ -1587,4 +1588,65 @@ export default class Utils {
...
+        }
+      } else {
+        // If no parent ID is defined push root site area to array
+        siteAreaTrees.push(hashTable[siteArea.id]);
+      }
+    });
+    return siteAreaTrees;
+  }
+
+  private static checkIfSiteAreaInTree(siteAreaTree: Partial<SiteArea>, siteAreaId: string): boolean {
```
LucasBrazi06, line 1636: This kind of check should be in buildTree.

LucasBrazi06 (changes requested):
Check my comments.
I'll check the front-end after your changes.
Thanks!

src/server/rest/v1/service/UtilsService.ts
```diff
@@ -1380,6 +1380,49 @@ export default class UtilsService {
...
+        });
+      }
+    });
+  }
+
+  public static checkIfSiteAreaTreeValid(siteAreaList: SiteArea[], req: Request): void {
+    let siteAreaTrees: SiteArea[];
+    const count = { value: 0 };
+    try {
+      siteAreaTrees = Utils.buildSiteAreaTrees(siteAreaList);
```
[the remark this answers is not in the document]
OliveGerste: I can rename it. I just want to give the reason for the plural real quick - The method returns an array of site area trees. 
So if you put 10 site areas as a parameter and there are 5 root site areas in it, you get 5 trees back. That is why plural for trees.

src/server/rest/v1/service/SiteAreaService.ts
```diff
@@ -288,6 +293,12 @@ export default class SiteAreaService {
       createdBy: { id: req.user.id },
       createdOn: new Date()
     } as SiteArea;
+    const siteAreas = await SiteAreaStorage.getSiteAreas(req.tenant, { siteIDs: [newSiteArea.siteID] }, Constants.DB_PARAMS_MAX_LIMIT, ['id', 'parentSiteAreaID', 'siteID', 'smartCharging', 'name', 'voltage', 'numberOfPhases']);
```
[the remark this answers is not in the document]
OliveGerste: Could you explain this a bit more? 
If I reduce the projection to only the ids, I cannot check that the properties like smart charging and siteID are all the same in one tree. 
So we could have site areas without smart charging and with smart charging enabled in one tree, which will cause errors when smart charging is executed.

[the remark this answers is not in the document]
LucasBrazi06: Okay, I understand now.
Just one question, if you have one Site Area at the end of the tree without Smart Charging, it should be okay I think, right?
And if the component has no Smart Charging, this check should not fail (which should already be the case I think)
Thanks!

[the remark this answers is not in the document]
OliveGerste: > Okay, I understand now. Just one question, if you have one Site Area at the end of the tree without Smart Charging, it should be okay I think, right? 

- If I understood correctly you mean a chain like this (x stands for smart charging enabled): 
Site Area (x) -> Site Area (x) -> Site Area ()
If the smart charging is disabled in the last child of the tree, the first two nodes do not know how to take into account the last site area as there is not smart charging enablement.

- If we have a single site area without dependencies to other site areas (neither parent or child) you can change the properties independently. All that mattes is that the different dependency trees of one site have the same enablement (enabled or disabled)

> And if the component has no Smart Charging, this check should not fail (which should already be the case I think) Thanks!

- If the smart charging component is not active we do not have the enablement checkbox in the edit site area dialog.
As a result all site areas will be saved without smart charging and will pass the test.
So all that matters is that the projected properties are all the same within one site area tree (chain). 
The different trees of the site can be seen independently.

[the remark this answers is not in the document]
LucasBrazi06: > * If I understood correctly you mean a chain like this (x stands for smart charging enabled):
>   Site Area (x) -> Site Area (x) -> Site Area ()

Yes

>   If the smart charging is disabled in the last child of the tree, the first two nodes do not know how to take into account the last site area as there is not smart charging enablement.

We should ignore the last Site Area in this case if this is possible or consider it as a static limitation?

> * If we have a single site area without dependencies to other site areas (neither parent or child) you can change the properties independently. All that mattes is that the different dependency trees of one site have the same enablement (enabled or disabled)

Ok

> * If the smart charging component is not active we do not have the enablement checkbox in the edit site area dialog.
>   As a result all site areas will be saved without smart charging and will pass the test.
>   So all that matters is that the projected properties are all the same within one site area tree (chain).
>   The different trees of the site can be seen independently.

Ok
Thanks a lot for the explanations!

OliveGerste:
@LucasBrazi06  Thanks a lot for the review. I implemented most of the changes, but need some additional input on some things:

- I am not completely sure about the renaming:
  - https://github.com/sap-labs-france/ev-server/pull/3062#discussion_r777597094 In fact I am building multiple trees and it returns an array of root site areas. So if there are 10 site areas in the input list, this could result in an array of 5 different root site areas (trees)
  
- Consistency in Site Area Tree: https://github.com/sap-labs-france/ev-server/pull/3062#discussion_r783796825
  - I think the last site area of a chain can only be ignored by excluding all charging stations on this site area from smart charging (Or using not supported ones), otherwise this will always exclude the whole power of the site area, which will result in almost no power left for other site areas even if no one is charging there.
  
- Circular dependency check in every tree build https://github.com/sap-labs-france/ev-server/pull/3062#discussion_r777577819:
  - We can check the constraint in every build, but I am not sure how to do the Error handling, because I need to specify the error type when catching it in UtilsService.  I have not found any reference in the server where this is done. I see two options:
    - Throwing a Backend Error and save the type in the message property --> check the type with switch case in Utilsservice
    - Introducing a new property like errorCode to the BackendError Type --> check the type with switch case in Utilsservice

Sorry for the long explanation. We can have a meeting if you think it is more effective.

LucasBrazi06:
> * [Sub site area enablement #3062 (comment)](https://github.com/sap-labs-france/ev-server/pull/3062#discussion_r777597094) In fact I am building multiple trees and it returns an array of root site areas. So if there are 10 site areas in the input list, this could result in an array of 5 different root site areas (trees)

Yes, this is the goal of the method: build a tree of Site Areas.
Next you need a new one to find the node (Site Area) you want in the tree.
And finally you check the Smart Charging consistency with a third one :-).
One method -> One task: so I can reuse the build of the site area tree outside the Smart Charging context for instance.

> * Consistency in Site Area Tree: [Sub site area enablement #3062 (comment)](https://github.com/sap-labs-france/ev-server/pull/3062#discussion_r783796825)

Okay, I understand, let's keep it simple then. No changes in Smart Charging constraints.

> * We can check the constraint in every build, but I am not sure how to do the Error handling, because I need to specify the error type when catching it in UtilsService.  I have not found any reference in the server where this is done.

You can throw a backend error when building the tree and in UtilsService you provide a specific try...catch around this method and return a specific error code to the UI?

OliveGerste:
> Yes, this is the goal of the method: build a tree of Site Areas. Next you need a new one to find the node (Site Area) you want in the tree. And finally you check the Smart Charging consistency with a third one :-). One method -> One task: so I can reuse the build of the site area tree outside the Smart Charging context for instance.

Building all the site area trees of one site refers to `Utils.buildSiteAreaTrees()` (We need this method to check consistency of all independent trees in one site)
This one should be renamed to `Utils.buildSiteAreaTree()`, but in fact it is not returning only one tree, but an array of trees 

Getting a specific tree refers to `Utils.buildSiteAreaTree()` Which first builds the trees and only returns the tree (rootSiteArea and all children) which contains the requested site area. I agree with renaming this to `findRootElementInSiteAreaTree()`

So according your comment building the tree and checking consistency should be completely separated? So that we have separate methods for this?

> You can throw a backend error when building the tree and in UtilsService you provide a specific try...catch around this method and return a specific error code to the UI?

Yes, my question was, which property of BackEndError I should use for saving the errorcode. 
When throwing the error in Utils, I could use e.g. the message property to save the error code.
After this I catch in UtilsService and determine the type by a switch case on the error message?

LucasBrazi06:
> Building all the site area trees of one site refers to `Utils.buildSiteAreaTrees()` (We need this method to check consistency of all independent trees in one site)
> This one should be renamed to `Utils.buildSiteAreaTree()`, but in fact it is not returning only one tree, but an array of trees

Ok to build Trees and name the method Utils.buildSiteAreaTrees() -> Build only an array Tree

> Getting a specific tree refers to `Utils.buildSiteAreaTree(siteAreas)` Which first builds the trees and only returns the tree (rootSiteArea and all children) which contains the requested site area. I agree with renaming this to `findRootElementInSiteAreaTree()`

I would rather name this method like findSiteAreaInTree(siteAreas);

> So according your comment building the tree and checking consistency should be completely separated? So that we have separate methods for this?

And a third one to check consistency like checkSiteAreaTree(siteArea).
So you have 3 methods with exactly one purpose each.
Separation of concerns they call it :-)

> Yes, my question was, which property of BackEndError I should use for saving the errorcode.
> When throwing the error in Utils, I could use e.g. the message property to save the error code.
> After this I catch in UtilsService and determine the type by a switch case on the error message?

I don't think we need a new property like 'errorCode' from the Backend exception.
We didn't need that concept so far but maybe I'm mistaken.
We can return a generic error in the UI like 'Inconsistencies occurred in the tree of Site Areas, check the logs for more details' or something more user friendly :-).
So, throwing a BackendError with a specific message to the Service layer which will log it and return always the same specific HTTP error code to the UI to display the message above would be sufficient I think.
And having to handle all the different error codes + translation will probably be a pain!
Anyway, to solve any issues with the Tree, the user will have to cancel his current change to go ahead.
What do you think?

OliveGerste:
> Ok to build Trees and name the method Utils.buildSiteAreaTrees() -> Build only an array Tree
> I would rather name this method like findSiteAreaInTree(siteAreas);
> And a third one to check consistency like checkSiteAreaTree(siteArea). So you have 3 methods with exactly one purpose each. Separation of concerns they call it :-)

Okay so :three: methods it will be :soon:

> I don't think we need a new property like 'errorCode' from the Backend exception. We didn't need that concept so far but maybe I'm mistaken. We can return a generic error in the UI like 'Inconsistencies occurred in the tree of Site Areas, check the logs for more details' or something more user friendly :-). So, throwing a BackendError with a specific message to the Service layer which will log it and return always the same specific HTTP error code to the UI to display the message above would be sufficient I think. And having to handle all the different error codes + translation will probably be a pain! Anyway, to solve any issues with the Tree, the user will have to cancel his current change to go ahead. What do you think?

All right. I am fine with this. The request for specific error messages in the front end came from a review from Claude I think. 
So I will just summarize these two codes:
 SITE_AREA_HIERARCHY_INCONSISTENCY_ERROR = 541,
 SITE_AREA_HIERARCHY_CIRCULAR_STRUCTURE_ERROR = 542,
to one code and send it to the front end displaying a generic error message.

src/storage/mongodb/SiteAreaStorage.ts
```diff
@@ -253,11 +277,43 @@ export default class SiteAreaStorage {
...
+      DatabaseUtils.pushSiteAreaLookupInAggregation({
+        tenantID: tenant.id, aggregation, localField: 'parentSiteAreaID', foreignField: '_id',
+        asField: 'parentSiteArea', oneToOneCardinality: true
+      });
+    }
     // Charging Stations
     if (params.withChargingStations || params.withOnlyChargingStations || params.withAvailableChargingStations) {
+      let pipelineMatch = {};
+      const additionalPipeline = [];
+      if (!Utils.isEmptyArray(params.chargingStationConnectorStatuses)) {
```
LucasBrazi06, line 291: I think that Charging Station's statuses should be done before the sort/skip/limit code like all the filters.

src/utils/Utils.ts
```diff
@@ -1607,4 +1608,95 @@ export default class Utils {
   public static hash(data: string): string {
     return crypto.createHash('sha256').update(data).digest('hex');
   }
+
+  public static findSiteAreaInTrees(siteAreaId: string, siteAreaTreeList: SiteArea[] = null, siteAreaTreeNode: SiteArea = null): SiteArea {
```
LucasBrazi06: I think that siteAreaTreeNode can be removed.
What about something like this (I haven't tested it)

```
  public static findSiteAreaInTrees(siteAreaID: string, siteAreaTrees: SiteArea[] = []): SiteArea {
    if (!Utils.isEmptyArray(siteAreaTrees)) {
      // Loop through trees to find specific tree
      for (const siteAreaTree of siteAreaTrees) {
        // Check node
        if (siteAreaTree.id === siteAreaID) {
          return siteAreaTree;
        }
        // Site Area ID is defined return tree, which contains the site area
        const requestedTree = this.findSiteAreaInTrees(siteAreaID, siteAreaTree.siteAreaChildren);
        if (requestedTree) {
          return siteAreaTree;
        }
      }
    }
  }

```

src/utils/Utils.ts
```diff
@@ -1607,4 +1608,92 @@ export default class Utils {
...
+        hashTable[siteArea.parentSiteAreaID].siteAreaChildren.push(hashTable[siteArea.id]);
+      } else {
+        // If no parent ID is defined push root site area to array
+        siteAreaTrees.push(hashTable[siteArea.id]);
+      }
+    }
+    return siteAreaTrees;
+  }
+
+  public static checkSiteAreaTrees(siteAreaTrees: SiteArea[], numberOfSiteAreas?: number,
```
LucasBrazi06, line 1663: Same comments as for method findSiteAreaInTrees(). Could be simplified.

src/utils/Utils.ts
```diff
@@ -1607,4 +1608,92 @@ export default class Utils {
...
+      if (!Utils.isEmptyArray(siteAreaTree.siteAreaChildren)) {
+        for (const siteAreaChild of siteAreaTree.siteAreaChildren) {
+          if (siteAreaTree.smartCharging !== siteAreaChild.smartCharging ||
+            siteAreaTree.numberOfPhases !== siteAreaChild.numberOfPhases ||
+            siteAreaTree.siteID !== siteAreaChild.siteID ||
+            siteAreaTree.voltage !== siteAreaChild.voltage) {
+            throw new BackendError({
+              siteAreaID: siteAreaChild.id,
+              method: 'checkSiteAreaTrees',
+              message: 'Property inconsistency in Site Area Tree',
```
LucasBrazi06, line 1677: Add detailedMessage to display what was wrong like:
`Expected numberOfPhases '3' from Site Area 'sss' but got '2' in Site Area 'ddd'`
Help yourself in prod :)

src/utils/Utils.ts
```diff
@@ -1607,4 +1608,92 @@ export default class Utils {
...
+        }
+      }
+      return count;
+    }
+    if (Array.isArray(siteAreaTrees)) {
+      // Loop through trees to check properties and count all site areas in it
+      for (const singleTree of siteAreaTrees) {
+        count += this.checkSiteAreaTrees(null, null, singleTree);
+      }
+      if (count !== numberOfSiteAreas) {
```
LucasBrazi06, line 1690: I would expect (count > numberOfSiteAreas) here and breaks the recursive loop.
How do you break the circular deps in one tree (A > B > C > A)?
I don't know if we have this use case.

LucasBrazi06 (changes requested):

src/utils/Utils.ts
OliveGerste: @LucasBrazi06 
Let me explain real quick:
The reason it works is the building process.
Here you see the build of a consistent tree list:

image
As you see in the last step the site area without a parent is pushed in the final list which is counted.

If you have a circular dependency the following happens:
image
The circular structure is not pushed to the final list, because there is no root. This will result in less site areas in total, then the source input

LucasBrazi06: Ok I understand, thanks for the explanation!
You are just checking about having no trees with no root element.

src/utils/Utils.ts
OliveGerste: This should work. 👍

LucasBrazi06 (changes requested):
A tree of Site Area should all have the same Site (check my comments)
Simplify and check this into account.
Check also the Unit Tests, there was a big change with Jest.
Thanks.

src/server/rest/v1/service/SiteAreaService.ts
```diff
@@ -354,26 +363,37 @@ export default class SiteAreaService {
...
-    siteArea.numberOfPhases = filteredRequest.numberOfPhases;
-    if (Utils.isComponentActiveFromToken(req.user, TenantComponents.OCPI)) {
-      if (Utils.objectHasProperty(filteredRequest, 'tariffID')) {
-        siteArea.tariffID = filteredRequest.tariffID;
-      }
-      if (!site.public) {
-        delete siteArea.tariffID;
-      }
+    if (siteArea.siteID !== filteredRequest.siteID) {
+      additionalSiteIDs.push(siteArea.siteID);
```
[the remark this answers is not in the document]
OliveGerste: @LucasBrazi06 Normally this should be the case already. The end user has to detach the site area first if he wants to change sites.
The reason for the array is that if you change the siteID of a parent, you also need to check if there are children left on the old site. So with the additional siteID, the trees from both sites (old and new) are build and checked for errors. If there is a child with another site id, an error is thrown. A better name for the array would probably be additionalSitesIDsToCheck.

We could also use the dependency method to check this, but this would mean another request to the db.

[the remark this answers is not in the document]
OliveGerste: @LucasBrazi06 Normally this should be the case already. The end user has to detach the site area first if he wants to change sites.
The reason for the array is that if you change the siteID of a parent, you also need to check if there are children left on the old site. So with the additional siteID, the trees from both sites (old and new) are build and checked for errors. If there is a child with another site id, an error is thrown. A better name for the array would probably be additionalSitesIDsToCheck.

We could also use the dependency method to check this, but this would mean another request to the db.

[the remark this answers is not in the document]
LucasBrazi06: > @LucasBrazi06 Normally this should be the case already. The end user has to detach the site area first if he wants to change sites. The reason for the array is that if you change the siteID of a parent, you also need to check if there are children left on the old site. So with the additional siteID, the trees from both sites (old and new) are build and checked for errors. If there is a child with another site id, an error is thrown. A better name for the array would probably be additionalSitesIDsToCheck.
> 
> We could also use the dependency method to check this, but this would mean another request to the db.

This is what I wrote, I think we should not allow to change site unless it's not part of the tree.
The tree should contain only Site Areas within the same Site always.

src/storage/mongodb/SiteAreaStorage.ts
```diff
@@ -280,6 +280,8 @@ export default class SiteAreaStorage {
         tenantID: tenant.id, aggregation, localField: '_id', foreignField: 'siteAreaID',
         asField: 'chargingStations', pipelineMatch
       }, additionalPipeline);
+      // Disable
+      params.withChargingStations = false;
```
OliveGerste, line 284: This is crashing the smart charging functionality, because on line 438 all the charging stations are deleted from the site area, if the parameter is set to false.

src/server/rest/v1/service/SiteAreaService.ts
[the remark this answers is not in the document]
OliveGerste: @LucasBrazi06 Yes this is how it has been handled since the beginning.
Let me explain in another way maybe. 
Let us assume we have an existing tree. Now someone wants to change the site of the parent. Then we also need to check if there are dependencies on the old site (existing children). And this is what additionalSites is doing.
Otherwise we would only check the new one, and there are no regressions of course. But on the old one we have children without a parent:
image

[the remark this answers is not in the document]
OliveGerste: I can change the way of checking and use the dependency method.

src/server/rest/v1/service/SiteAreaService.ts
```diff
@@ -354,26 +363,37 @@ export default class SiteAreaService {
...
     }
+    // Delete Charging Profiles
     let actionsResponse: ActionsResponse;
     if (siteArea.smartCharging && !filteredRequest.smartCharging) {
       actionsResponse = await OCPPUtils.clearAndDeleteChargingProfilesForSiteArea(
```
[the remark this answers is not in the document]
OliveGerste: @LucasBrazi06 
We are currently checking if all the profiles are cleared successfully after deactivating smart charging.
If a profile was not cleared, we tell the users to do it manually in the UI. 
I think this cannot be done when handling this async.

LucasBrazi06:
Closed, created a new branch

PR #3064, opened 2021-11-16
modified schema validator for pricing date from

cabch:
modified schema validator for pricing date from that changed to not mandatory


ClaudeROSSI:
already changed in a different branch

PR #3079, opened 2021-11-18
Create local cars when starting the server

mahdibenromdhane:


src/assets/cars/cars-definition.json
```diff
@@ -1,25 +1,90 @@
 [
   {
-    "Vehicle_ID": 100000,
-    "Vehicle_Make": "GOUPIL",
-    "Vehicle_Model": "G5 benne",
-    "Vehicle_Model_Version": "",
+  "Vehicle_ID": 100000,
```
LucasBrazi06, line 3: Tabulation not correct?

src/integration/car/local/LocalCarIntegration.ts
```diff
@@ -0,0 +1,235 @@
+import { CarCatalog, CarCatalogChargeAlternativeTable, CarCatalogChargeOptionTable, CarCatalogConverter } from '../../../types/Car';
```
LucasBrazi06, line 1: Not the right place, Car integration class must implement the interface which is not the case here.
Move this file to the newly created directory 'bootstrap' where I put the CS Template import.
Proposal for the naming: LocalCarCatalogBootstrap.uploadLocalCarCatalogsFromFile()

LucasBrazi06 (changes requested):
Small changes

src/assets/cars/cars-definition.json
mahdibenromdhane: Fixed

src/integration/car/local/LocalCarIntegration.ts
mahdibenromdhane: Done

LucasBrazi06 (approved):

PR #3089, opened 2021-11-19
pricing - Task MigrateSimplePricing

ClaudeROSSI:
IMPORTANT!
Should ONLY run when the new pricing definition UI gets released to all customers

The tariff name and description is not translated. The customer will have to change it manually!


src/migration/tasks/MigrateSimplePricing.ts
```diff
@@ -0,0 +1,66 @@
...
+import TenantStorage from '../../storage/mongodb/TenantStorage';
+import Utils from '../../utils/Utils';
+
+const MODULE_NAME = 'MigrateSimplePricing';
+
+export default class SimplePricingMigrationTask extends MigrationTask {
+  async migrate(): Promise<void> {
+    const tenants = await TenantStorage.getTenants({}, Constants.DB_PARAMS_MAX_LIMIT);
+    for (const tenant of tenants.result) {
+      await this.migrateTenant(tenant);
```
LucasBrazi06, line 20: You can also check if the simple pricing is active for the tenant, but the setting should suffice.

LucasBrazi06 (approved):

PR #3111, opened 2021-11-25
pricing - restrictions - checkTimeValidity

ClaudeROSSI:
pricing - restriction - capability to define a tariff for a particular time range


src/integration/pricing/ConsumptionPricer.ts
```diff
@@ -62,6 +63,25 @@ export default class ConsumptionPricer {
     return true;
   }
 
+
+  private checkTimeValidity(restrictions: PricingRestriction): boolean {
```
LucasBrazi06, line 67: We would need this in OCPI.
New method in Utils or somewhere else to check it?

LucasBrazi06 (approved):
Just one comment

PR #3124, opened 2021-11-29
pricing - consider transaction timezone 

ClaudeROSSI:
Some changes to make sure the location of the charging station is taken into account when checking "time range" restrictions while pricing the charging session consumption.


LucasBrazi06:
@ClaudeROSSI We should also have a timezone translation to display the time of the charging station in the UI.
Same problem in all the app ;-).

PR #3134, opened 2021-12-01
Fixed user's email case in query

Ninjeneer:


src/storage/mongodb/UserStorage.ts
```diff
@@ -602,7 +602,7 @@ export default class UserStorage {
     }
     // Email
     if (params.email) {
-      filters.email = params.email;
+      filters.email = { $regex: `^${params.email}$`, $options: 'i' };
```
LucasBrazi06, line 605: I don't like the regex with an email that could contain special chars like '.' or @.
Here I want the exact match.

LucasBrazi06 (changes requested):

Ninjeneer:
> Try to find an updateMany by reusing a ref of the email document that you will put in lowercase :-) So no need anymore to read the user. If you cannot find a way, you can run some JS script in the DB in V4.4 but I would prefer the first solution.

Makes sense, the read of users is useless
Done

LucasBrazi06 (approved):

PR #3143, opened 2021-12-03
pricing - fix - mochatest create context

ClaudeROSSI:
When the pricing module is ON
- A default tariff is created (price: 1 EUR/kWh) on all test tenants

This to make sure the test behaves as before when using the Simple Pricing logic.

Todo: cleanup some test helpers that are not used anymore !


ClaudeROSSI:
tests are green ... but the main tariff is created via the script: mochatest:createContext.

price: 1 Euro/kwh

Some tests have been updated accordingly.
It would make sense to have a different price and smarter tests to check rounding issues as well

PR #3146, opened 2021-12-06
Fixed tenant logo without Tenant's "LastChangedOn" property

Ninjeneer:
The Tenant's logo URL concatenation with "LastChangedOn" property fails if the tenant does not have the property. In addition, the parameter is not used in the `handleGetTenantLogo` method, so can be removed

### DIFF
--- a/src/storage/mongodb/CarStorage.ts
+++ b/src/storage/mongodb/CarStorage.ts
@@ -688,10 +688,10 @@ export default class CarStorage {
             $cond: {
               if: { $gt: ['$carCatalog.image', null] }, then: {
                 $concat: [
-                  `${Utils.buildRestServerURL()}/v1/util/CarCatalogImage?ID=`,
+                  `${Utils.buildRestServerURL()}/v1/util/car-catalogs/`,
                   '$carCatalog.id',
                   {
-                    $ifNull: [{ $concat: ['&LastChangedOn=', { $toString: '$carCatalog.lastChangedOn' }] }, ''] // Only concat 'lastChangedOn' if not null
+                    $ifNull: [{ $concat: ['?LastChangedOn=', { $toString: '$carCatalog.lastChangedOn' }] }, ''] // Only concat 'lastChangedOn' if not null
                   }
                 ]
               }, else: null
--- a/src/storage/mongodb/CarStorage.ts
+++ b/src/storage/mongodb/CarStorage.ts
@@ -108,10 +108,12 @@ export default class CarStorage {
             $cond: {
               if: { $gt: ['$image', null] }, then: {
                 $concat: [
-                  `${Utils.buildRestServerURL()}/client/util/CarCatalogImage?ID=`,
+                  `${Utils.buildRestServerURL()}/v1/util/car-catalogs/`,
                   { $toString: '$_id' },
-                  '&LastChangedOn=',
-                  { $toString: '$lastChangedOn' }
+                  '/image',
+                  {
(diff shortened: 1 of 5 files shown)


src/storage/mongodb/TenantStorage.ts
```diff
@@ -190,10 +190,8 @@ export default class TenantStorage {
         $addFields: {
           logo: {
             $concat: [
-              `${Utils.buildRestServerURL()}/client/util/TenantLogo?ID=`,
-              { $toString: '$_id' },
-              '&LastChangedOn=',
-              { $toString: '$lastChangedOn' }
+              `${Utils.buildRestServerURL()}/v1/util/tenants/logo?ID=`,
+              { $toString: '$_id' }
```
LucasBrazi06, line 194: The last change timestamp is necessary to force the browser to reload the cache when there is change in the logo ;-).

Ninjeneer: Ok ! Fixed to concatenation only if possible

src/storage/mongodb/TenantStorage.ts
```diff
@@ -190,10 +190,11 @@ export default class TenantStorage {
...
           logo: {
             $concat: [
-              `${Utils.buildRestServerURL()}/client/util/TenantLogo?ID=`,
+              `${Utils.buildRestServerURL()}/v1/util/tenants/logo?ID=`,
               { $toString: '$_id' },
-              '&LastChangedOn=',
-              { $toString: '$lastChangedOn' }
+              {
+                $ifNull: [{ $concat: ['&LastChangedOn=', { $toString: '$lastChangedOn' }] }, ''] // Only concat 'lastChangedOn' if not null
+              }
```
LucasBrazi06, line 197: Ok I got it.
Why not do the contrary, ensure that the tenant has always a lastChangedOn when changed at lest once?
If not, use the createdOn?

Ninjeneer: > ensure that the tenant has always a lastChangedOn when changed at lest once?

Isn't it already how it works ?

I don't really get the purpose of adding the 'createdOn' in the request : if the tenant has not been updated, then it is not necessary to force the browser cache reload with an extra property, right ?

LucasBrazi06: > Isn't it already how it works ?

Do not ask me the question, check it ;-)

> I don't really get the purpose of adding the 'createdOn' in the request : if the tenant has not been updated, then it is not necessary to force the browser cache reload with an extra property, right ?

I agree, just check that the logo is reloaded from the browser's on a second call when lastChanged is not provided.
You can also check the Cars, CarCatalogs (master tenant)
Thansk!

Ninjeneer: Done

LucasBrazi06 (approved):
Thanks!

src/storage/mongodb/CarStorage.ts
```diff
@@ -686,10 +688,11 @@ export default class CarStorage {
             $cond: {
               if: { $gt: ['$carCatalog.image', null] }, then: {
                 $concat: [
-                  `${Utils.buildRestServerURL()}/client/util/CarCatalogImage?ID=`,
+                  `${Utils.buildRestServerURL()}/v1/util/car-catalogs/`,
                   '$carCatalog.id',
-                  '&LastChangedOn=',
-                  { $toString: '$carCatalog.lastChangedOn' }
+                  {
```
LucasBrazi06, line 693: @Ninjeneer Missing '/image' in URL here.
@mahdibenromdhane found the issue.

Ninjeneer: Fixed in #3151

PR #3157, opened 2021-12-09
Refuse create/update unknown settings

Ninjeneer:
Also added missing `bakcupSensitiveData` attribute in every setting schemas


LucasBrazi06:
@ClaudeROSSI Could you check this one?
Thanks!

src/assets/schemas/setting/setting.json
```diff
@@ -16,6 +16,11 @@
       },
       "sanitize": "mongo"
     },
+    "backupSensitiveData": {
```
ClaudeROSSI, line 19: As far as I know **backupSensitiveData** is never used client-side.
I therefore do not see why it should be visible in our REST API.

From my perspective, we should just skip this information to make sure the POST/PUT request cannot override this critical information.
@LucasBrazi06 : Do we have a use-case where the backup of the sensitive data has to be manipulated thru a REST API?

src/assets/server/rest/v1/schemas/setting/setting-analytics-set.json
```diff
@@ -38,10 +38,13 @@
     "identifier": {
       "type": "string",
       "sanitize": "mongo",
-      "enum": ["sac"]
+      "enum": ["analytics"]
     },
     "sensitiveData": {
       "$ref": "setting#/definitions/sensitiveData"
+    },
+    "backupSensitiveData": {
```
ClaudeROSSI, line 46: same comment

src/assets/server/rest/v1/schemas/setting/setting-asset-set.json
```diff
@@ -143,6 +143,9 @@
     },
     "sensitiveData": {
       "$ref": "setting#/definitions/sensitiveData"
+    },
+    "backupSensitiveData": {
```
ClaudeROSSI, line 147: idem

src/assets/server/rest/v1/schemas/setting/setting-organization-set.json
```diff
@@ -0,0 +1,32 @@
...
+    },
+    "identifier": {
+      "type": "string",
+      "sanitize": "mongo",
+      "enum": ["organization"]
+    },
+    "sensitiveData": {
+      "$ref": "setting#/definitions/sensitiveData"
+    },
+    "backupSensitiveData": {
```
ClaudeROSSI, line 25: idem

src/server/rest/v1/service/SettingService.ts
```diff
@@ -151,7 +151,7 @@ export default class SettingService {
       });
     }
     // Filter
-    const filteredRequest = await SettingService.filterSetting(action, req, true);
+    const filteredRequest = SettingService.filterSetting(action, req) as SettingDB;
```
ClaudeROSSI, line 154: I do not think it make sense to cast as SettingDB here.
I would suggest that SettingService.**filterSetting** method has **SettingDB** as return type then.

src/server/rest/v1/service/SettingService.ts
```diff
@@ -173,7 +173,7 @@ export default class SettingService {
 
   public static async handleUpdateSetting(action: ServerAction, req: Request, res: Response, next: NextFunction): Promise<void> {
     UtilsService.assertIdIsProvided(action, req.body.id, MODULE_NAME, 'handleUpdateSetting', req.user);
-    const filteredRequest = await SettingService.filterSetting(action, req, false);
+    const filteredRequest = SettingService.filterSetting(action, req) as SettingDB;
```
ClaudeROSSI, line 176: idem

src/server/rest/v1/validator/SettingValidator.ts
```diff
@@ -84,4 +90,16 @@ export default class SettingValidator extends SchemaValidator {
   public validateSettingCarConnectorSetReq(data: Record<string, unknown>): HttpSettingCarConnectorSetRequest {
```
ClaudeROSSI, line 90: The return type should be SettingDB

src/server/rest/v1/validator/SettingValidator.ts
```diff
@@ -84,4 +90,16 @@ export default class SettingValidator extends SchemaValidator {
...
     return this.validate(this.settingCarConnectorSet, data);
   }
+
+  public validateSettingCarSetReq(data: Record<string, unknown>) {
```
ClaudeROSSI, line 94: As discussed on the phone the return type should be SettingDB for all methods.
And all HttpSetting*interface (added by Remi) are then from my perspective useless!
@LucasBrazi06 : We have new interfaces that are never used (such as **HttpSettingCarConnectorSetRequest**) ... I'm not sure why they have been added!

test/api/SettingTest.ts
```diff
@@ -540,28 +540,54 @@ describe('Setting', function() {
...
         expect(create.status).to.equal(StatusCodes.OK);
       });
       it('Should be able to update a setting', async () => {
-        const read = await testData.centralService.settingApi.readByIdentifier({ 'Identifier': 'test' });
-        read.data.content.type = 'updated';
-        const update = await testData.centralService.settingApi.update(read.data);
+        const read = await testData.centralService.settingApi.readByIdentifier({ 'Identifier': 'pricing' });
+        const oldPricingType = read.data.content.type;
+        read.data.content.type = 'simple';
+        let update = await testData.centralService.settingApi.update(read.data);
```
ClaudeROSSI, line 559: This update does nothing because the type si by default set to 'simple' when using the **utall** tenant.
do you agree?

ClaudeROSSI (changes requested):
I have some comments.

Please let me know what you think.
I won't push the changes that we discussed and that I did on my side.

LucasBrazi06:
@ClaudeROSSI Need your approval

LucasBrazi06:
Moved to dev-cross-review

LucasBrazi06:
@ClaudeROSSI My bad, made mistake on the wrong pipeline :-(, I just put in back in requested change.

test/api/SettingTest.ts
Ninjeneer: Yep ! 
Switched from 'simple' to 'convergentCharging' for test

src/assets/schemas/setting/setting.json
Ninjeneer: You're right, it isn't sent from the client side. 
I removed every "backupSensitiveData" in schemas

src/server/rest/v1/service/SettingService.ts
Ninjeneer: Fixed, same for below comments

ClaudeROSSI (approved):
Approved.

But I think we can have get rid of some interfaces which are not used anymore!
What do you think?

image

Ninjeneer:
> Approved.
> 
> But I think we can have get rid of some interfaces which are not used anymore! What do you think?
> 
You're right, I removed it. We can still retrieve it from the Git history in case we need it later.

PR #3163, opened 2021-12-13
REST > Billing getTaxes endpoint/doc + get/update settings doc

Ninjeneer:


LucasBrazi06:
@ClaudeROSSI Could you check this one?
Thanks!

src/assets/server/rest/v1/docs/e-mobility-oas.json
```diff
@@ -12999,6 +12999,218 @@
...
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "id": {
+                      "$ref": "#/components/schemas/Setting/properties/id"
+                    },
+                    "identifier": {
+                      "$ref": "#/components/schemas/Setting/properties/identifier"
+                    },
+                    "sensitiveData": {
```
ClaudeROSSI, line 13083: do we really need to expose this information?

ClaudeROSSI (approved):
approved.

But a comment on the REST endpoint doc!

ClaudeROSSI:
@LucasBrazi06 : Approved but I do not understandf why we put "sensitiveData" in the rest endpoint documentation

src/assets/server/rest/v1/docs/e-mobility-oas.json
LucasBrazi06: If the sensitiveData is driven by the backend we should remove it otherwise, we keep it until we have dedicated endpoints for the settings.

PR #3166, opened 2021-12-13
Registration token new auth migration

LucasBrazi06:


LucasBrazi06:
@mahdibenromdhane Could you check with @Ramzay the migration of the Registration Token.
Also check with Postman.
Thanks!

src/authorization/AuthorizationsDefinition.ts
```diff
@@ -346,8 +348,21 @@ export const AUTHORIZATION_DEFINITION: AuthorizationDefinition = {
...
-      { resource: Entity.TOKEN, action: [Action.CREATE, Action.READ, Action.UPDATE, Action.DELETE] },
+      {
+        resource: Entity.REGISTRATION_TOKEN, action: Action.LIST,
+        attributes: [
+          'id', 'status', 'description', 'createdOn', 'lastChangedOn', 'expirationDate', 'revocationDate',
+          'siteAreaID', 'siteArea.name', 'createdBy.name', 'createdBy.firstName', 'lastChangedBy.name', 'lastChangedBy.firstName'
+        ]
+      },
+      {
+        resource: Entity.REGISTRATION_TOKEN, action: [Action.READ, Action.CREATE, Action.UPDATE],
```
mahdibenromdhane, line 359: Can an admin create a registration token without a site Area ?

src/authorization/AuthorizationsDefinition.ts
```diff
@@ -1167,11 +1157,62 @@ export const AUTHORIZATION_DEFINITION: AuthorizationDefinition = {
...
+          'id', 'status', 'description', 'createdOn', 'lastChangedOn', 'expirationDate', 'revocationDate',
+          'siteAreaID', 'siteArea.name', 'createdBy.name', 'createdBy.firstName', 'lastChangedBy.name', 'lastChangedBy.firstName'
+        ]
+      },
+      {
+        resource: Entity.REGISTRATION_TOKEN, action: [Action.CREATE, Action.UPDATE],
+        condition: {
+          Fn: 'custom:dynamicAuthorizations',
+          args: {
+            asserts: ['SiteAreaMandatory'],
```
mahdibenromdhane, line 1198: Site Admin can create registration tokens of site different from site in which his not site Admin via Postman.

src/authorization/AuthorizationsDefinition.ts
LucasBrazi06: No, the Site Area is mandatory for Site Admin

src/authorization/AuthorizationsDefinition.ts
LucasBrazi06: Good catch!
Feel free to provide a fix ;-)

LucasBrazi06: Normally, the code below should prevent it:

Ramzay: @LucasBrazi06  I think the issue comes from the Action.READ which is not restrictive enough, since a **Site Admin** can read / see all site areas associated to his sites / organisation. 
And as a consequence if we provide in the Postman request a site area that is in his site / organisation but NOT in his assigned site, the filter in checkAndGetSiteAreaAuthorization won't work and the call will succeed !
To confirm my theory I used Action.READ_ASSETS_FROM_SITE_AREA and after the modification the behavior works as expected since the action has a SiteAdmin filter

mahdibenromdhane (approved):

PR #3167, opened 2021-12-14
REST > getUserPayments / setup-delete payment methods doc + schemas

Ninjeneer:


src/assets/server/rest/v1/schemas/billing/billing-delete-payment-method.json
```diff
@@ -0,0 +1,17 @@
+{
+  "$id": "billing-delete-payment-method",
+  "title": "Delete user payment method Request",
+  "type": "object",
+  "properties": {
+    "userID": {
+      "$ref": "common#/definitions/id"
+    },
+    "paymentMethodId": {
+      "$ref": "common#/definitions/id"
```
ClaudeROSSI, line 10: This is wrong.
This ID is in a format which is specific to the BILLING service provider ( i.e.: in a STRIPE-specific format).
Make sure to be more flexible here!
Suggestion:
      **"type": "string",
      "sanitize": "mongo"**

I assume we cannot remove "payment methods" anymore if we keep that definition!

src/assets/server/rest/v1/schemas/billing/billing-setup-payment-method.json
```diff
@@ -0,0 +1,17 @@
...
+  "title": "Setup user payment method Request",
+  "type": "object",
+  "properties": {
+    "userID": {
+      "$ref": "common#/definitions/id"
+    },
+    "paymentMethodId": {
+      "type": "string",
+      "sanitize": "mongo"
+    }
```
ClaudeROSSI, line 12: This one is correct!

ClaudeROSSI (changes requested):
small mistake on DELETE payment methods

src/assets/server/rest/v1/schemas/billing/billing-delete-payment-method.json
Ninjeneer: Indeed, fixed.

ClaudeROSSI (approved):
approved

PR #3168, opened 2021-12-15
Pricing restrictions

cabch:


ClaudeROSSI:
This is not correct.
@LucasBrazi06 : The pricing engine expects numbers.
@cabch : put it as dev review first.
Can you revert please!

PR #3174, opened 2021-12-16
repair sessions priced at zero

ClaudeROSSI:
repair v2 for sessions priced a zero and lasting more than 48h hours


ClaudeROSSI:
this PR only repairs old sessions (the one using the simple pricing).

Session at zero after that are a different issue ... under investigation.

image

PR #3183, opened 2021-12-17
Fix site create car authorization

Ramzay:


src/authorization/AuthorizationsDefinition.ts
```diff
@@ -1084,7 +1084,15 @@ export const AUTHORIZATION_DEFINITION: AuthorizationDefinition = {
           'user.id', 'user.name', 'user.firstName', 'userID'
         ],
       },
-      { resource: Entity.CAR, action: Action.CREATE },
+      { resource: Entity.CAR, action: Action.CREATE,
+        condition: {
+          Fn: 'custom:dynamicAuthorizations',
+          args: {
+            asserts: [],
+            filters: ['SitesAdmin']
```
LucasBrazi06, line 1092: SitesAdmin should return a filter on siteIDs which will used in the MongoDB request but in the Create action, there is no object to retrieve from MongoDB like Update.
So how this filter will make the request failing?
Thanks!

Ramzay: Without this modification the CREATE action was using the OwnUserDynamicAuthorizationFilter which only allowed the SiteAdmin to create a car for himself. This modification allowed a SiteAdmin to create cars for his users.
There is a check performed later in the code on the userID, if the userID is not assigned to the site admin the creation will fail.

Ramzay: @LucasBrazi06 I removed the SiteAdmin filter as it is not necessary here, just as you stated.
Please note that we added this Authorization definition to override the one that is defined in the Basic User that only allowed to create a Car for themselves.

PR #3199, opened 2021-12-21
Allow to select no car when starting a transaction

mahdibenromdhane:


LucasBrazi06 (changes requested):
I would expect that if the carID is not provided we force the car to not be selected.
The remoteStart flag is not necessary.
I think the line 1777
`if (filteredRequest.carID && filteredRequest.carID !== user.lastSelectedCarID) {`
Should be
`if (filteredRequest.carID) {`
Do you agree?

mahdibenromdhane:
> I would expect that if the carID is not provided we force the car to not be selected. The remoteStart flag is not necessary. I think the line 1777 `if (filteredRequest.carID && filteredRequest.carID !== user.lastSelectedCarID) {` Should be `if (filteredRequest.carID) {` Do you agree?

If the user start a transaction using his badge directly from the charging station, we se the default car in the transaction.
image

The flag will prevent the processTransactionCar() from setting the default car if the session is started remotely from the dashboard or the mobile by setting the lastSelectedCar flag to true.

LucasBrazi06:
> The flag will prevent the processTransactionCar() from setting the default car if the session is started remotely from the dashboard or the mobile by setting the lastSelectedCar flag to true.

This is how it is designed to work :-), then why do you introduce a new flag 'remoteStart' in this PR?

mahdibenromdhane:
> > The flag will prevent the processTransactionCar() from setting the default car if the session is started remotely from the dashboard or the mobile by setting the lastSelectedCar flag to true.
> 
> This is how it is designed to work :-), then why do you introduce a new flag 'remoteStart' in this PR?

It's for allowing the user to start a transaction without a car (now we set the default if the user set the input field of the car empty in the start transaction dialog in the UI).
 
But I'v just realised that setting te lastSelectedCar to true without setting lastSelectedCarID will create a regression and we won't set the default car if the user start a session with his badge next time. 
I will try to find another way to solve the issue.

LucasBrazi06:
> But I'v just realised that setting te lastSelectedCar to true without setting lastSelectedCarID will create a regression and we won't set the default car if the user start a session with his badge next time.

We can discuss it tomorrow :-)

LucasBrazi06 (approved):
Do not apply eslint formatting for the whole file, it's making too many changes when I review your PR.
Thanks.

LucasBrazi06:
> Do not apply eslint formatting for the whole file, it's making too many changes when I review your PR. Thanks.

Your eslint changes created conflicts that I had to deal with grrr ;-)

mahdibenromdhane:
> > Do not apply eslint formatting for the whole file, it's making too many changes when I review your PR. Thanks.
> 
> Your eslint changes created conflicts that I had to deal with grrr ;-)

Okey, I will format only my changes next time :)

PR #3209, opened 2021-12-28
Pricing - extended time restrictions

ClaudeROSSI:
**Pricing - Time restrictions**

The two following time restrictions examples are now valid:

- from 08:00 to 20:00 
- from **20:00 to 08:00** 

You may notice that the second definitions **spans two days** ... this was not supported by the previous implementation!

**Pricing - Parking Time Withdrawal**

The parking time logic as been improved to make sure we do not bill the parking time when we have gaps in the energy consumption. Tests have been adapted to check this specific situation - the parking time is withdrawn when the energy consumption starts again.

Here is an example:


LucasBrazi06:
@ClaudeROSSI You cannot rely on gap between 2 consumptions.
The end-user may have programmed his VE to stop at 80% and stays for 30 mins in parking time and next the charging session resumes and in that case you should not remove the parking time.
We have also no consumption when applying charging profiles and we do not count the parking time in that case in the consumption metric.

ClaudeROSSI:
> @ClaudeROSSI You cannot rely on gap between 2 consumptions. The end-user may have programmed his VE to stop at 80% and stays for 30 mins in parking time and next the charging session resumes and in that case you should not remove the parking time. We have also no consumption when applying charging profiles and we do not count the parking time in that case in the consumption metric.

let's discuss this tomorrow!

ClaudeROSSI:
Ok ... closing this PR. 
I'll split this in two different pull requets.

PR #3230, opened 2022-01-06
Pricing - price extra inactivity

ClaudeROSSI:
**pricing of the extra inactivity**
- billing not done on STOP transaction anymore

The Transaction.END event is now used:
- To price the extra inactivity 
- To trigger the invoice generation


ClaudeROSSI:
I have some doubts about this PR. The changes are very sensitive to the connector status and to the corresponding OCPP status notifications. 
What happens if we restart some server processes while transactions are in progress? 

**Billing changes** are the ones below:
1) the billing module does not trigger the billing task on a STOP transation anymore.
It now simply puts the transaction.billingData.stop.status to **Billing.PENDING**!
2) the actual billing async task is now only created when the extra inactivity is computed.
The billing module expects to be called with a **Transaction.END** action (and not **Transaction.STOP**). 
3) Checks have been added to make sure we do not trigger this twice
4) billing data thresholds are only checked during the STOP transaction

Pricing changes:
1) A final call to the PricingEngine is made when computing the extra inactivity.
Method: OCPPUtils.**checkAndUpdateLastCompletedTransaction**() 
2) PricingEngine will create chunks to properly price the final consumption but this is part of another PR.

LucasBrazi06:
> What happens if we restart some server processes while transactions are in progress?

The charger should resume the meter values which it was not able to send.

LucasBrazi06:
> I have some doubts about this PR. The changes are very sensitive to the connector status and to the corresponding OCPP status notifications.

We also need to add a Job when no Status Notification have been sent. Like the CDR.

ClaudeROSSI:
> > I have some doubts about this PR. The changes are very sensitive to the connector status and to the corresponding OCPP status notifications.
> 
> We also need to add a Job when no Status Notification have been sent. Like the CDR.

right

src/scheduler/tasks/BillPendingTransactionTask.ts
```diff
@@ -0,0 +1,152 @@
...
+            // Get all finished Transaction with no CDR
+              const transactionsMDB: {_id: number}[] = await global.database.getCollection<{_id: number}>(tenant.id, 'transactions')
+                .aggregate<{_id: number}>(
+                [
+                  {
+                    // Sessions with a PENDING billing status and with a unknown extra inactivity
+                    $match: {
+                      'stop': { $exists: true },
+                      'stop.extraInactivityComputed': false,
+                      'billingData.stop.status': BillingStatus.PENDING
```
ClaudeROSSI, line 40: @mahdibenromdhane : as discussed, my job is looking for transactions where the Transaction.STOP has been received but the extra inactivity is unknown (extraInactivityComputed===false). In other words, the OCPP notification regarding connector the connector availability has not been processed.
In that case I need to price the last consumption and trigger the billing.
Your use-case where you need to stop "ghost sessions" is from my perspective similar (and needs to be priced as well). We therefore need to be in sync regarding the logic on how to detect old transactions. Maybe a single job would be sufficient for both use-cases.

src/scheduler/tasks/BillPendingTransactionTask.ts
```diff
@@ -0,0 +1,152 @@
...
+                          message: `Charging Station '${transaction.chargeBoxID}' not found`,
+                        });
+                        continue;
+                      }
+                      // Check for the last transaction
+                      const lastTransaction = await TransactionStorage.getLastTransactionFromChargingStation(tenant, transaction.chargeBoxID, transaction.connectorId);
+                      if (transaction.id === lastTransaction?.id) {
+                        // Avoid conflict with a session which is still in progress
+                        const connector = Utils.getConnectorFromID(chargingStation, transaction.connectorId);
+                        if (connector.status !== ChargePointStatus.AVAILABLE) {
```
LucasBrazi06, line 85: Do not check the connector's status but it's transaction ID:
`      if (connector.currentTransactionID === transaction.id) {`

src/scheduler/tasks/BillPendingTransactionTask.ts
```diff
@@ -0,0 +1,152 @@
...
+                          // Do nothing - connector is being used
+                          continue;
+                        }
```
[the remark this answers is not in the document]
LucasBrazi06: Nothing to do with closing an opened transaction

src/scheduler/tasks/BillPendingTransactionTask.ts
```diff
@@ -0,0 +1,152 @@
...
+                    } catch (error) {
+                      await Logging.logError({
+                        tenantID: tenant.id,
+                        action: ServerAction.BILLING_BILL_PENDING_TRANSACTION,
+                        module: MODULE_NAME, method: 'processTenant',
+                        message: `Failed to bill pending transaction '${transactionMDB._id}'`,
+                        detailedMessages: { error: error.stack, transaction: transactionMDB }
+                      });
+                    } finally {
+                    // Release the lock
```
LucasBrazi06, line 134: Wrong Indentation (WI)

src/scheduler/tasks/BillPendingTransactionTask.ts
```diff
@@ -0,0 +1,152 @@
...
+                      await LockingManager.release(transactionLock);
+                    }
+                  }
+                }
+              }
+            } finally {
+            // Release the lock
```
LucasBrazi06, line 141: WI

src/server/ocpp/services/OCPPService.ts
```diff
@@ -799,16 +799,29 @@ export default class OCPPService {
...
-                await Logging.logInfo({
-                  tenantID: tenant.id,
-                  ...LoggingHelper.getChargingStationProperties(chargingStation),
-                  user: lastTransaction.userID,
-                  module: MODULE_NAME, method: 'checkAndUpdateLastCompletedTransaction',
-                  action: ServerAction.EXTRA_INACTIVITY,
-                  message: `${Utils.buildConnectorInfo(lastTransaction.connectorId, lastTransaction.id)} Extra Inactivity of ${lastTransaction.stop.extraInactivitySecs} secs has been added`,
-                  detailedMessages: { statusNotification, connector, lastTransaction }
-                });
+                const lastConsumption = await OCPPUtils.buildExtraConsumptionInactivity(tenant, chargingStation, lastTransaction);
```
LucasBrazi06, line 802: Just after the IF statement, is it possible to build methods like this.buildLastTransactionExtraInactivity(...) and put all that code in it to improve readability?

src/server/ocpp/services/OCPPService.ts
```diff
@@ -824,6 +837,8 @@ export default class OCPPService {
             }
             // Flag
             lastTransaction.stop.extraInactivityComputed = true;
+            // Billing - We now know the extra inactivity - the invoice can be generated
+            await OCPPUtils.processTransactionBilling(tenant, lastTransaction, TransactionAction.END);
```
LucasBrazi06, line 841: Possible to put this call just before the first CDR call in line 845?

LucasBrazi06 (changes requested):
Check my comments

src/scheduler/tasks/BillPendingTransactionTask.ts
```diff
@@ -0,0 +1,152 @@
...
+                          action: ServerAction.BILLING_BILL_PENDING_TRANSACTION,
+                          module: MODULE_NAME, method: 'processTenant',
+                          message: `Charging Station '${transaction.chargeBoxID}' not found`,
+                        });
+                        continue;
+                      }
+                      // Check for the last transaction
+                      const lastTransaction = await TransactionStorage.getLastTransactionFromChargingStation(tenant, transaction.chargeBoxID, transaction.connectorId);
+                      if (transaction.id === lastTransaction?.id) {
+                        // Avoid conflict with a session which is still in progress
```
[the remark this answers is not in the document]
ClaudeROSSI: > Not possible as on going transaction does not have the 'stop' property.

But here I'm trying to detect the situation where the stop has been received but the car is still plugged to the charging station. So for me the transaction is still on-going because the extra inactivity is not yet known.
Do you agree that I must handle it here?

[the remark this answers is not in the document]
LucasBrazi06: Okay I got it, so it's fine!

src/server/ocpp/services/OCPPService.ts
ClaudeROSSI: > Possible to put this call just before the first CDR call in line 845?

I do not feel confortable. 
If we move it we then take the risk to call several times the billing process.
Are we sure "**checkAndUpdateLastCompletedTransaction**" is never called twice?

LucasBrazi06: Be sure that you handle this use case in the billing ;-)

ClaudeROSSI: > Be sure that you handle this use case in the billing ;-)

I billing handles it. It throws an exception. What about the calling layer? 
What shall we do if the billing rejects a second call?

src/scheduler/tasks/BillPendingTransactionTask.ts
ClaudeROSSI: > Yes, only one job should trigger calculation of non existing extra inactivity that will trigger the CDR and the Billing. Mahdi's job about Ghost sessions is another use case, it's based on non existing 'stop' property and will not trigger the extra inactivity calculation if I'm not wrong.

You're right, Mehdi's job won't trigger the extra inactivity calculation. But then, if we stop the transaction without simulating the OCPP call changing the connector status, we will then have a transaction with no invoice! Right?

src/server/ocpp/services/OCPPService.ts
LucasBrazi06: If it throws an exception during a Status Notification it has no effect on the OCPP return call (always return an empty response).
In the end you will have only a log in error.
Whether add a try...catch and log a warning in the caller or remove the exception throw in the billing and log a warning.
As you want.

ClaudeROSSI: > If it throws an exception during a Status Notification it has no effect on the OCPP return call (always return an empty response). In the end you will have only a log in error. Whether add a try...catch and log a warning in the caller or remove the exception throw in the billing and log a warning. As you want.

OK - I'll remove the exceptions - and log warnings when suspicious calls are detected on the billing side

ClaudeROSSI: > > If it throws an exception during a Status Notification it has no effect on the OCPP return call (always return an empty response). In the end you will have only a log in error. Whether add a try...catch and log a warning in the caller or remove the exception throw in the billing and log a warning. As you want.
> 
> OK - I'll remove the exceptions - and log warnings when suspicious calls are detected on the billing side

well ... this is not possible! please call me!

src/scheduler/tasks/BillPendingTransactionTask.ts
LucasBrazi06: Indeed, but your job's next run should see it and bill it ;-)

src/server/ocpp/services/OCPPService.ts
LucasBrazi06: We can discuss this tomorrow onsite ;-)

src/scheduler/tasks/BillPendingTransactionTask.ts
ClaudeROSSI: > Indeed, but your job's next run should see it and bill it ;-)

Right! This is the "cat pattern" ... cats always land on their feet lol

src/integration/pricing/sap-convergent-charging/SapConvergentChargingPricingIntegration.ts
```diff
@@ -112,6 +112,32 @@ export default class SapConvergentChargingPricingIntegration extends PricingInte
...
+        currencyCode: rateResult.transactionsToConfirm.getCurrencyCode(),
+        amount: rateResult.transactionsToConfirm.getTotalUnroundedAmount(),
+        pricingSource: PricingSource.CONVERGENT_CHARGING
+      };
+    }
+    await this.handleError(transaction, ServerAction.UPDATE_TRANSACTION, consumptionData, result);
+    return null;
+  }
+
+  public async endSession(transaction: Transaction, consumptionData: Consumption): Promise<PricedConsumption> {
```
ClaudeROSSI, line 140: @LucasBrazi06 : Convergent Charging - endSession has been added but not tested!

src/integration/pricing/PricingIntegration.ts
```diff
@@ -23,4 +23,6 @@ export default abstract class PricingIntegration<T extends PricingSetting> {
   public abstract updateSession(transaction: Transaction, consumptionData: Consumption, chargingStation?: ChargingStation): Promise<PricedConsumption>;
 
   public abstract stopSession(transaction: Transaction, consumptionData: Consumption, chargingStation?: ChargingStation): Promise<PricedConsumption>;
+
+  public abstract endSession(transaction: Transaction, consumptionData: Consumption, chargingStation?: ChargingStation): Promise<PricedConsumption>;
```
ClaudeROSSI, line 27: @LucasBrazi06 : I changed the pricing API to have both the stopSession, endSession method!

src/server/ocpp/services/OCPPService.ts
ClaudeROSSI: done

src/scheduler/tasks/BillPendingTransactionTask.ts
ClaudeROSSI: ok

src/scheduler/tasks/BillPendingTransactionTask.ts
ClaudeROSSI: ok

src/scheduler/tasks/BillPendingTransactionTask.ts
```diff
@@ -0,0 +1,152 @@
...
+                        await Logging.logError({
+                          tenantID: tenant.id,
+                          action: ServerAction.BILLING_BILL_PENDING_TRANSACTION,
+                          module: MODULE_NAME, method: 'processTenant',
+                          message: `Transaction '${transactionMDB._id}' not found`,
+                        });
+                        continue;
+                      }
+                      // Get Charging Station
+                      const chargingStation = await ChargingStationStorage.getChargingStation(tenant, transaction.chargeBoxID);
```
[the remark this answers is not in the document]
ClaudeROSSI: done

ClaudeROSSI:
All tests are green!
Please review again ;) ... thanks

ClaudeROSSI:
Some Transaction tests are failing because of the chunk logic used when pricing.

Exemple below: 2 sessions of one hour, consuming only 1 kWh.
This generates 60 chunks with a very low consumption.
The accumulated amount (priced at 1 euro per kWh) loses some precision.

Here is the result: **totalPrice** is not the expected one!



I'm not sure we can fix that one!

LucasBrazi06:
> I'm not sure we can fix that one!

Oups!
Is this a real use case?

ClaudeROSSI:
> > I'm not sure we can fix that one!
> 
> Oups! Is this a real use case?

Well ... real use-case or not, the root cause is the fact that we do not preserve the precision when storing the data in the DB.
c.f.: below an example reproducing the root cause:
- **totalPriceAsDecimal** is correct
- **totalPrice** is wrong and corresponds to what we do (since we convert to number the persisted data)

image

I'm investigation on how to persist the accumulated amount with its 'full' precision!

LucasBrazi06:
> I'm investigating on how to **persist the accumulated amount** with its 'full' precision!

Okay.

ClaudeROSSI:
> > > I'm not sure we can fix that one!
> > 
> > 
> > Oups! Is this a real use case?
> 
> Well ... real use-case or not, the root cause is the fact that we do not preserve the precision when storing the data in the DB. I'm investigating on how to **persist the accumulated amount** with its 'full' precision!
> 
> c.f.: below an example reproducing the root cause:
> 
> * **totalPriceAsDecimal** is correct
> * **totalPrice** is wrong and corresponds to what we do (since we convert to number the persisted data)
> 
> image

Problem solved by serializing the Decimal object instead of converting it to a number.
The Decimal object can be serialized and de-serialized thanks to some undocumented capabilities of the Decimal.js module.

LucasBrazi06 (approved):

PR #3245, opened 2022-01-17
Use average per asset per minute before summing to site area level while displaying site area chart component

RohanAntony:
Test cases: 
- Check if the site area consumption chart aggregates per minute per asset before summing up to site area aggregation. 
  1. Manually add additional consumption to a site area asset for a given minute in Consumptions.
  2. Check if the average of the instantWatts is calculated for the given minute instead of summing the instantWatts within a minute.
- Check if the axis on site area consumption chart shows 'Consumption by Site Area'
- Check if the axis on site area consumption chart shows the values with 'A' or 'kW'
- Check if the negative axis is shown indicating production values.

Fixes:
- Uses non facet, only grouping based mongo db aggregation pipeline methods to aggregate.

Sample:
image

[ev-dashboard changes](https://github.com/sap-labs-france/ev-dashboard/pull/2919)


LucasBrazi06:
@OliveGerste Could you check this PR?
I'll move it to Cross review pipeline
Thanks!

PR #3268, opened 2022-01-20
Use storage to retrieve assigned sites

Ramzay:
- Call storage to retrieve assigned site ID
- Not rely on 'static 'user token
- remove unused method


Ramzay:
> Authorization.ts should not depend on AuthorizationService.ts (circular dep) What did you want to achieve here?

Hello @LucasBrazi06,
Goal was to not duplicate the functions that retrieved the logged user assigned sites but since it creates a circular dependency it's not optimal :/ In the future I will try to detect them before submitting a PR !
The PR in essence is to call the storage when we want to retrieve the logged user assigned site instead of relying on the usertoken. Which fixes the bug users are facing.

LucasBrazi06:
> Goal was to not duplicate the functions that retrieved the logged user assigned sites but since it creates a circular dependency it's not optimal :/ In the future I will try to detect them before submitting a PR !

No problem, I added a golden rule for you :-)



> The PR in essence is to call the storage when we want to retrieve the logged user assigned site instead of relying on the usertoken. Which fixes the bug users are facing.

Ok, I understand, I let you see if there is another way to do it without the circular dep.

LucasBrazi06 (changes requested):
I would have used the existing data source from the DynamicAuthorizationFactory class which returns you already the list Site IDs:

```
      const siteAdminDataSource = await DynamicAuthorizationFactory.getDynamicDataSource(
        req.tenant, req.user, DynamicAuthorizationDataSourceName.SITES_ADMIN) as SitesAdminDynamicAuthorizationDataSource;
      const siteAdminsIDs = siteAdminDataSource.getData();
```
Let me know.

Ramzay:
> I would have used the existing data source from the DynamicAuthorizationFactory class which returns you already the list Site IDs:

You are absolutely right, the DynamicAuthorizationFactory performs the exact same storage call as the ones I moved from AuthorizationService.ts. Let me know if you see anything else !

LucasBrazi06 (approved):
Ok for the quick fix considering that this will not be necessary when ChargingStation/Transaction Service will be migrated to the new auth framework.

PR #3279, opened 2022-01-24
REST > Settings validation schemas

Ninjeneer:


src/assets/server/rest/v1/schemas/setting/settings-get.json
```diff
@@ -0,0 +1,33 @@
...
+  "title": "Settings get Request",
+  "description": "",
+  "type": "object",
+  "properties": {
+    "Identifier": {
+      "$ref": "setting#/definitions/identifier"
+    },
+    "ContentFilter": {
+      "type": "boolean",
+      "sanitize": "mongo"
```
ClaudeROSSI, line 12: @Ninjeneer : It seems that ContentFilter is not used at all. Can you double check?

ClaudeROSSI (approved):
just a minor comment

src/assets/server/rest/v1/schemas/setting/settings-get.json
Ninjeneer: Indeed, it was useless, removed

PR #3293, opened 2022-01-27
billing - check whether access control is ON

ClaudeROSSI:
A lot of unexpected problems with that one.
This is a DRAFT - with a lot of stuff to clarify (marked with TODO comments)
@LucasBrazi06 : As discussed ... I need your advise for this PR

### DIFF
--- a/src/integration/billing/BillingIntegration.ts
+++ b/src/integration/billing/BillingIntegration.ts
@@ -1,4 +1,4 @@
-import { BillingChargeInvoiceAction, BillingDataTransactionStart, BillingDataTransactionStop, BillingDataTransactionUpdate, BillingInvoice, BillingInvoiceItem, BillingInvoiceStatus, BillingOperationResult, BillingPaymentMethod, BillingStatus, BillingTax, BillingUser, BillingUserSynchronizeAction } from '../../types/Billing';
+import { BillingChargeInvoiceAction, BillingDataTransactionStart, BillingDataTransactionStop, BillingDataTransactionUpdate, BillingInvoice, BillingInvoiceItem, BillingInvoiceStatus, BillingOperationResult, BillingPaymentMethod, BillingStatus, BillingTax, BillingUser } from '../../types/Billing';
 import Tenant, { TenantComponents } from '../../types/Tenant';
 import Transaction, { StartTransactionErrorCode } from '../../types/Transaction';
 import User, { UserStatus } from '../../types/User';
@@ -310,6 +310,44 @@ export default abstract class BillingIntegration {
         });
       }
     }));
+  }
+
+  protected async checkBillingDataThreshold(transaction: Transaction): Promise<boolean> {
+    // Do not bill suspicious StopTransaction events
+    if (!Utils.isDevelopmentEnv()) {
+    // Suspicious StopTransaction may occur after a 'Housing temperature approaching limit' error on some charging stations
+      const timeSpent = this.computeTimeSpentInSeconds(transaction);
+      // TODO - make it part of the pricing or billing settings!
+      if (timeSpent < 60 /* seconds */ || transaction.stop.totalConsumptionWh < 1000 /* 1kWh */) {
+        await Logging.logWarning({
+          ...LoggingHelper.getTransactionProperties(transaction),
+          tenantID: this.tenant.id,
+          user: transaction.userID,
+          action: ServerAction.BILLING_TRANSACTION,
+          module: MODULE_NAME, method: 'stopTransaction',
+          message: `Transaction data is suspicious - billing operation has been aborted - transaction ID: ${transaction.id}`,
+        });
+        // Abort the billing process - thresholds are not met!
+        return false;
+      }
+    }
+    // billing data sounds correct
+    return true;
+  }
+
+  protected computeTimeSpentInSeconds(transaction: Transaction): number {
+    let totalDuration: number;
+    if (!transaction.stop) {
+      totalDuration = moment.duration(moment(transaction.lastConsumption.timestamp).diff(moment(transaction.timestamp))).asSeconds();
+    } else {
+      totalDuration = moment.duration(moment(transaction.stop.timestamp).diff(moment(transaction.timestamp))).asSeconds();
+    }
+    return totalDuration;
+  }
+
+  protected convertTimeSpentToString(transaction: Transaction): string {
+    const totalDuration = this.computeTimeSpentInSeconds(transaction);
+    return moment.duration(totalDuration, 's').format('h[h]mm', { trim: false });
   }
 
   private async _synchronizeUser(user: User, forceMode = false): Promise<BillingUser> {


src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -300,7 +300,7 @@ export default class OCPPUtils {
     }
   }
 
-  public static async processTransactionBilling(tenant: Tenant, transaction: Transaction, action: TransactionAction): Promise<void> {
+  public static async processTransactionBilling(tenant: Tenant, transaction: Transaction, chargingStation: ChargingStation, action: TransactionAction): Promise<void> {
```
ClaudeROSSI, line 303: @LucasBrazi06 : As discussed I have tried to align the logic with the processTransactionRoaming approach. 
But I couldn't apply exactly the same logic and I'm not satisfied with the current solution.
I do not like the fact that the signature of the **processTransactionBilling** method expects parameters that are most of the time not necessary. The charging station information is now provided but only used during the START transaction.
From my perspective, to be on the safe side, we should avoid the **switch** on the **TransactionAction**. Each event (START, STOP, END) should have its own signature to enforce the calling layer to only pass the required information.
Let me know what you think.

LucasBrazi06: We should have the same approach when processing this kind of component whatever you choose.
Make me a proposal.

src/integration/billing/stripe/StripeBillingIntegration.ts
```diff
@@ -795,6 +795,12 @@ export default class StripeBillingIntegration extends BillingIntegration {
         withBillingActive: false
       };
     }
+    if (!transaction.chargeBox?.siteArea?.accessControl) {
```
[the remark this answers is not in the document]
ClaudeROSSI: transaction.siteArea is not set when calling start transaction. 
I now pass the information as an additional parameter.

src/server/ocpp/utils/OCPPUtils.ts
ClaudeROSSI: Well ... I would be to move the split the code that we have in OCPPUtils.processTransactionBilling and move it as static methods of the BillingIntegration class.
- BillingIntegration.processStartTransation()
- BillingIntegration.processStopTransation()
- etc

But I'm not sure the same logic will match the processTransactionRoaming() use-case.

Can you review the current solution and let me know if the current implementation is acceptable?

src/assets/server/rest/v1/schemas/user/user-default-tag-car-get.json
```diff
@@ -6,6 +6,9 @@
     "UserID": {
       "$ref": "user#/definitions/id"
     },
+    "ChargingStationID": {
+      "$ref": "common#/definitions/string-ids"
```
LucasBrazi06, line 10: It's only one charging station ID so use 'string-id'

src/integration/billing/BillingIntegration.ts
```diff
@@ -228,7 +229,7 @@ export default abstract class BillingIntegration {
     }
   }
 
-  public checkStartTransaction(transaction: Transaction): void {
+  public checkStartTransaction(transaction: Transaction, chargingStation: ChargingStation): void {
```
LucasBrazi06, line 232: Pass also User and SiteArea in param?

src/integration/billing/BillingIntegration.ts
```diff
@@ -247,18 +248,23 @@ export default abstract class BillingIntegration {
...
+    if (!chargingStation) {
+      throw new BackendError({
+        message: 'The charging station is mandatory to start a transaction',
+        module: MODULE_NAME,
+        method: 'checkStartTransaction',
+        action: ServerAction.BILLING_TRANSACTION
+      });
+    }
+    // Check for the Site Area
+    if (!chargingStation.siteArea) {
```
LucasBrazi06, line 260: If Org is not active, no Site Area will be provide.
Check if the org component is active.
Missing unit tests? ;-)

src/server/ocpp/utils/OCPPUtils.ts
LucasBrazi06: I checked and pass also Site Area and User in addition to Charging Station.
I can do the same for the Roaming.

src/server/rest/v1/router/api/UserRouter.ts
```diff
@@ -64,7 +64,11 @@ export default class UserRouter {
 
   private buildRouteUserDefaultCarTag(): void {
     this.router.get(`/${ServerRoute.REST_USER_DEFAULT_TAG_CAR}`, async (req: Request, res: Response, next: NextFunction) => {
-      req.query.ID = req.params.id;
```
LucasBrazi06, line 67: Do not remove req.query.ID assignment as it has to work in the future when www and app will be fixed.

LucasBrazi06 (changes requested):

src/assets/server/rest/v1/schemas/user/user-default-tag-car-get.json
ClaudeROSSI: done

src/server/ocpp/utils/OCPPUtils.ts
ClaudeROSSI: Same comment: 4 distinct methods in **BillingFacade** class

src/integration/billing/BillingIntegration.ts
ClaudeROSSI: Same comment: 4 distinct methods in **BillingFacade** class

src/integration/billing/BillingIntegration.ts
ClaudeROSSI: done.
I still have missing tests - but couldn't implement them with the new Jest framework!

src/integration/billing/BillingFacade.ts
```diff
@@ -0,0 +1,133 @@
...
+  public static async processStartTransaction(tenant: Tenant, transaction: Transaction, chargingStation: ChargingStation, siteArea: SiteArea): Promise<void> {
+    if (!transaction.user || !transaction.user.issuer) {
+      return;
+    }
+    const billingImpl = await BillingFactory.getBillingImpl(tenant);
+    if (billingImpl) {
+      try {
+        let withBillingActive = false;
+        // Check parameters' consistency
+        if (billingImpl.checkStartTransaction(transaction, chargingStation, siteArea)) {
```
ClaudeROSSI, line 25: The **checkTransactionMethod**() is now explicitly called by the abstraction layer. The concrete implementation (i.e. STRIPE) does not need to take care of this logic anymore.

src/integration/billing/BillingIntegration.ts
```diff
@@ -247,18 +251,64 @@ export default abstract class BillingIntegration {
...
 
-  private async _getUsersWithNoBillingData(): Promise<User[]> {
-    const newUsers = await UserStorage.getUsers(this.tenant,
-      {
-        statuses: [UserStatus.ACTIVE],
-        notSynchronizedBillingData: true
-      }, Constants.DB_PARAMS_MAX_LIMIT);
-    if (newUsers.count > 0) {
-      return newUsers.result;
+  protected async updateTransactionsBillingData(billingInvoice: BillingInvoice): Promise<void> {
```
ClaudeROSSI, line 283: This code has be moved out of the STRIPE implementation. 
It is now part of the billing abstraction to make sure all future implementations can benefit from it.

src/server/rest/v1/service/UserService.ts
```diff
@@ -81,9 +82,22 @@ export default class UserService {
         );
       }
     }
+    let withBillingChecks = true ;
+    if (filteredRequest.ChargingStationID) {
+      // TODO - The ChargingStationID is optional but only for backward compatibility reasons - make it mandatory as soon as possible
+      const chargingStation = await ChargingStationStorage.getChargingStation(req.tenant, filteredRequest.ChargingStationID,
```
LucasBrazi06, line 88: To use the Auth check method:
```
    const chargingStation = await UtilsService.checkAndGetChargingStationAuthorization(
      req.tenant, req.user, filteredRequest.ID, action, null, {
        withSite: true,
        withSiteArea: true,
      }, true);

```

src/server/rest/v1/router/api/UserRouter.ts
LucasBrazi06: To not forget to change the WWW et App to use this URL path correctly

LucasBrazi06 (approved):

PR #3295, opened 2022-01-27
Align migrated abac endpoints

Ramzay:
This PR was opened in order to be aligned with the Car GUI auth alignement:
- Full car component alignement
- "Light" alignement with other components: other PR will be opened to create / adapt auth definition


src/authorization/AuthorizationsDefinition.ts
```diff
@@ -82,7 +82,7 @@ export const AUTHORIZATION_DEFINITION: AuthorizationDefinition = {
           'fastChargePowerMax', 'drivetrainPowerHP'
         ]
       },
-      { resource: Entity.CAR_CATALOG, action: Action.SYNCHRONIZE },
+      { resource: Entity.CAR_CATALOG, action: [Action.SYNCHRONIZE, Action.DEBUG] },
```
LucasBrazi06, line 85: Debug??? :-)

src/authorization/AuthorizationsDefinition.ts
```diff
@@ -487,7 +487,7 @@ export const AUTHORIZATION_DEFINITION: AuthorizationDefinition = {
           'chargeAlternativePhase', 'chargeOptionPower', 'chargeOptionPhase', 'image', 'chargeOptionPhaseAmp', 'chargeAlternativePhaseAmp'
         ]
       },
-      { resource: Entity.CAR, action: [Action.CREATE, Action.UPDATE, Action.DELETE] },
+      { resource: Entity.CAR, action: [Action.CREATE, Action.CREATE_POOL_CAR, Action.UPDATE, Action.DELETE] },
```
LucasBrazi06, line 490: Action.CREATE_POOL_CAR should be a CREATE + do not provide the option Pool Car in the UI + Assert to not have the value Pool Car in the backend.

src/server/rest/v1/service/AuthorizationService.ts
```diff
@@ -594,31 +594,34 @@ export default class AuthorizationService {
...
     car.canRead = true; // Always true as it should be filtered upfront
-    car.canDelete = await AuthorizationService.canPerformAuthorizationAction(
-      tenant, userToken, Entity.CAR, Action.DELETE, authorizationFilter, { CarID: car.id }, car);
-    car.canUpdate = await AuthorizationService.canPerformAuthorizationAction(
-      tenant, userToken, Entity.CAR, Action.UPDATE, authorizationFilter, { CarID: car.id }, car);
+    car.canDelete = await AuthorizationService.canPerformAuthorizationAction(tenant, userToken, Entity.CAR, Action.DELETE, authorizationFilter, { CarID: car.id }, car);
+    car.canUpdate = await AuthorizationService.canPerformAuthorizationAction(tenant, userToken, Entity.CAR, Action.UPDATE, authorizationFilter, { CarID: car.id }, car);
+    car.canListUsers = await AuthorizationService.canPerformAuthorizationAction(tenant, userToken, Entity.USER, Action.LIST, authorizationFilter);
+    car.canCreatePoolCar = await AuthorizationService.canPerformAuthorizationAction(tenant, userToken, Entity.CAR, Action.CREATE_POOL_CAR, authorizationFilter);
```
LucasBrazi06, line 610: Too specific.

src/server/rest/v1/service/AuthorizationService.ts
```diff
@@ -646,6 +649,8 @@ export default class AuthorizationService {
       tenant, userToken, Entity.CAR_CATALOG, Action.DELETE, authorizationFilter, { CarCatalogID: carCatalog.id }, carCatalog);
     carCatalog.canUpdate = await AuthorizationService.canPerformAuthorizationAction(
       tenant, userToken, Entity.CAR_CATALOG, Action.UPDATE, authorizationFilter, { CarCatalogID: carCatalog.id }, carCatalog);
+    carCatalog.canDebug = carCatalog.canUpdate = await AuthorizationService.canPerformAuthorizationAction(
```
LucasBrazi06, line 652: To remove the canDebug ?

src/types/Authorization.ts
```diff
@@ -156,6 +156,8 @@ export enum Action {
   EXPORT_OCPP_PARAMS = 'ExportOCPPParams',
   GENERATE_QR = 'GenerateQrCode',
   MAINTAIN_PRICING_DEFINITIONS = 'MaintainPricingDefinitions',
+  CREATE_POOL_CAR = 'CreatePoolCar',
+  DEBUG = 'Debug',
```
LucasBrazi06, line 160: Check my comments on those two actions.

src/utils/LoggingHelper.ts
```diff
@@ -46,4 +49,27 @@ export default class LoggingHelper {
...
+      registrationTokenID: registrationToken.id,
+      siteAreaID: registrationToken.siteAreaID
+    };
+  }
+
+  public static getTagProperties(tag: Tag): { tagID: string; userID: string, visualID: string } {
+    return {
+      tagID: tag.id,
+      userID: tag.userID,
+      visualID: tag.visualID
```
LucasBrazi06, line 72: Same for tagID and visualID

src/utils/LoggingHelper.ts
```diff
@@ -46,4 +49,27 @@ export default class LoggingHelper {
...
+    return {
+      carID: car.id,
+      userID: car.userID,
+      carCatalogID: car.carCatalogID
+    };
+  }
+
+  public static getRegistrationTokenProperties(registrationToken: RegistrationToken): { registrationTokenID: string; siteAreaID: string } {
+    return {
+      registrationTokenID: registrationToken.id,
```
LucasBrazi06, line 63: Same for registrationTokenID

src/utils/LoggingHelper.ts
```diff
@@ -46,4 +49,27 @@ export default class LoggingHelper {
       siteID: siteArea.siteID,
     };
   }
+
+  public static getCarProperties(car: Car): { carID: string; userID: string, carCatalogID: number } {
+    return {
+      carID: car.id,
```
LucasBrazi06, line 55: Do both carID and carCatalogID are stored in the logs?
This is the purpose of this class.

LucasBrazi06 (changes requested):
Check my comments.
Thanks for the LoggingHelper refactoring!!!

src/authorization/AuthorizationsDefinition.ts
Ramzay: Yeah the superadmin has a tab called "advanced" where we display a json of the current car catalog item:
image
Maybe we should remove this tab entirely ?

src/server/rest/v1/service/AuthorizationService.ts
Ramzay: There was a rule in the GUI that only allowed admin to create a pool car (site admin not allowed to see the pool car option in the create dialog) maybe this is not the desired behavior ? 
Actually a site admin can edit a pool car but is not allowed to transform a "normal car" to a "pool car" but only the other way round: pool car -> normal car

Ramzay: Create pop up for site admin: no pool car option (same as edit dialog for pool car)
image
Site admin edit pool car
image

src/authorization/AuthorizationsDefinition.ts
LucasBrazi06: LOL, I forgot this one!!!
Yes remove the tab :-) and the projection to not retrieve it from the DB.
Thanks!

src/server/rest/v1/service/AuthorizationService.ts
LucasBrazi06: Normally, only Admin/Site Admin can create a pool car and change it to another type of car and vice-versa.
So I would expect in the UI that for Basic user you don't display the pool car type (with metaData?) and in the backend an assertion that checks that basic user cannot set this value.
Let me know.

LucasBrazi06 (approved):
GUT!!!

PR #3305, opened 2022-01-28
Billing - remove feature toggle lazy user sync

ClaudeROSSI:
cleanup code which is now useless


ClaudeROSSI:
@LucasBrazi06 , @Ninjeneer : I removed all obsolete endpoints regarding the billing user and invoice sync.

The only one which remains is:
BillingService.**handleForceSynchronizeUser**()

users with "no billing data" are not considered as an error anymore. Billing data is created on-demand when the user enters a payment method.

test/api/BillingTest.ts
```diff
@@ -356,64 +354,15 @@ describe('Billing', function() {
...
-          expect(response.data.inSuccess).to.be.eq(1);
-        });
-
         it('Should list invoices', async () => {
           const response = await billingTestHelper.userService.billingApi.readInvoices({}, TestConstants.DEFAULT_PAGING, TestConstants.DEFAULT_ORDERING);
           expect(response.status).to.be.eq(StatusCodes.OK);
           expect(response.data.result.length).to.be.gt(0);
         });
-
-        xit('Should list filtered invoices', async () => {
```
Ninjeneer, line 380: Should we not keep the tests on filtering invoices ?

ClaudeROSSI: > Should we not keep the tests on filtering invoices ?

you're right, but I have not been able to have this one being stable - the result was depending on the other test results and on the order of the execution as well.

I consider this one useless because we have other tests using the filtering for instance when retrieving DRAFT invoices. c.f.: method: **BillingTestHelper.getDraftInvoices()**

PR #3306, opened 2022-01-28
billing - preserve session billing data

ClaudeROSSI:
URGENT!

hot fix to make sure the billing data of the last transaction is preserved when computing the extra inactivity

### DIFF
--- a/src/integration/billing/stripe/StripeBillingIntegration.ts
+++ b/src/integration/billing/stripe/StripeBillingIntegration.ts
@@ -997,15 +997,12 @@ export default class StripeBillingIntegration extends BillingIntegration {
         module: MODULE_NAME, method: 'endTransaction',
         message: `Operation aborted - unexpected situation - the session has already been billed - transaction ID: ${transaction.id}`
       });
-      return {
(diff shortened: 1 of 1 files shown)


ClaudeROSSI:
@LucasBrazi06 : This one is a bit important!
I will have to also have a task to repair sessions where the billing data does not include the invoice number anymore!

LucasBrazi06:
> @LucasBrazi06 : This one is a bit important! I will have to also have a task to repair sessions where the billing data does not include the invoice number anymore!

You need a deployment?

ClaudeROSSI:
> > @LucasBrazi06 : This one is a bit important! I will have to also have a task to repair sessions where the billing data does not include the invoice number anymore!
> 
> You need a deployment?

a deployment would be cool ... but anyway, we already have a lot of sessions to repair ... so the deployment can wait.
This only impact the session history ... where the invoice number is not shown anymore!

LucasBrazi06:
> a deployment would be cool ... but anyway, we already have a lot of sessions to repair ... so the deployment can wait.
> This only impact the session history ... where the invoice number is not shown anymore!

Ok, let me know then.

PR #3312, opened 2022-01-31
Billing - moment issues - format duration

ClaudeROSSI:
DRAFT = to investigate moment formatting issues on session duration


ClaudeROSSI:
current solution is acceptable - no customer complaints so far - closing this PR

PR #3319, opened 2022-02-04
fix - Clear Connector Runtime Data on status Notif

ClaudeROSSI:


src/server/ocpp/services/OCPPService.ts
```diff
@@ -815,7 +815,9 @@ export default class OCPPService {
         if (transactionUpdated || billingDataUpdated || ocpiUpdated || oicpUpdated) {
           await TransactionStorage.saveTransaction(tenant, lastTransaction);
         }
-      } else if (!Utils.isNullOrUndefined(lastTransaction)) {
+      }
```
ClaudeROSSI, line 818: DRAFT - I'm not sure about that change.

The goal is to prevent the following error:

**Error: "Cannot soft stop an ongoing transaction"**

ClaudeROSSI, line 818: DRAFT - I'm not sure about that change.

The goal is to prevent the following error:

**Error: "Cannot soft stop an ongoing transaction"**

ClaudeROSSI:
@LucasBrazi06 : This is a DRAFT because I'm not sure about the impact

src/server/ocpp/services/OCPPService.ts
```diff
@@ -798,32 +798,35 @@ export default class OCPPService {
...
-      } else if (!Utils.isNullOrUndefined(lastTransaction)) {
-        await Logging.logWarning({
-          tenantID: tenant.id,
-          ...LoggingHelper.getChargingStationProperties(chargingStation),
-          module: MODULE_NAME, method: 'checkAndUpdateLastCompletedTransaction',
-          action: ServerAction.OCPP_STATUS_NOTIFICATION,
-          message: `${Utils.buildConnectorInfo(lastTransaction.connectorId, lastTransaction.id)} Received Status Notification '${statusNotification.status}' while a transaction is ongoing`,
-          detailedMessages: { statusNotification }
-        });
+        // Clear Connector Runtime Data
```
ClaudeROSSI, line 829: @LucasBrazi06 : DRAFT - I'm not sure about that change. The goal is to prevent the following error:
**Error: "Cannot soft stop an ongoing transaction"**

We now clear the connnector runtime data event when the lastTransaction.stop was set.
Without that change the UT testing the SOFT STOP does not work.

PR #3320, opened 2022-02-04
Fix lookup site in charging station

mahdibenromdhane:


mahdibenromdhane:
> No, wrong fix. The Charging Station has companyID, siteID and siteAreaID. Check with siteID is not provided in the Charging Station object.

I'v used a different approach to set/get the good company ID and site ID in charging stations. 
Let me know if I missed something

src/migration/tasks/FixChargingStationCompanyIDsTask.ts
```diff
@@ -0,0 +1,55 @@
...
+  public async migrateTenant(tenant: Tenant): Promise<void> {
+    let updated = 0, siteArea, site;
+    // Get all the charging stations
+    const chargingStations = await global.database.getCollection<any>(tenant.id, 'chargingstations').find({})
+      .project({ siteAreaID: 1, _id: 1 })
+      .toArray();
+    if (!Utils.isEmptyArray(chargingStations)) {
+      for (const chargingStation of chargingStations) {
+        siteArea = await global.database.getCollection<any>(tenant.id, 'siteareas').findOne({ _id: chargingStation.siteAreaID });
+        site = await global.database.getCollection<any>(tenant.id, 'sites').findOne({ _id: siteArea.siteID });
```
LucasBrazi06, line 21: Check if the Site Area exists otherwise log an error.
Same for Site.

src/migration/tasks/FixChargingStationSiteIDsTask.ts
```diff
@@ -0,0 +1,54 @@
...
+import Logging from '../../utils/Logging';
+import { ServerAction } from '../../types/Server';
+import Tenant from '../../types/Tenant';
+import TenantMigrationTask from '../TenantMigrationTask';
+import Utils from '../../utils/Utils';
+import global from '../../types/GlobalType';
+
+const MODULE_NAME = 'FixChargingStationSiteIDsTask';
+
+export default class FixChargingStationSiteIDsTask extends TenantMigrationTask {
```
LucasBrazi06, line 11: No need this task.
Why not fixing it in FixChargingStationCompanyIDsTask?

src/server/rest/v1/service/SiteService.ts
```diff
@@ -400,7 +400,10 @@ export default class SiteService {
       req.tenant, req.user, filteredRequest.companyID, Action.READ, action, filteredRequest);
     // Update
     site.name = filteredRequest.name;
-    site.companyID = filteredRequest.companyID;
+    if (site.companyID !== filteredRequest.companyID) {
```
LucasBrazi06, line 403: Same remark, after the save if the Site

src/server/rest/v1/service/SiteAreaService.ts
```diff
@@ -364,7 +365,10 @@ export default class SiteAreaService {
     }
     siteArea.smartCharging = filteredRequest.smartCharging;
     siteArea.accessControl = filteredRequest.accessControl;
-    siteArea.siteID = filteredRequest.siteID;
+    if (siteArea.siteID !== filteredRequest.siteID) {
+      await ChargingStationStorage.updateChargingStationsSite(req.tenant, filteredRequest.id, filteredRequest.siteID);
```
LucasBrazi06, line 369: Always call this method.
To put after the save.

src/storage/mongodb/ChargingStationStorage.ts
```diff
@@ -969,6 +968,32 @@ export default class ChargingStationStorage {
...
 
+  public static async updateChargingStationsSite(tenant: Tenant, siteAreaID: string, siteID: string): Promise<void> {
+    const startTime = Logging.traceDatabaseRequestStart();
+    DatabaseUtils.checkTenantObject(tenant);
+    await global.database.getCollection<any>(tenant.id, 'chargingstations').updateMany(
+      {
+        siteAreaID: DatabaseUtils.convertToObjectID(siteAreaID),
+      },
+      {
+        $set: { siteID: DatabaseUtils.convertToObjectID(siteAreaID) }
```
LucasBrazi06, line 979: Not tested!!!
siteAreaID -> siteID !!!

LucasBrazi06, line 979: I would have expected also an update with the companyID.

LucasBrazi06 (changes requested):
Test more your changes!
Check my comments.

I think we have the same problem with Asset.
Double check Asset and if necessary provide a fix in another PR (at the same time you can also add the missing companyID in Asset not yet saved).
Thanks!

src/migration/tasks/FixChargingStationSiteIDsTask.ts
mahdibenromdhane: True, now it's updated in the same task

src/server/rest/v1/service/SiteAreaService.ts
mahdibenromdhane: Done

src/storage/mongodb/ChargingStationStorage.ts
mahdibenromdhane: I didn’t understand this remark, The company ID would only change in the charging station if the site has changed the company.

mahdibenromdhane:
> Test more your changes! Check my comments.
> 
> I think we have the same problem with Asset. Double check Asset and if necessary provide a fix in another PR (at the same time you can also add the missing companyID in Asset not yet saved). Thanks!

I will check the assets and provide a fix if necessary.

src/storage/mongodb/ChargingStationStorage.ts
LucasBrazi06: If you change the site of a site area, both the siteID and the companyID of the charging station are changed also.

LucasBrazi06 (changes requested):
I refactored the code to reuse MDB requests, updated the migration task to work on site areas instead of charging stations and I added the companyID in Asset.
Double check the changes.
Provide/Complete unit tests that checks:
- Asset has company ID when created/updated
- Change of Site / Site Area: check Asset, Charging Stations, Transactions

PR #3331, opened 2022-02-10
Billing - pay invoice

ClaudeROSSI:
DRAFT

work in progress to view an invoice and pay (if necessary)


ClaudeROSSI:
too old - and no option to pay with a temporary payment intent. STRIPE api is deprecated and no pressure from customers to have it.

PR #3340, opened 2022-02-14
Handle particular case in CleanTransactionsInProgressTask

mahdibenromdhane:


src/scheduler/tasks/CloseTransactionsInProgressTask.ts
```diff
@@ -35,14 +38,22 @@ export default class CloseTransactionsInProgressTask extends TenantSchedulerTask
...
+                  tenant, transaction.chargeBox, transaction, transaction.tagID, true);
+              }
+              // Update Transaction with Stop Transaction and Stop MeterValues
+              OCPPUtils.updateTransactionWithStopTransaction(transaction, transaction.chargeBox ? transaction.chargeBox : null, {
+                transactionId: transaction.id,
+                chargeBoxID: transaction.chargeBoxID,
+                idTag: transaction.tagID,
+                timestamp: Utils.convertToDate(transaction.lastConsumption ? transaction.lastConsumption.timestamp : transaction.timestamp).toISOString(),
+                meterStop: transaction.lastConsumption ? transaction.lastConsumption.value : transaction.meterStart
+              }, authUser?.user, authUser?.alternateUser, transaction.tagID, true);
```
ClaudeROSSI, line 53: one remark: if authUser is null, the billing process will simply be skipped.
This will result in having a session with no billing data.
I assume this is acceptable but I couldn't test it!

src/scheduler/tasks/CloseTransactionsInProgressTask.ts
```diff
@@ -35,14 +38,22 @@ export default class CloseTransactionsInProgressTask extends TenantSchedulerTask
...
+              // Save the transaction
+              await TransactionStorage.saveTransaction(tenant, transaction);
+              result.inSuccess++;
```
ClaudeROSSI, line 56: wrong ServerAction is used on line 71 
>> ServerAction.**TAGS_IMPORT**

ClaudeROSSI (approved):
a few comments.

How can we proceed to test it?

src/scheduler/tasks/CloseTransactionsInProgressTask.ts
mahdibenromdhane: I understand that you need a reproduction scenario to have stopped session with authUser null. 
If it is the case, you need just to delete a tag of an ongoing session to stop and run the schedular task.

PR #3369, opened 2022-02-21
Jest extension integration

Ramzay:
Base configuration to make the Jest extension work with our project


ClaudeROSSI (approved):
vscode integration is still a bit frustrating but works properly ;)

ClaudeROSSI:
@LucasBrazi06 : just merged that one because i need the jest config for testing the other branches

PR #3371, opened 2022-02-21
Test - create context - POC no additional test conf

ClaudeROSSI:
Just a try to avoid the additional jest config 

- Surprisingly, this the existing code is not transpiled properly!
- Let's discuss that with the team


ClaudeROSSI:
@Ramzay : npm run **test:createContext-poc**

ClaudeROSSI:
Gave up !

PR #3377, opened 2022-02-23
billing test - adaptation to jest framework + tz checks

ClaudeROSSI:
- Test Adaptation to jest framework
- Additional checks to make it run whatever the timezone is

image


Ramzay:
@ClaudeROSSI could you also use global assert in BillingTest instead of chai ? 
PS: Unit test failed :/ (but now the message is clearer !)

ClaudeROSSI:
@Ramzay : billing tests do not work anymore.
I'll isolate the jest adaptation in another branch and close that PR.
Sorry for the confusion.

PR #3386, opened 2022-02-25
billing - check free access before checking customerID

ClaudeROSSI:
Free Access:

- Putting back the check of the free access to the **BillingAbstraction.checkStartTransaction()** METHOD
- Moved down the check of the stripe **customerID** 

Note: This also impacts the check of the "siteArea.accessControl" flag! - we potencially had the same problem!


LucasBrazi06 (approved):
Thanks!

PR #3388, opened 2022-02-25
Transaction Service - MongoDB optimization - LoadAllConsumptions 

RohanAntony:
1. In sessions tab, consumption chart for any session must load only a subset of data. This should be one point every 5 minutes including the last minute in the session.
2. On selecting "Show full chart resolution", entire session is to be displayed.
3. On clicking a legend, the respective dataset in the chart is either hidden or shown. Make sure that this happens and the state of all the legends is maintained when clicking "Show full chart resolution".
4. Make sure the grid line on the background is always visible as long as a dataset is visible.


test/api/client/TransactionApi.ts
```diff
@@ -24,7 +24,8 @@ export default class TransactionApi extends CrudApi {
   }
 
   public async readAllConsumption(params) {
-    return super.read(params, `${this.buildRestEndpointUrl(RESTServerRoute.REST_TRANSACTION_CONSUMPTIONS, { id: params.TransactionId })}?WithUser=true`);
+    const id = params.TransactionId;
+    return super.read(params, `${this.buildRestEndpointUrl(RESTServerRoute.REST_TRANSACTION_CONSUMPTIONS, { id })}?WithUser=true&LoadAllConsumptions=true`);
```
ClaudeROSSI, line 28: @LucasBrazi06 : I'm a bit surprised that we now need to force **LoadAllConsumption** to true to have the tests running properly again. Is this normal?

ClaudeROSSI (approved):
approved and tested 

But I see a suspicious change in the unit tests ... not sure this is normal!

test/api/client/TransactionApi.ts
LucasBrazi06: Yes if you don't want to miss any values.
The optimized version may not return all the points now.

PR #3416, opened 2022-03-15
init user settings template

Ramzay:


cabch (approved):
ok for me

.gitignore
```diff
@@ -52,7 +52,7 @@ tmp
 
 # vscode
 .vscode/*
-!.vscode/settings.json
+!.vscode/settings-template.json
```
ClaudeROSSI, line 55: This change will gitignore the settings.json file which is still in the repo.
I think we need to remove it from the depot before adding it to gitignore. right?

ClaudeROSSI (approved):

PR #3417, opened 2022-03-15
Uts for orgs

mahdibenromdhane:


src/storage/mongodb/SiteAreaStorage.ts
```diff
@@ -45,7 +45,8 @@ export default class SiteAreaStorage {
           {
             $set: {
               siteAreaID: DatabaseUtils.convertToObjectID(siteArea.id),
-              siteID: DatabaseUtils.convertToObjectID(siteArea.siteID)
+              siteID: DatabaseUtils.convertToObjectID(siteArea.siteID),
+              companyID: DatabaseUtils.convertToObjectID(siteArea.site?.companyID)
```
ClaudeROSSI, line 49: I do not understand why we need to add companyID here.

The query is done using { withSite: true } so the information is already available, right?
>> siteArea.site?.companyID

src/server/rest/v1/service/AuthorizationService.ts
```diff
@@ -554,7 +554,7 @@ export default class AuthorizationService {
         'site.id', 'site.public', 'site.name', 'site.address', 'siteID', 'voltage', 'coordinates', 'forceInactive', 'manualConfiguration', 'firmwareUpdateStatus', 'tariffID',
         'capabilities', 'endpoint', 'chargePointVendor', 'chargePointModel', 'ocppVersion', 'ocppProtocol', 'lastSeen',
         'firmwareVersion', 'currentIPAddress', 'ocppStandardParameters', 'ocppVendorParameters', 'connectors', 'chargePoints',
-        'createdOn', 'chargeBoxSerialNumber', 'chargePointSerialNumber', 'powerLimitUnit'
+        'createdOn', 'chargeBoxSerialNumber', 'chargePointSerialNumber', 'powerLimitUnit', 'companyID'
```
ClaudeROSSI, line 557: why don't we expose 'site.companyID' instead?

src/storage/mongodb/SiteAreaStorage.ts
```diff
@@ -66,7 +67,8 @@ export default class SiteAreaStorage {
           {
             $set: {
               siteAreaID: null,
-              siteID: null
+              siteID: null,
+              companyID: null
```
ClaudeROSSI, line 71: idem ... this should not be necessary

ClaudeROSSI (changes requested):
please have a look to my comments.

BTW: automated tests are failing. Can you check why?

LucasBrazi06:
@ClaudeROSSI Could you check the latest Mahdi's changes?

ClaudeROSSI (approved):
@mahdibenromdhane: your code was correct.

My latest comments can be ignored. After verifying again with Serge we could confirm that the siteArea does not include the companyID and this is normal.

PR #3421, opened 2022-03-17
Transaction - consumptions - add priceUnit to projection

ClaudeROSSI:
The mobile app needs the currency code when displaying the amount curve in the session chart.
This information was not part of the projection!


ClaudeROSSI:
@AlixH : let me know if this is sufficient for your use-case.

AlixH (approved):

PR #3454, opened 2022-04-05
Set limits of ocpp charging profiles to one decimal place

OliveGerste:


ClaudeROSSI (approved):
@LucasBrazi06 : seems ok to me.
A customer requested to limit charging profiles to one decimal to make it compliant with the OCPP spec.
Tobias wants to merge it as soon as possible.
What do you think?

LucasBrazi06:
> @LucasBrazi06 : seems ok to me. A customer requested to limit charging profiles to one decimal to make it compliant with the OCPP spec. Tobias wants to merge it as soon as possible. What do you think?

Okay for me.

PR #3456, opened 2022-04-05
Sub-Site Area review

LucasBrazi06:


src/server/rest/v1/service/UtilsService.ts
```diff
@@ -2025,4 +2005,99 @@ export default class UtilsService {
     }
     return tag;
   }
+
+  private static buildSiteAreaTree(currentSiteArea: SiteArea, siteAreas: SiteArea[]): SiteArea {
```
OliveGerste, line 2009: This method is also used by the smart charging integration. So I think it should not be placed in UtilsService and should not be private.
I saw that you removed the s for multiple trees from the method name. The reason for the s was that there can be multiple trees on one site. E.G. Site: Mougins --> 3 root site areas --> 3 independent Trees which are build with this method

src/server/rest/v1/service/UtilsService.ts
```diff
@@ -1518,34 +1518,14 @@ export default class UtilsService {
     }
   }
 
-  public static async checkIfSiteAreaTreeValid(siteArea: SiteArea, tenant: Tenant, user: UserToken): Promise<void> {
-    const siteAreas = await SiteAreaStorage.getSiteAreas(tenant,
-      { siteIDs: [siteArea.siteID] }, Constants.DB_PARAMS_MAX_LIMIT,
+  public static async checkIfSiteAreaTreeValid(siteArea: SiteArea): Promise<void> {
+    const siteAreas = await SiteAreaStorage.getSiteAreas(req.tenant,
```
OliveGerste, line 1522: req does not exist in the parameters of the method. This causes build errors.

src/server/rest/v1/service/UtilsService.ts
```diff
@@ -1518,34 +1518,14 @@ export default class UtilsService {
     }
   }
 
-  public static async checkIfSiteAreaTreeValid(siteArea: SiteArea, tenant: Tenant, user: UserToken): Promise<void> {
-    const siteAreas = await SiteAreaStorage.getSiteAreas(tenant,
-      { siteIDs: [siteArea.siteID] }, Constants.DB_PARAMS_MAX_LIMIT,
+  public static async checkIfSiteAreaTreeValid(siteArea: SiteArea): Promise<void> {
```
OliveGerste, line 1521: The changed type definition causes errors in site area service.

OliveGerste (changes requested):
@LucasBrazi06  I made some comments on the PR as I saw that it is to be reviewed.
The check tree method was moved to UtilsService. Do you think the check of the tree will never be needed anywhere else in the application? Maybe at some kind of migration task or something like this, where properties of site areas are changed?

LucasBrazi06:
> @LucasBrazi06 I made some comments on the PR as I saw that it is to be reviewed. The check tree method was moved to UtilsService. Do you think the check of the tree will never be needed anywhere else in the application? Maybe at some kind of migration task or something like this, where properties of site areas are changed?

Not yet to be reviewed, that's my pipeline, otherwise I would have put it in Cross review ;-)
I haven't finished and I'm doing this exercise with Claude who will get back to you to explain/cross check.

OliveGerste:
> > @LucasBrazi06 I made some comments on the PR as I saw that it is to be reviewed. The check tree method was moved to UtilsService. Do you think the check of the tree will never be needed anywhere else in the application? Maybe at some kind of migration task or something like this, where properties of site areas are changed?
> 
> Not yet to be reviewed, that's my pipeline, otherwise I would have put it in Cross review ;-) I haven't finished and I'm doing this exercise with Claude who will get back to you to explain/cross check.

Ah okay. Now I understand. Sorry for my misunderstanding. Let me know if you need anything in this matter.

LucasBrazi06:
Dups, new PR created from another branch

PR #3457, opened 2022-04-06
Site area  consumption chart aggregation should average at connector level before summing up at charging station level

RohanAntony:
Site area consumption chart was aggregating data on average value of charging station across multiple connectors instead of average value of connectors summed to charging station level per minute.


LucasBrazi06:
@RohanAntony Thanks for the fix but how can we test this?

RohanAntony:
> @RohanAntony Thanks for the fix but how can we test this?

We need the output of one connector to be 0 and other to be value X over a given time period. Previously, we would average it to X/2 per minute at charger level but now its sums up the averages of each connector per minute. I tested it by running the simulator and seeing the values close to as expected. Alternate could be to replace existing values in the database with 0 for a consistent period of time and see if the connectors are summed to each charger instead of averaged.

PR #3458, opened 2022-04-07
Meter value v2

melvyndubois:


src/server/ocpp/validation/OCPPValidation.ts
```diff
@@ -108,6 +109,8 @@ export default class OCPPValidation extends SchemaValidator {
   public async validateMeterValues(tenantID: string, chargingStation: ChargingStation, meterValues: OCPPMeterValuesRequestExtended): Promise<void> {
     // Always integer
     meterValues.connectorId = Utils.convertToInt(meterValues.connectorId);
```
LucasBrazi06, line 111: Utils.convertToInt() is not need, the schema should convert it already.
You can set the default value to 0 in the schema for the connectorId.

src/server/ocpp/validation/OCPPValidation.ts
```diff
@@ -108,6 +109,8 @@ export default class OCPPValidation extends SchemaValidator {
...
+    this.validate(this.meterValueRequest, meterValues);
+
```
LucasBrazi06, line 113: BL

src/server/ocpp/validation/OCPPValidation.ts
```diff
@@ -131,9 +134,10 @@ export default class OCPPValidation extends SchemaValidator {
         action: ServerAction.OCPP_METER_VALUES,
         message: `Connector ID '${meterValues.connectorId}' not found for Transaction ID '${meterValues.transactionId}'`
       });
+      return;
     }
     // Transaction ID is provided on Connector
-    if (foundConnector.currentTransactionID > 0) {
+    if (foundConnector?.currentTransactionID > 0) {
       // Check if provided in Meter Values
       meterValues.transactionId = Utils.convertToInt(meterValues.transactionId);
```
LucasBrazi06, line 142: Same as above, no need to convert it.

LucasBrazi06 (changes requested):
I put the ocpp definition in meter value schema file, no need a dedicated ocpp file as it is only used in one place.
Check my comments and fix.

LucasBrazi06:
@cabch @melvyndubois Check my feedbacks. Thanks!

LucasBrazi06:
@cabch @melvyndubois Nothing has been done on this PR according my feedback.
I would like to merge it next week.

cabch:
> @cabch @melvyndubois Nothing has been done on this PR according my feedback. I would like to merge it next week.

sorry, didn't check my github notification, melvyn is on it, will be done before noon

src/server/ocpp/validation/OCPPValidation.ts
LucasBrazi06: Yes

LucasBrazi06: You can also change the code to not check the '0' with the Utils.isUndef..()

cabch: in the if statement right ?

LucasBrazi06: > in the if statement right ?
Yes.
Test it with Posman.

melvyndubois: the case where the meterValue.transactionID is 0 is not check by utils.isNullorUndef, so if our value is 0, it will be not be true but we wait "true"

LucasBrazi06: > the case where the meterValue.transactionID is 0 is not check by utils.isNullorUndef, so if our value is 0, it will be not be true but we wait "true"

If meterValue.transactionID is filtered by the schema, it should be undefined when checked.
Test with Postman!

LucasBrazi06:
@melvyndubois If you have done with the requested changes, please move this issue into the 'To be Reviewed' pipeline

PR #3466, opened 2022-04-12
Send no image from the server

mahdibenromdhane:
@AlixH please check if there is any regression on the mobile side.


AlixH:
> @AlixH please check if there is any regression on the mobile side.

I'm checking

LucasBrazi06 (approved):
GUT!

PR #3467, opened 2022-04-12
billing - better handling of invoices with no metadata

ClaudeROSSI:
Avoid reusing DRAFT invoices with no metadata.


ClaudeROSSI:
@LucasBrazi06 : No risks on that one ... Minor change to avoid using DRAFT invoices having no metadata (Mehmet issue)

PR #3468, opened 2022-04-13
Billing - convert invoiceID to ObjectID

ClaudeROSSI:
Minor change to make sure the invoiceID is always persisted as an ObjectId().

Migration of old transactions should be necessary, but the information is not yet used by the UI and a SynchronizeInvoice endpoint is planned. This endpoint will somehow fix old transactions as well.


ClaudeROSSI:
@LucasBrazi06 : Minor change to normalize invoiceID in billing data

PR #3481, opened 2022-04-19
Clean up unnecessary auth checks in service classes

LucasBrazi06:


src/server/rest/v1/service/AssetService.ts
```diff
@@ -246,8 +246,8 @@ export default class AssetService {
     // Filter
     const filteredRequest = AssetValidator.getInstance().validateAssetsGetReq(req.query);
     // Check dynamic auth
-    const authorizationAssetsInErrorFilter = await AuthorizationService.checkAndGetAssetsAuthorizations(req.tenant, req.user, Action.IN_ERROR, filteredRequest);
-    if (!authorizationAssetsInErrorFilter.authorized) {
+    const authorizations = await AuthorizationService.checkAndGetAssetsAuthorizations(req.tenant, req.user, Action.IN_ERROR, filteredRequest);
```
ClaudeROSSI, line 249: @LucasBrazi06 : 
**canPerformAuthorizationAction** is called with the parameter **failsWithException** set to true.

My understanding is that we never reach line 251. 
>> UtilsService.sendEmptyDataResult(res, next);

Are we supposed to force **failsWithException** to false for all endpoints getting data for a list?

src/server/rest/v1/service/BillingService.ts
```diff
@@ -185,14 +185,9 @@ export default class BillingService {
     // Check dynamic authorization
     const authorizations = await AuthorizationService.checkAndGetInvoicesAuthorizations(req.tenant, req.user, filteredRequest);
     if (!authorizations.authorized) {
-      throw new AppAuthError({
-        errorCode: HTTPAuthError.FORBIDDEN,
-        user: req.user,
-        entity: Entity.INVOICE, action: Action.LIST,
-        module: MODULE_NAME, method: 'handleGetInvoices',
-      });
+      UtilsService.sendEmptyDataResult(res, next);
```
ClaudeROSSI, line 188: @LucasBrazi06 : regex to find all checks where **failsWithException** is set to true

canPerformAuthorizationAction.*Action.LIST.*((\n)+).*true

PR #3493, opened 2022-04-21
Sub site area - follow up

ClaudeROSSI:


LucasBrazi06:
@ClaudeROSSI Change of Site has been handled in sub-site area, you can check it and give me feedback.
Remaining tasks so far:
- Change of Smart Charging, Voltage, Nbr of Phases -> Propose to change the fill hierarchy
- Change of Power, check children's won't go over the parent's power
@OliveGerste Fyi, we recreated the branch for this topic. You can also check the feature if you want.
Thanks to both of you!

LucasBrazi06:
@ClaudeROSSI, as agreed I just made the following changes:
- Put Site selection before the sub-Site Area
- Assign Site when sub-Site Area is selected first
- Display sub-Site Area in read-only mode for Basic user

Could you double check and let me know so I can move ahead with the next changes:
- Change of Smart Charging, Voltage, Nbr of Phases -> Propose to change the fill hierarchy
- Change of Power, check children's won't go over the parent's power

ClaudeROSSI:
> @ClaudeROSSI I just added the update of transactions, charging stations, assets Org IDs when the Site of a Site Areas (or sub Site Area) has been changed (case 1). Let me know if you find something else, I will implement the Smart Charging flag update in the tree. Thanks!

tested. c'est bon pour moi - sauf les tests qui echouent sur les nouveau code d'erreurs. Nornmal je pense!

LucasBrazi06:
> tested. c'est bon pour moi - sauf les tests qui echouent sur les nouveau code d'erreurs. Nornmal je pense!

Oui, je vais regarder.
Merci !!!

LucasBrazi06:
@ClaudeROSSI 
Unit Tests fixed.
i18n fixed.
Added the update of the Smart Charging in the whole Site Area tree.
Let me know.
I'll do the latest one 'Check of power limits in children'

OliveGerste:
@LucasBrazi06  Thanks for all the refinements. 
One remark, I do not understand why we prevent the children power to be higher then the parents power. I thought this is exactly the use case we want to have. 
Example:
- Parent 100kW
- 1 Child 70kW
- 2 Child 70kW
Both children could consume140kW, smart charging will automatically take this into account and prevent blowing the parent fuse, by limiting the sub fuses.
Additionally parent limit could be increased by solar production to provide a higher limit.

LucasBrazi06:
> @LucasBrazi06 Thanks for all the refinements. One remark, I do not understand why we prevent the children power to be higher than the parents power. I thought this is exactly the use case we want to have. Example:
> 
> * Parent 100kW
> * 1 Child 70kW
> * 2 Child 70kW
>   Both children could consume140kW, smart charging will automatically take this into account and prevent blowing the parent fuse, by limiting the sub fuses.
>   Additionally parent limit could be increased by solar production to provide a higher limit.

@OliveGerste Thanks for your feedback!
Indeed, if it's managed by the optimizer, then no need to check the children's power.
I'll remove the assertion.

LucasBrazi06:
> @OliveGerste Thanks for your feedback!
> Indeed, if it's managed by the optimizer, then no need to check the children's power.
> I'll remove the assertion.

Done.

LucasBrazi06 (approved):

PR #3503, opened 2022-04-28
CANCELLED - Enpoints/Migration task to save/get mobile app settings

mahdibenromdhane:


src/migration/tasks/FillMobileAppSettingsTask.ts
```diff
@@ -0,0 +1,56 @@
...
+import SettingStorage from '../../storage/mongodb/SettingStorage';
+import Tenant from '../../types/Tenant';
+import TenantMigrationTask from '../TenantMigrationTask';
+import Utils from '../../utils/Utils';
+
+const MODULE_NAME = 'FillMobileAppSettingsTask';
+
+export default class FillMobileAppSettingsTask extends TenantMigrationTask {
+  public async migrateTenant(tenant: Tenant): Promise<void> {
+    let updated = 0;
```
ClaudeROSSI, line 15: a boolean would be sufficient here. Right?

src/migration/tasks/FillMobileAppSettingsTask.ts
```diff
@@ -0,0 +1,56 @@
...
+    let mobileSettingToSave: Partial<SettingDB>;
+    // Get mobile setting
+    const mobileSetting = await SettingStorage.getSettingByIdentifier(tenant, TechnicalSettings.MOBILE);
```
ClaudeROSSI, line 18: As discussed with Kinza, we would like to have a more generic module called WHITE_LABEL and thus corresponding settings. We will for sure have additional settings there that are not only related to the MOBILE app.
Can you change it?

ClaudeROSSI, line 18: As discussed with Kinza, we would like to have a more generic module called WHITE_LABEL and thus corresponding settings. We will for sure have additional settings there that are not only related to the MOBILE app.
Can you change it?

src/types/Setting.ts
```diff
@@ -106,6 +114,17 @@ export interface RoamingSettings extends Setting {
...
 }
+export interface MobileSettings extends Setting {
+  identifier: TechnicalSettings.MOBILE;
+  type: MobileSettingsType;
+  mobile?: MobileSetting;
+}
+
+export interface MobileSetting {
+  settingsIOSMobileAppID: string;
+  settingsAndroidMobileAppID: string;
```
ClaudeROSSI, line 125: well ... why not:

- iosMobileAppID
- androidMobileAppID
-

ClaudeROSSI, line 125: well ... why not:

- iosMobileAppID
- androidMobileAppID
-

ClaudeROSSI (changes requested):

ClaudeROSSI:
@mahdibenromdhane : as discussed, we need a White Label module. Thanks

PR #3511, opened 2022-05-05
refactor: removed useless utils checks

Ninjeneer:


LucasBrazi06 (changes requested):
Do you think we can also remove these methods from UtilsService?
image

Ninjeneer:
Yes, I'm working on it.

LucasBrazi06 (changes requested):
Some check in the remaining methods can be moved in the Json Schema:
image

LucasBrazi06 (approved):

PR #3520, opened 2022-05-11
fix api user login

ClaudeROSSI:
DRAFT - does it still work when doing B2B calls?


ClaudeROSSI:
@LucasBrazi06 - workaround to prevent API users to be able to login from our react native app.
What do you think?

LucasBrazi06:
> @LucasBrazi06 - workaround to prevent API users to be able to login from our react native app. What do you think?

Gut !!!
I thought also about a fix in the mobile/www app when the user managed to log in, you check the User API flag and log him out with an error message.

LucasBrazi06 (approved):

ClaudeROSSI:
> > @LucasBrazi06 - workaround to prevent API users to be able to login from our react native app. What do you think?
> 
> Gut !!! I thought also about a fix in the mobile/www app when the user managed to log in, you check the User API flag and log him out with an error message.

@LucasBrazi06 , @AlixH 
To be on the safe side it would make sense to also check the source. As far as I can see with the output of the debug session (from Alix) we get a source mentioning 'Darwin' 



With my changes, an API User CANNOT log in anymore with POSTMAN!!!

Darwin is for IOS ... I'm not sure what we get when trying with Android

PR #3532, opened 2022-05-18
Endpoint to export ocpi consumption data for advenir

mahdibenromdhane:


ClaudeROSSI:
@mahdibenromdhane : I did some fixes 
- the consumption values where in a random order (missing field in the projection)
- the timestamp was a string (instead of a number)
- the timestamp was not the one ADVENIR expects

I also changed the ADVENIR types to makevahe a strong-typed payload

Please try it again on your side and let me know what you think

ClaudeROSSI:
@mahdibenromdhane : It would be convenient to pass the userID has a query parameter to have it already populated in the response. What do you think?

image

mahdibenromdhane:
> @mahdibenromdhane : I did some fixes
> 
> * the consumption values where in a random order (missing field in the projection)
> * the timestamp was a string (instead of a number)
> * the timestamp was not the one ADVENIR expects
> 
> I also changed the ADVENIR types to makevahe a strong-typed payload
> 
> Please try it again on your side and let me know what you think

It works good thanks :)

mahdibenromdhane:
> @mahdibenromdhane : It would be convenient to pass the userID has a query parameter to have it already populated in the response. What do you think?
> 
> image

Good Idea, I did the necessary changes.

ClaudeROSSI (approved):

PR #3583, opened 2022-06-03
pricing - price roamin sessions

ClaudeROSSI:
putting back the pricing of roaming sessions


ClaudeROSSI:
@LucasBrazi06 : pricing roaming sessions: 
est-ce que j'oublie quelque chose ?

LucasBrazi06 (approved):
Looks gut!

LucasBrazi06:
@ClaudeROSSI As discussed on the phone, let's put this PR in Iceboxed pipeline the time we have a contextual pricing via tags :-)

ClaudeROSSI:
DO NOT MERGE IT!

After several discussions, we agreed that it does not make sense to price Roaming sessions on our side. 
The actual tariff is not really known and most of the time the result we get on our side does not match the price computed by the roaming platform. 

My proposal would be to introduce a second Pricing Engine, which will be dedicated to pricing the session based on the actual OCPI pricing definition or to get pricing information directly from the roaming platform.

PR #3608, opened 2022-06-14
Billing platform interfaces

ClaudeROSSI:
Billing PLatform - Added interfaces for enabling transfer of funds


Ninjeneer (approved):
Minor correction, but it's ok

PR #3646, opened 2022-06-22
Added check for energy consumption

OliveGerste:


test/api/BillingTest.ts
```diff
@@ -959,6 +959,18 @@ describeif(isBillingProperlyConfigured)('Billing', () => {
...
+
+      it('should detect and ignore inconsistent meter values', async () => {
+        await billingTestHelper.initChargingStationContext2TestCS3Phased();
+        await billingTestHelper.userService.billingApi.forceSynchronizeUser({ id: billingTestHelper.userContext.id });
+        const userWithBillingData = await billingTestHelper.billingImpl.getUser(billingTestHelper.userContext);
+        await billingTestHelper.assignPaymentMethod(userWithBillingData, 'tok_fr');
+        const transactionID = await billingTestHelper.generateTransactionWithWrongMeterValues(billingTestHelper.userContext);
+        assert(transactionID, 'transactionID should not be null');
+        // Check that we have a new invoice with an invoiceID and an invoiceNumber
+        await billingTestHelper.checkTransactionBillingData(transactionID, BillingInvoiceStatus.PAID, 1502.00);
```
ClaudeROSSI, line 971: @OliveGerste : I added that test ... but with a wrong expected value for now (1502.00) ... 
I need to improve the test to only send a few inconsistent values and  have a decent price at the end.

test/api/BillingTestHelper.ts
```diff
@@ -635,6 +635,54 @@ export default class BillingTestHelper {
...
+    const currentTime = startDate.clone();
+    let cumulated = 0;
+    // Send meter values
+    for (let index = 0; index < 50; index++) {
+      cumulated += meterValue;
+      await this.sendConsumptionMeterValue(connectorId, transactionId, currentTime, cumulated);
+    }
+    // Wrong meter value
+    const wrongAccumulatedValue = cumulated + (meterValue * 100);
+    await this.sendConsumptionMeterValue(connectorId, transactionId, currentTime, wrongAccumulatedValue);
```
ClaudeROSSI, line 666: @OliveGerste : I'm trying here to simulate the wrong value (as it occurred on the faulty charger).
However, I'm not sure this approach is correct!

test/api/BillingTestHelper.ts
```diff
@@ -109,8 +109,8 @@ export default class BillingTestHelper {
 
   public async initChargingStationContext2TestChargingTime() : Promise<ChargingStationContext> {
     this.siteContext = this.tenantContext.getSiteContext(ContextDefinition.SITE_CONTEXTS.SITE_BASIC);
-    this.siteAreaContext = this.siteContext.getSiteAreaContext(ContextDefinition.SITE_AREA_CONTEXTS.WITH_SMART_CHARGING_THREE_PHASED);
-    this.chargingStationContext = this.siteAreaContext.getChargingStationContext(ContextDefinition.CHARGING_STATION_CONTEXTS.ASSIGNED_OCPP16 + '-' + ContextDefinition.SITE_CONTEXTS.SITE_BASIC + '-' + ContextDefinition.SITE_AREA_CONTEXTS.WITH_SMART_CHARGING_THREE_PHASED + '-singlePhased');
+    this.siteAreaContext = this.siteContext.getSiteAreaContext(ContextDefinition.SITE_AREA_CONTEXTS.WITH_TARIFFS);
+    this.chargingStationContext = this.siteAreaContext.getChargingStationContext(ContextDefinition.CHARGING_STATION_CONTEXTS.ASSIGNED_OCPP16 + '-' + ContextDefinition.SITE_CONTEXTS.SITE_BASIC + '-' + ContextDefinition.SITE_AREA_CONTEXTS.WITH_TARIFFS);
```
ClaudeROSSI, line 113: @OliveGerste : Tests were using a slow charger (connector type 2) - I'm not using a DC charger instead

ClaudeROSSI (commented):
Still some room for imprements in the tests

src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -1838,4 +1839,37 @@ export default class OCPPUtils {
...
     }
   }
+
+  private static async isValidConsumption(meterValueWh: number, lastMeterValueWh: number, durationSecs: number, chargingStation: ChargingStation, transaction: Transaction,
+      tenantID: string, meterValue: OCPPNormalizedMeterValue): Promise<boolean> {
+    // Calculate consumed energy
+    const consumptionWh = Utils.createDecimal(meterValueWh).minus(lastMeterValueWh).toNumber();
+    // Calculate max drawable energy plus buffer of 20%
+    const power = Utils.getChargingStationPower(chargingStation, null, transaction.connectorId);
+    const maxWh = Utils.createDecimal(power).div(3600).mul(durationSecs).mul(1.2).toNumber();
```
ClaudeROSSI, line 1849: @tobias: Why do we have here an extra buffer (20%) ... is this necessary?

src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -1838,4 +1839,37 @@ export default class OCPPUtils {
       chargePoint.connectorIDs.push(connector.connectorId);
     }
   }
+
+  private static async isValidConsumption(meterValueWh: number, lastMeterValueWh: number, durationSecs: number, chargingStation: ChargingStation, transaction: Transaction,
+      tenantID: string, meterValue: OCPPNormalizedMeterValue): Promise<boolean> {
+    // Calculate consumed energy
+    const consumptionWh = Utils.createDecimal(meterValueWh).minus(lastMeterValueWh).toNumber();
+    // Calculate max drawable energy plus buffer of 20%
+    const power = Utils.getChargingStationPower(chargingStation, null, transaction.connectorId);
```
ClaudeROSSI, line 1848: @OliveGerste : Is it normal to have null as a second parameter? What is the impact here?

ClaudeROSSI (changes requested):
I have some questions/comments!

src/server/ocpp/utils/OCPPUtils.ts
OliveGerste: @ClaudeROSSI The second parameter represents the charge point. So if it is not filled the method should check at connector level and return the connector power.

src/server/ocpp/utils/OCPPUtils.ts
OliveGerste: @ClaudeROSSI We can remove the buffer if you want. I thought to remember a conversation, where we said to add a buffer, for only small inconsistencies, but maybe I have confused this. Let me know

test/api/BillingTestHelper.ts
OliveGerste: @ClaudeROSSI From my perspective the approach is correct.
The check uses the time between the meter values and the capacity of the connector to check the consistency.
So in this case the time difference is 60 seconds. and the meter value is 1000Wh * 100 -> 100000Wh in 60 seconds.
The charger (150kW, when one connector is used) can do 150000 / 60 --> 2500Wh in 60 seconds. 
The consumption should be ignored in this case.

src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -664,7 +664,8 @@ export default class OCPPUtils {
         const meterValueWh = meterValue.attribute.unit === OCPPUnitOfMeasure.KILO_WATT_HOUR ?
           Utils.createDecimal(Utils.convertToFloat(meterValue.value)).mul(1000).toNumber() : Utils.convertToFloat(meterValue.value);
         // Check if valid Consumption
-        if (await OCPPUtils.isValidConsumption(meterValueWh, lastConsumption.value, durationSecs, chargingStation, transaction, tenant.id, meterValue)) {
+        if (meterValueWh > lastConsumption.value
```
OliveGerste, line 667: Why do we keep this check in the initial if statement? We will not be able to log this error later in this case right?

ClaudeROSSI: > Why do we keep this check in the initial if statement? We will not be able to log this error later in this case right?

Without that check some tests fail ... because we do not reach the else block anymore when meterValueWh === lastConsumption.value. 
Let's have a short call to clarify this!

src/server/ocpp/utils/OCPPUtils.ts
ClaudeROSSI: please remove the extra buffer

src/server/ocpp/utils/OCPPUtils.ts
```diff
@@ -1838,4 +1838,43 @@ export default class OCPPUtils {
...
+        action: ServerAction.OCPP_METER_VALUES,
+        message: `Energy consumption is not inconsistent and will be ignored - transaction ID: ${transaction.id} - consumption: ${consumptionWh} Wh - maximum value allowed: ${maxWh} Wh`,
+        detailedMessages: { meterValue, transaction }
+      });
+      return false;
+    }
+    if (meterValueWh > lastMeterValueWh) {
+      return true;
+    }
+    return false;
```
ClaudeROSSI, line 1878: I do not understand - line 1878 cannot be reached since the two first checks already exit the method. Right?

ClaudeROSSI (changes requested):
some remarks

src/server/ocpp/utils/OCPPUtils.ts
OliveGerste: We have some unit tests where meterValueWh is NaN, which will fail without the statement. 
image

src/server/ocpp/utils/OCPPUtils.ts
ClaudeROSSI: > @ClaudeROSSI The second parameter represents the charge point. So if it is not filled the method should check at the connector level and return the connector power.

I do not feel comfortable with that "null" as the second parameter. The getChargingStationPower coding is confusing to me. It sounds like the behavior may change depending on how the template is being defined (by Yassine). What happens if the template does not provide any information at the connector level. This is allowed? Is this supposed to work?

OliveGerste: In this case getChargingStationPower relays on the connector of the charging station in the database. 
If the connectors power is not defined, it tries to use amperage and voltage of the connector to calculate the cs power.
If the template is not defining the required values, the method should return 0. In the case of 0 the check should be skipped.

PR #3648, opened 2022-06-23
fix - billingInvoice.user is not set

ClaudeROSSI:
New invoice notifications are never sent!

image

### DIFF
--- a/src/notification/NotificationTask.ts
+++ b/src/notification/NotificationTask.ts
@@ -29,10 +29,7 @@ export default interface NotificationTask {
   sendComputeAndApplyChargingProfilesFailed(data: ComputeAndApplyChargingProfilesFailedNotification, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
   sendSessionNotStarted(data: SessionNotStartedNotification, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
   sendEndUserErrorNotification(data: Partial<EndUserErrorNotification>, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
-  // TODO : Delete sendBillingNewInvoice ???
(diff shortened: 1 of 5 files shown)


src/notification/NotificationTask.ts
```diff
@@ -29,7 +29,7 @@ export default interface NotificationTask {
   sendComputeAndApplyChargingProfilesFailed(data: ComputeAndApplyChargingProfilesFailedNotification, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
   sendSessionNotStarted(data: SessionNotStartedNotification, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
   sendEndUserErrorNotification(data: Partial<EndUserErrorNotification>, user: User, tenant: Tenant, severity: NotificationSeverity): Promise<void>;
-  // TODO : Delete sendBillingNewInvoice ??
+  // TODO : Delete sendBillingNewInvoice ???
```
ClaudeROSSI, line 32: @cabch : Remaining todo! Do we need to clean up some code here?

src/types/UserNotifications.ts
```diff
@@ -285,7 +285,7 @@ export interface BillingPeriodicOperationFailedNotification extends BaseNotifica
   evseDashboardBillingURL: string;
 }
 
-// TODO: delete unused Utils.build urls
+// TODO: delete unused Utils.build urls ???
```
ClaudeROSSI, line 288: @cabch : idem - remaining todo - is there something to remove here?

src/notification/NotificationTask.ts
cabch: sendBillingNewInvoice has been modified inside to handle Paid or Open notifications, avoid having two sendBillingNotif

PR #3666, opened 2022-06-29
Fix site/sitearea/company filters

Ramzay:
- Remove include external sites auth filter and OR definition
- Create 2 dynamic auth filters to handle external elements
- Update charging station auth filters
- Update site, sitearea, company auth definition
- Update OCPI auth check


Ramzay:
@LucasBrazi06 as discussed please have a look at this PR.
The filters now work as expected, and charging stations / connector seem to also work like before (no regression)
Will tacle the OwnUser issue using this branch as a reference.

PS: this modification doubles the number of requests to the DB, don't know if we can cache some auth queries to reduce number of requests ?

LucasBrazi06 (approved):
GUT

PR #3670, opened 2022-06-30
billing taxes endpoint

ClaudeROSSI:
getTaxes was using the deprecated REST endpoint.


src/server/rest/v1/service/BillingService.ts
```diff
@@ -175,7 +175,13 @@ export default class BillingService {
     }
     // Get taxes
     const taxes = await billingImpl.getTaxes();
-    res.json(taxes);
+    const dataResult: BillingTaxDataResult = {
+      count: taxes.length,
+      result: taxes,
+    };
+    // We do not have a tax collection on our side - the data comes from the connected account (STRIPE)
+    // await AuthorizationService.addTaxesAuthorizations(req.tenant, req.user, dataResult, authorizations, filteredRequest);
```
ClaudeROSSI, line 183: @Ramzay , @cabch - I would need your advice regarding authorizations.
The tax collection does not exists on our side. The data comes from STRIPE.
Do we have to implement the commented line (shown here) in that case?

ClaudeROSSI: > @Ramzay , @cabch - I would need your advice regarding authorizations. The tax collection does not exist on our side. The data comes from STRIPE. Do we have to implement the commented line (shown here) in that case?

Well ... forget it ... I've created the corresponding interfaces even though we do not use them!

PR #3676, opened 2022-07-02
Redirect simplification

LucasBrazi06:


src/utils/Logging.ts
```diff
@@ -677,7 +677,8 @@ export default class Logging {
     }
     // Send error
     if (!res.headersSent) {
-      const errorCode = exception['params'] ? exception['params']['errorCode'] : HTTPError.GENERAL_ERROR;
+      const errorCode = exception['params'] && exception['params']['errorCode'] ?
+        exception['params']['errorCode'] : HTTPError.GENERAL_ERROR;
```
ClaudeROSSI, line 681: @LucasBrazi06 : I confirm that this solves the timeout issues that we had with the billing platform tests.
However, it's a bit confusing to see that the fix is in the loging.ts file. I was expecting to have the request/response logic somewhere else.

ClaudeROSSI (approved):
ok. This solves the test timeout issues!

I have a comment! ;)

PR #3728, opened 2022-07-20
jest-test - allow stopping DB connection

ClaudeROSSI:
minimal changes but please double check!

### DIFF
--- a/test/api/BillingPlatformTest.ts
+++ b/test/api/BillingPlatformTest.ts
@@ -39,6 +39,7 @@ describeif(isBillingProperlyConfigured)('Billing Platform (utbillingplatform)', 
   });
 
   afterAll(async () => {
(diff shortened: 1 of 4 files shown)


src/storage/mongodb/MongoDBStorage.ts
```diff
@@ -269,7 +278,8 @@ export default class MongoDBStorage {
       }
     );
     // Get the EVSE DB
-    this.database = mongoDBClient.db();
+    this.mongoDBClient = mongoDBClient;
+    this.database = this.mongoDBClient.db();
```
ClaudeROSSI, line 282: @LucasBrazi06 / @Ramzay - just keeping a ref to the mongo DB client! ... nothing more to avoid any risks!

PR #3778, opened 2022-08-08
email - simplified, flexible and translatable

ClaudeROSSI:


ClaudeROSSI:
@vicyyn : This is a DRAF PR just to make sure the server tests are still running

ClaudeROSSI:
Here is the result:

PR #3783, opened 2022-08-10
Add platform fee tax ID support in UI when billing platform is enabled

RohanAntony:
[ev-dashboard](https://github.com/sap-labs-france/ev-dashboard/pull/3335) changes


src/assets/server/rest/v1/schemas/billing/billing-setting-update.json
```diff
@@ -29,7 +29,11 @@
           "type": "boolean",
           "sanitize": "mongo"
         },
-        "taxID": {
+        "customerTaxID": {
+          "type": "string",
+          "sanitize": "mongo"
+        },
```
ClaudeROSSI, line 35: @RohanAntony : taxID is used when billing end-users sessions. You must keep it!
Simply mimic what we have and add a second taxID for billing the commission fees.
Please call it platfiormFeeTaxID to make it clear where it is used!

The taxID that MUST be removed is the one that we have in the BillingAccount interface (in Billing.ts). But don't remove it yet. Just add the new one for now!

ClaudeROSSI (changes requested):
@RohanAntony : Check my first comment.

Please also run the billing tests locally, to make sure you ddo not break the current billing process.

>>> npm run test:create:utbilling
>>> npm run test:billing

ClaudeROSSI (approved):
ok

PR #3808, opened 2022-08-23
Add targetStateOfCharge to transaction

OliveGerste:


src/server/rest/v1/service/TransactionService.ts
```diff
@@ -383,6 +383,8 @@ export default class TransactionService {
   }
 
   public static async handleTransactionStart(action: ServerAction, req: Request, res: Response, next: NextFunction): Promise<void> {
+    // Cache request to avoid modifications from validator
+    const cachedReq = Utils.cloneObject(req);
```
ClaudeROSSI, line 387: @OliveGerste : Why do we need to clone?
I do not understand. The purpose of the validator is typically to make sure the request does not contain anything suspicious. If you clone, you simply bypass the validator. rigth? 
let's sync on that.

ClaudeROSSI (changes requested):
Please have a look to my comment

src/server/rest/v1/service/TransactionService.ts
OliveGerste: @ClaudeROSSI Hi Claude, this is a fix for a special case we have with the remote start transaction. 
This is executable from 2 different endpoints. Charging Station endpoint and Transaction endpoint.
The transaction endpoint is reusing the functionality of the charging station endpoint.
Because of this the validator is executed twice. 
This would be no problem if the validator would not change anything, but with the custom type 'date' which is used for the departure time, the first validation will transform the property on the request from string into a date, which lets the second validation fail (it checks for type string again)
This is why I clone the request to let the second validation use the original request (which is the same one as the first validation)

ClaudeROSSI: ok. I understand ... but i think we already had such use-case and the solution was to ask the validator to clone it before modifying anything.

Can you remove your explicit call to clone and try the option the validator layer provides instead.

Maybe:

ClaudeROSSI (changes requested):
see my comment.
Thanks
Claude

src/server/rest/v1/service/TransactionService.ts
OliveGerste: @ClaudeROSSI Ah great news! Thanks for the information.
I changed it. Now the parameter of the validation method is used to clone the request on the remote starts.

ClaudeROSSI (approved):
thanks

PR #3837, opened 2022-09-01
Realign ev-database imports

mahdibenromdhane:


ClaudeROSSI:
@Ramzay : could you please review that one? 
@mahdibenromdhane : Don't we have a new URL to call ? I thought that v26 has to be replaced by v29 somewhere. 
Thanks ,
Claude

src/integration/car/ev-database/EVDatabaseCarIntegration.ts
```diff
@@ -39,54 +40,96 @@ export default class EVDatabaseCarIntegration extends CarIntegration {
...
-          chargePhaseVolt: data.Charge_Standard_Table[chargeStandard].Charge_PhaseVolt,
-          chargePhaseAmp: data.Charge_Standard_Table[chargeStandard].Charge_PhaseAmp,
-          chargePhase: data.Charge_Standard_Table[chargeStandard].Charge_Phase,
-          chargePower: data.Charge_Standard_Table[chargeStandard].Charge_Power,
-          chargeTime: data.Charge_Standard_Table[chargeStandard].Charge_Time,
-          chargeSpeed: data.Charge_Standard_Table[chargeStandard].Charge_Speed,
-        };
-        chargeStandardTables.push(chargeStandardTable);
+      let chargeStandardTableUniversal;
+      if (data.Charge_Standard_Table_UK) {
```
Ramzay, line 44: Is there a reason why we prioritise UK > NL > DE for the standard table ?

src/integration/car/ev-database/EVDatabaseCarIntegration.ts
```diff
@@ -39,54 +40,96 @@ export default class EVDatabaseCarIntegration extends CarIntegration {
...
-            evsePhase: data.Charge_Standard_Table[chargeOption].EVSE_Phase,
-            chargePhaseVolt: data.Charge_Standard_Table[chargeOption].Charge_PhaseVolt,
-            chargePhaseAmp: data.Charge_Standard_Table[chargeOption].Charge_PhaseAmp,
-            chargePhase: data.Charge_Standard_Table[chargeOption].Charge_Phase,
-            chargePower: data.Charge_Standard_Table[chargeOption].Charge_Power,
-            chargeTime: data.Charge_Standard_Table[chargeOption].Charge_Time,
-            chargeSpeed: data.Charge_Standard_Table[chargeOption].Charge_Speed,
-          };
-          chargeOptionTables.push(chargeAlternativeTable);
+        if (data.Charge_Alternative_Table) {
```
Ramzay, line 68: According to the spec this fiels no longer exists... So will always be empty.

Shouldn't we put all other Charge_Standard_Table_<COUNTRY> that wasn't used as standard (and make sure not null)

src/integration/car/ev-database/EVDatabaseCarIntegration.ts
```diff
@@ -39,54 +40,96 @@ export default class EVDatabaseCarIntegration extends CarIntegration {
...
+              chargePhaseAmp: chargeStandardTableUniversal[chargeAlternative].Charge_PhaseAmp,
+              chargePhase: chargeStandardTableUniversal[chargeAlternative].Charge_Phase,
+              chargePower: chargeStandardTableUniversal[chargeAlternative].Charge_Power,
+              chargeTime: chargeStandardTableUniversal[chargeAlternative].Charge_Time,
+              chargeSpeed: chargeStandardTableUniversal[chargeAlternative].Charge_Speed,
+            };
+            chargeAlternativeTables.push(chargeAlternativeTable);
+          }
+        }
+        if (data.Charge_Option_Table_DE) {
```
Ramzay, line 85: Here you prioritise DE > UK > NL (not the same as standard table)

Why don't you build the chargeOptionTables using the 3 countries and make sure that we have unique  chargeOption ? This would render the logique less dodgy and would handle the case where an option was set for one country but not another ?

Also: code duplication create a method

src/integration/car/ev-database/EVDatabaseCarIntegration.ts
```diff
@@ -10,6 +10,7 @@ import Logging from '../../../utils/Logging';
 import { ServerAction } from '../../../types/Server';
 import Utils from '../../../utils/Utils';
 import { Voltage } from '../../../types/ChargingStation';
+import fs from 'fs';
```
Ramzay, line 13: Not used

Ramzay (changes requested):
Please check my remarks !

There is a great lack of documentation in this code, it is difficult to understand what is going on here !
Since the last version there where new fields that were added in the response of the API would be nice to add them ex: Battery_TMS

Please make sure to read the API documentation and changes.

Ramzay:
> @Ramzay : could you please review that one? @mahdibenromdhane : Don't we have a new URL to call ? I thought that v26 has to be replaced by v29 somewhere.
> 
> Is it in the kubernetes config?
> 
> ## "EVDatabase": {
> Thanks , Claude

I will add the configuration in the other repo (aws and kubernetes)

@mahdibenromdhane could you also create an issue and link it to this PR ?

src/integration/car/ev-database/EVDatabaseCarIntegration.ts
mahdibenromdhane: The Charge_Standard_Table_UK is not allowed to be null according to the documentation. 
However, for some cases it's null in UK and provided in NL and DE.

mahdibenromdhane:
> Please check my remarks !
> 
> There is a great lack of documentation in this code, it is difficult to understand what is going on here ! Since the last version there where new fields that were added in the response of the API would be nice to add them ex: Battery_TMS
> 
> Please make sure to read the API documentation and changes.

I added the new missing attributes of the V29 api.

src/integration/car/ev-database/EVDatabaseCarIntegration.ts
mahdibenromdhane: removed.

src/integration/car/ev-database/EVDatabaseCarIntegration.ts
mahdibenromdhane: Done, now the standard table and the option table is created using 3 the 3 countries table.
Priority: DE > NL > UK. 
Also I created a method to avoid code duplication.

src/integration/car/ev-database/EVDatabaseCarIntegration.ts
mahdibenromdhane: Removed the chargeAlternativeTable since is no longer used.

src/integration/car/ev-database/EVDatabaseCarIntegration.ts
```diff
@@ -36,171 +34,57 @@ export default class EVDatabaseCarIntegration extends CarIntegration {
...
-        drivetrainPower: data.Drivetrain_Power,
         drivetrainPowerHP: data.Drivetrain_Power_HP,
         drivetrainTorque: data.Drivetrain_Torque,
         performanceAcceleration: data.Performance_Acceleration,
         performanceTopspeed: data.Performance_Topspeed,
-        rangeWLTP: data.Range_WLTP,
-        rangeWLTPEstimate: data.Range_WLTP_Estimate,
-        rangeNEDC: data.Range_NEDC,
-        rangeNEDCEstimate: data.Range_NEDC_Estimate,
         rangeReal: data.Range_Real,
```
ClaudeROSSI, line 49: As far as I can see the Car Catalog list is showing the rangeWLTP.

Please put it back!

Check the screenshot below.

Some columns are always empty when importing using the latest API.

image

ClaudeROSSI (changes requested):
Some data is missing. 

Sopme columns are empty in the car catalog.

image

Please double-check.
Thanks,
Claude

PR #3838, opened 2022-09-01
Admin can update tenant data

mahdibenromdhane:


ClaudeROSSI (changes requested):
coordinations - does not mean what you think

do you mean coordinates?

Anyway - let's use a more generic term

I would suggest "DATA" or "CONFIGURATION"

- tenant-update-data
or 
- tenant-update-configuration

Please change the term used everywhere!

ClaudeROSSI:
@Ramzay : I have asked Mehdi to allow the user to change the tenant settings (without being super admin).
Please help me to review this PR from the authorization perspective.
@mahdibenromdhane : You have been adding new endpoints - I'm wondering if existing endpoints could be reused instead. Just an idea ... I'm not sure if this makes sense at all.

src/authorization/AuthorizationsDefinition.ts
```diff
@@ -102,6 +102,10 @@ export const AUTHORIZATION_DEFINITION: AuthorizationDefinition = {
   },
   admin: {
     grants: [
+      {
+        resource: Entity.TENANT, action: [Action.READ, Action.UPDATE],
+        attributes: ['id', 'name', 'email', 'logo', 'address']
```
Ramzay, line 107: You are missing some attributes for both READ and UPDATE actions.
 Since both actions also require different attributes would be best to separate them.

src/server/rest/v1/service/TenantService.ts
```diff
@@ -381,6 +381,46 @@ export default class TenantService {
...
+      throw new AppAuthError({
+        errorCode: HTTPAuthError.FORBIDDEN,
+        user: req.user,
+        action: Action.UPDATE, entity: Entity.TENANT,
+        module: MODULE_NAME, method: 'handleUpdateTenantCoordinations',
+        value: filteredRequest.id
+      });
+    }
+    // Get
+    const tenant = await TenantStorage.getTenant(filteredRequest.id);
```
Ramzay, line 397: Apply projected field you are returning the full object

Ramzay (changes requested):

src/authorization/AuthorizationsDefinition.ts
mahdibenromdhane: I'v separated the actions

src/server/rest/v1/service/TenantService.ts
mahdibenromdhane: As discussed, this will be available after the tenants auth migration.

mahdibenromdhane:
> coordinations - does not mean what you think
> 
> do you mean coordinates?
> 
> Anyway - let's use a more generic term
> 
> I would suggest "DATA" or "CONFIGURATION"
> 
> * tenant-update-data
>   or
> * tenant-update-configuration
> 
> Please change the term used everywhere!

Done, replaced with data.

mahdibenromdhane:
> @Ramzay : I have asked Mehdi to allow the user to change the tenant settings (without being super admin). Please help me to review this PR from the authorization perspective. @mahdibenromdhane : You have been adding new endpoints - I'm wondering if existing endpoints could be reused instead. Just an idea ... I'm not sure if this makes sense at all.

@ClaudeROSSI the admin cannot update tenant's components, subdomain .. Thats why I'v separated the updateTenantData from the updateTenant.

PR #3839, opened 2022-09-02
Smart charging - Prioritization parameters

OliveGerste:


src/assets/server/rest/v1/docs/e-mobility-oas.json
```diff
@@ -16260,6 +16260,18 @@
                   },
                   "limitBufferAC": {
                     "type": "number"
+                  },
+                  "usePrioritizationParameters": {
```
ClaudeROSSI, line 16264: As discussed during our call, I think we may have additional options here (in the future). So, I would like a more generic name:

Something like:
**useAdvancedParameters**

Let me know what you think.

src/assets/server/rest/v1/docs/e-mobility-oas.json
```diff
@@ -16260,6 +16260,18 @@
...
+                    "type": "boolean"
+                  },
+                  "defaultSessionTimeHours": {
```
ClaudeROSSI, line 16267: A still do not understand the purpose of a "**defaultSessionTimeHours**" with a max value of 20 hours.
Why 20 and not 21?

When discussing with the boss ;) the use-case we have in mind does not need duration. We need instead to declare the time when the end-user is supposed to leave (and unplug his/her car).

So, something like:

Regular departure time: 18 o'clock - Which of course depends of the timezone of the charging station location!
What would you suggest?

Thanks

ClaudeROSSI (changes requested):
Some comments.

PR #3850, opened 2022-09-15
test - create context - troubleshoot boot notification

ClaudeROSSI:
The script **test:createContext** does not work anymore.

Charging stations cannot be created because we do not get the expected data after a boot notification.
Sounds like the new template format has a impact. But it works when running it locally.

I'll add some logs to troubleshoot the create context behavior.


ClaudeROSSI:
@cabch : The tests are failing because we have no templates in the DB.

According to the traces that I have just added the first charging station being created is the one below:
- "chargePointVendor":"Schneider Electric"
- "chargePointModel":"MONOBLOCK",
- "chargePointSerialNumber":"9vbyq4ogfa6rkdg9q9q4iy05f"

Without a template the boot notification succeeds but the charging station does not match what the test expects.
Let's clarify how to solve this issue.

image

ClaudeROSSI:
@cabch : The **charging station factory** provides the list of template that are mandatory for running the tests.

Here it is:
image

PR #3878, opened 2022-09-29
Retrieve charging stations without auth (canXXX) for each record

LucasBrazi06:


LucasBrazi06:
@Ramzay Useful if you want to retrieve the list of CS without the canXXX (B2B), it can divide by up to 10 the volume of data.
Let me know and if you agree, and I'll propagate the changes the other endpoints that return a list + update OAS file to include the new query param WithAuth.

Ramzay:
> @Ramzay Useful if you want to retrieve the list of CS without the canXXX (B2B), it can divide by up to 10 the volume of data. Let me know and if you agree, and I'll propagate the changes the other endpoints that return a list + update OAS file to include the new query param WithAuth.

Fine by me ! Please propagate the modification to the other endpoints and we will merge it !

LucasBrazi06:
@ClaudeROSSI  Did it for all endpoints returning a list for the map where you only need connector statuses + GPS coordinates to display all the markers.
The mobile has to retrieve all the CS (max 1000) as the clustering is done in the app and not in the backend ;-) (JNF).

Ramzay (approved):

PR #3898, opened 2022-10-05
added mentions legales to technical settings / tenant

cabch:


src/assets/schemas/common/common.json
```diff
@@ -187,6 +187,27 @@
       "type": "object",
       "nullable": true,
       "sanitize": "mongo"
+    },
+    "tenantSupport": {
```
LucasBrazi06, line 191: To move tenantSupport -> tenant.json ?
Why in common.json as it's specific to the Tenant object?

src/assets/server/rest/v1/schemas/tenant/tenant-data-update.json
```diff
@@ -10,13 +10,41 @@
...
+    },
+    "siret": {
+      "type": "number",
+      "sanitize": "mongo",
+      "pattern": "^[0-9]{14}$"
+    },
+    "legalStatus": {
+      "type": "string",
+      "sanitize": "mongo"
     }
```
LucasBrazi06, line 48: Maybe all this data goes to common.js as it can be reused in Company ;P
And why not making it a component like the address ?

src/assets/storage/schemas/tenant/tenant-save.json
```diff
@@ -20,6 +20,34 @@
     },
     "components": {
       "$ref": "tenant-components#/definitions/components"
+    },
+    "support": {
```
LucasBrazi06, line 24: Also generic, can be reused in Company for instance ;-)

LucasBrazi06 (commented):
My 2 cents

src/assets/schemas/common/common.json
cabch: no specific reson, I just didn't realized i left it here 😇

src/assets/server/rest/v1/schemas/tenant/tenant-data-update.json
cabch: it could be reused by companies but we won't use it as seen with @ClaudeROSSI

cabch: and yes i should have made a new component as address or support 🤷🏻‍♀️

PR #3912, opened 2022-10-14
Track white label mobile apps + used mobile version

LucasBrazi06:
Store user mobile Version to track currently used apps.
Store all user mobile root info into a new mobileData property.
Return properly empty Promises in RemotePushNotificationTask class.

@vicyyn
Having a type named NotificationResult which is generic for all notif implementations (email, RPN) that contains props like 'html' is bad practice and means nothing when you are in RemotePushNotificationTask class.
Create a main class with common attributes and create sub-classes it for each implementation.

@AlixH 
Mobile App needs to send this information :
```
          await this.centralServerProvider.saveUserMobileData(this.centralServerProvider.getUserInfo().id, {
            mobileToken: this.getToken(),
            mobileOS: Platform.OS,
            mobileBundleID: DeviceInfo.getBundleId(),
            mobileAppName: DeviceInfo.getApplicationName(),
            mobileVersion: DeviceInfo.getVersion(),
          });

```
Btw, CentralServerProvider.saveUserMobileToken() has been renamed to saveUserMobileData()


ClaudeROSSI:
@LucasBrazi06 : One test is failing. 
It seems that the root cause is the renaming of CentralServerProvider.saveUserMobileToken() to saveUserMobileData().
I didn't check it yet!

LucasBrazi06:
> @LucasBrazi06 : One test is failing. 
> It seems that the root cause is the renaming of CentralServerProvider.saveUserMobileToken() to saveUserMobileData().
> I didn't check it yet!

Fixé dans les unit tests.
The Mobile OS value was wrong !
Nothing to do with the endpoint 😉.

LucasBrazi06:
@ClaudeROSSI Btw, the endpoint is backward compatible with old versions of the mobile app like white labels.

PR #3919, opened 2022-10-18
stripe - use 'express' as sub account types

ClaudeROSSI:


Ramzay (approved):

ClaudeROSSI:
Tests are not working as expected when using 'express' account.
We get STRIPE exception when doing transfers because the destination account  is missing some capabilities.
I've tried to explicit enable the capabilities that we need (at the account creation) but this has no impact. 
I'll need to sync with our STRIPE contact to clarify how automated tests can somehow 'complete' the account creation and activate all the required capabilities!

PR #3943, opened 2022-10-26
Email - subject - more info

ClaudeROSSI:


src/assets/i18n/de.json
```diff
@@ -343,7 +343,7 @@
       ]
     },
     "charging-station-status-error": {
-      "title": "Ladestation - Fehler",
+      "title": "Ladestation - Fehler an {{chargeBoxID}}, Ladepunkt {{connectorId}}",
```
ClaudeROSSI, line 346: @OliveGerste : customers want more info in the subject.
The subject includes the title!
So I changed it ... is it still correct in German?

Thanks
Claude

ClaudeROSSI:
@OliveGerste : I had to change the subject of one email.
Can you please double check the German translation?

PR #3975, opened 2022-11-09
New user context method based on handle get user default tag car

OliveGerste:
Added prioritization defaults for smart charging as well. Therefore the data model for smart charging settings was changed.
If confirmed the data model in the smart charging prio draft PR will be adapted as well. It can be easily extended by other defaults


src/server/rest/v1/service/UserService.ts
```diff
@@ -49,6 +51,70 @@ import moment from 'moment';
...
+      withBillingChecks = false;
+    }
+    // Check for billing errors
+    const errorCodes: Array<StartTransactionErrorCode> = [];
+    if (withBillingChecks) {
+      // Check for the billing prerequisites (such as the user's payment method)
+      await UserService.checkBillingErrorCodes(action, req.tenant, req.user, user, errorCodes);
+    }
+    const settingAuthorization = await AuthorizationService.checkAndGetSettingAuthorizations(req.tenant, req.user, Action.READ);
+    let prioritizationParameters: PrioritizationParameters;
```
ClaudeROSSI, line 103: **PrioritizationParameters** is a specific type from the SmartCharging settings. I don't like the fact that we directly pass this information to the client.

The contract between the client and the server should use a dedicated interface in order to abstract a bit from the actual implementation.
e.g.: **SmartChargingSessionParameters** 

I discussed the main use-case again (with our boss) and this interface will certainly combine information coming from both the **SmartCharging** settings but also from the Site or Site Area (or maybe one day from the user preferences).

src/types/Setting.ts
```diff
@@ -207,6 +207,14 @@ export interface SapSmartChargingSetting extends SmartChargingSetting {
...
   limitBufferDC: number;
   limitBufferAC: number;
+  prioritizationParameters: PrioritizationParameters;
+}
+
+export interface PrioritizationParameters {
+  active: boolean;
+  defaultSessionTimeHours: number;
+  defaultInitialStateOfCharge: number;
+  defaultTargetStateOfCharge: number;
```
ClaudeROSSI, line 217: As already mentioned, this interface is for the global settings.
We need a second interface to define the contract with the client (i.e.: the mobile app and the web app).
I'll update our wiki page to make sure we are on the same page regarding the actual use-case.

ClaudeROSSI (changes requested):
I'll update our wiki page to make sure we are on the same page regarding the main use-case that Hanno has in mind.
We will then do a sync again.

src/types/Setting.ts
ClaudeROSSI: @OliveGerste, @AlixH : PLease have a look to my proposal at:
- https://wiki.one.int.sap/wiki/display/OpenEMob/Smart+Charging+-+Departure+Time
Let me know if you feel confortable with it!

OliveGerste: Looks good to me 👍

AlixH: I think carStateOfCharge should also be null when on a DC connector so the UI can hide it
Everything is clear otherwise

PR #3989, opened 2022-11-22
admin cannot have admin nor owner roles

cabch:
- duplicated transaction refund to admin when he is also owner


cabch:
admin being siteAdmin or siteOwner is misbehaving because auths are not taken in account correctly, for now we will continue to duplicate all the authorizations when needed

Ramzay (approved):

PR #4001, opened 2022-12-02
Fix github unit-test workflow

Ramzay:


Ramzay:
@LucasBrazi06 in case you are facing the same issue since yesterday ;)

LucasBrazi06:
@Ramzay Thank you!!!

PR #4008, opened 2022-12-14
added sitesOwnerUsers filter

cabch:
@Ramzay I have a doubt on SitesOwnerUsersDynamicAuthorizationDataSourceData props ?


Ramzay (changes requested):
Hey @cabch,
In SitesOwnerUsersDynamicAuthorizationFilter.ts, you  added a new filter "siteOwnerIDs", which is good, but this filter is never used :)
When creating / setting new filters you always need to make sure that it is handled ! There is unfortunately no black magic.
In this PR, I expect that the transaction entity (transactionStorage) handles this new flag.

cabch:
@ClaudeROSSI this is the PR to merge please

PR #4019, opened 2022-12-27
Logperf2

ClaudeROSSI:


ClaudeROSSI:
just a test on QA with less logs and traces

PR #4072, opened 2023-02-09
Version 2.7.5 - Optimize DB size

ClaudeROSSI:
- expiration on performance collection > 2 days max
- index removed


ClaudeROSSI:
giing up on that one - conflict on aws sub-module configuration

PR #4126, opened 2023-03-01
Logperf7

ClaudeROSSI:


ClaudeROSSI:
@OliveGerste : Tobias, can you help?
We have a smartcharging test failing ... I think I didn't merge properly some changes, but I do not find which one. 
Can you please have a look to the failing test?
Any hint would be helpful!

image

OliveGerste:
> @OliveGerste : Tobias, can you help? We have a smartcharging test failing ... I think I didn't merge properly some changes, but I do not find which one. Can you please have a look to the failing test? Any hint would be helpful!
> 

Hi @ClaudeROSSI,

Not an easy one to be honest. The fault is not on your side:

Short answer: Performance is too good now.
Long answer: The test assumes that the first transaction starts shortly before the second one. Normally this worked because the code execution took more time, so that we had 1 second difference between the transaction starts. 
Due to the performance increase they start at the exact same time and the parameters are exactly the same which leads to a other order how the the transactions are prioritised.
I provided a PR to fix this https://github.com/sap-labs-france/ev-server/pull/4173
Let me know if this is sufficient.

PR #4165, opened 2023-03-30
move handleForceSynchronizeUser to restful

mahdibenromdhane:


src/server/rest/v1/router/api/BillingRouter.ts
```diff
@@ -96,6 +97,12 @@ export default class BillingRouter {
     });
   }
 
+  private buildRouteForceSynchronizeUser(): void {
+    this.router.post(`/${RESTServerRoute.REST_FORCE_SYNCHRONIZE_USER}`, (req: Request, res: Response, next: NextFunction) => {
+      void RouterUtils.handleRestServerAction(BillingService.handleForceSynchronizeUser.bind(this), ServerAction.BILLING_FORCE_SYNCHRONIZE_USER, req, res, next);
+    });
+  }
+
```
ClaudeROSSI, line 105: @mahdibenromdhane : I would like to have an endpoint similar to

>> RESTServerRoute.REST_BILLING_TRANSFER_FINALIZE

src/types/Server.ts
```diff
@@ -704,6 +704,7 @@ export enum RESTServerRoute {
 
   REST_BILLING_SETTING = 'billing-setting', // GET and PUT
   REST_BILLING_CHECK = 'billing/check',
+  REST_FORCE_SYNCHRONIZE_USER = 'billing/force-synchronize-user',
```
ClaudeROSSI, line 707: PLease chenge it to:

REST_BILLING_USER_SYNCHRONIZE = 'billing/users/:id/synchronize',

I need the id in the URL

The HTTP verb should be PATCH instead of POST

Let me know what you think!

ClaudeROSSI (changes requested):
please check my comments

src/types/Server.ts
mahdibenromdhane: Yes, it make more sense this way. 
Done for the changes.

src/server/rest/v1/router/api/BillingRouter.ts
```diff
@@ -96,6 +97,13 @@ export default class BillingRouter {
     });
   }
 
+  private buildRouteForceSynchronizeUser(): void {
+    this.router.patch(`/${RESTServerRoute.REST_BILLING_USER_SYNCHRONIZE}`, (req: Request, res: Response, next: NextFunction) => {
+      req.body.id = req.params.id;
```
mahdibenromdhane, line 102: @ClaudeROSSI let me know if you want to use the directly the req.params in the filteredrequest.
Meanwhile it works this way.

ClaudeROSSI (approved):
thanks

PR #4183, opened 2023-04-13
sm - tests failing if no sort?

ClaudeROSSI:


ClaudeROSSI:
ok ... solved.
Some tests were sensitive to the implicit sort ...
Not the case anymore

